PR #4, opened 2024-09-27
Start implementing nominal type resolution for jextract-swift

DougGregor:
Introduce a basic implementation that can resolve the names of nominal types within a given module, including nested types and extensions of types. Use this to enable translating the members of extensions of types into members of the resulting Java class. It will also (eventually) allow us to cross-reference uses of Swift types within other Swift types.


Sources/JExtractSwift/NominalTypeResolution.swift
```diff
@@ -0,0 +1,353 @@
...
+      assert(numExtensionsBefore > unresolvedExtensions.count)
+    }
+
+    // Any unresolved extensions at this point are fundamentally unresolvable.
+    return unresolvedExtensions
+  }
+}
+
+extension ExtensionDeclSyntax {
+  /// Produce the nested type name for the given
```
ktoso:
```suggestion
  /// Produce the nested type name for the given decl
```

minor cut sentence

Tests/JExtractSwiftTests/NominalTypeResolutionTests.swift
```diff
@@ -0,0 +1,63 @@
...
+import SwiftParser
+import Testing
+
+@Suite("Nominal type lookup")
+struct NominalTypeLookupSuite {
+  func checkNominalRoundTrip(
+    _ resolution: NominalTypeResolution,
+    name: String,
+    fileID: String = #fileID,
+    fileParh: String = #filePath,
```
ktoso, line 25: small typo

```suggestion
    filePath: String = #filePath,
```

ktoso (approved):
Very nice, lgtm

Tests/JExtractSwiftTests/NominalTypeResolutionTests.swift
ktoso: Fixing the typo separately :)

PR #8, opened 2024-09-27
Finish the migration from ImportedTypeName over to TranslatedType

DougGregor:
`TranslatedType` captures more information about how the type is represented in Swift, Java, and the intermediate stages in between. Migrate all remaining functionality from `ImportedTypeName` to `TranslatedType`.

Start resolving translated types through a centralized "lookup" mechanism rather than assuming that the types named in Swift are always going to have Java counterparts. For now, this means we won't try to translate functions that reference types that we don't know will be imported into Java. In the future, it means we'll be able to reason about these types to (e.g.) pass arguments and return values of those types correctly across the Swift/Java barrier.


DougGregor:
@ktoso  I'm going to go ahead and merge this, because I have a follow-up PR that benefits from being separate.

PR #9, opened 2024-09-27
Expose Java-friendly instance methods and constructors for Swift types

DougGregor:
When projecting a Swift type into a Java class, map Swift methods into instance methods (rather than static methods) and Swift initializers into Java constructors (rather than static 'init' methods). This means our example Swift class like this:

```swift
public class MySwiftClass {
  public init(len: Int, cap: Int) { ... }
  public func voidMethod() { ... }
} 
```

turns into

```java
public final class MySwiftClass {
  public MySwiftClass(long let, long cap) { /* downcall to Swift */ }
  public void voidMethod() { /* downcall to Swift */ }
}
```

as one would expect.

### DIFF
--- a/JavaSwiftKitDemo/src/main/java/org/example/HelloJava2Swift.java
+++ b/JavaSwiftKitDemo/src/main/java/org/example/HelloJava2Swift.java
@@ -60,9 +60,12 @@ public class HelloJava2Swift {
 
         JavaKitExample.globalTakeInt(1337);
 
-        MySwiftClass obj = MySwiftClass.init(2222, 7777);
+        MySwiftClass obj = new MySwiftClass(2222, 7777);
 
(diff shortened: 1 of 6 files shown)


lokesh-tr:
Also worth highlighting that this PR fixes an issue where running the `make jextract-run` produces java files with Canonical class names to represent the type instead of its simple name. For example: at the current state `make jextract-run` will produce a java file with
```
public final class com.example.generated.MySwiftClass {
     // it also uses com.example.generated.MySwiftClass in function return types & parameters, etc. where MySwiftClass is supposed to be
```
instead of
```
public final class MySwiftClass {
```

People who try the Swift2Java route are very likely to hit this issue unless this PR gets merged.

DougGregor:
Thank you, Lokesh. I've gone ahead and merged it.

PR #10, opened 2024-09-28
[CI] Add support for GitHub Actions

shahmishal:


ktoso:
Replaced by https://github.com/swiftlang/swift-java/pull/14 since we have some special steps and requirements -- only nightly toolchain, added the JDK installations (checking their checksums etc)

ktoso:
Thank you for preparing these and the shared repo!

PR #11, opened 2024-09-28
Refactor `@ExposeToJava` to `@ImplementsJava` in the `USER_GUIDE`

lokesh-tr:
It seems like `ImplementsJavaMacro` is the name of the macro which allows Swift to provide native implementation of Java methods. Yet, the file name and the `USER_GUIDE` still refer to the macro as `ExposeToJavaMacro`.

This PR renames the file `ExposeToJavaMacro.swift` to `ImplementsJavaMacro.swift` whose macro is already known as the latter. This PR also updates the `USER_GUIDE` accordingly.

*NOTE:* This PR is not proposing any name changes, this just makes the documentation consistent with the implementation.


ktoso (approved):
That's right, the rename was pretty recent -- thanks for the PR!

PR #14, opened 2024-09-30
Prepare for Linux build and CI

ktoso:
This prepares for Linux CI.

- Include docker files for local Linux development
- Pull OpenJDK 23, as 22 is not supported anymore since this month, so we should stick to "supported" versions I think (we can debate this later)
- Disable the `nm` hack on Linux, the entire approach was going to be removed anyway so no reason to implement any of the `Dylib...` stuff parity on other platforms -- we'll just remove it all and replace by depending on [Option to emit mangled names for public symbols into the .swiftinterface #76600](https://github.com/swiftlang/swift/pull/76600) in snapshot Swift
- Pick changes from https://github.com/swiftlang/swift-java/pull/10/ 
    - [x] will polish that up to use the supported versions and commands


lokesh-tr:
Almost ready to merge? 👀🤞🏻

ktoso:
Okey this is a good first step: we have the swift tests passing and we can work on getting the java ones passing next.

all the other soundness things also ok, except the format task has some issues.

Overall we're now waiting for a new nightly toolchain and then we should merge https://github.com/swiftlang/swift-java/pull/27

Meanwhile we can continue to rely on the swift tests for the time being, this is already much better than no CI after all 😉

lokesh-tr:
@ktoso That's great!

ktoso:
This has had a lot of churn so I'll squash it all as "initial ci support" basically

PR #27, opened 2024-10-01
Extract mangled names from .swiftinterface rather than using `nm`

DougGregor:
The Swift compiler recently added a frontend option (in https://github.com/swiftlang/swift/pull/76600) `-abi-comments-in-module-interface` that emits ABI-related comments into the generated Swift interface file. Parse the resulting `// MANGLED NAME:` comments to extract mangled name information for public symbols instead of relying on hardcoded `nm` invocations on the shared library.

Note that this pull request means that `jextract-swift` will require an as-yet-unreleased Swift compiler that comes from the `main` branch. At present, there is no downloadable toolchain containing the compiler fix, so I'm starting by opening this as a draft. Once downloadable toolchains are available for all platforms, we can update the documentation to point `jextract-swift` users there and merge this pull request.

Remove all of the supporting code for querying `nm`, which turns out to be a lot :).

PR #40, opened 2024-10-03
Unlock Java CI, disable Java tests which rely on mangled names

ktoso:
This should bring us to a working Linux CI for Java as well.

The way we load libraries is platform dependent it turns out, so this needed some adjustments. Also fixed some paths and disabled tests which we cannot do until we merge https://github.com/swiftlang/swift-java/pull/27


ktoso:
We're not in great shape yet but I'm trying to get us some level of soundness and at least not breaking basic compilation and things like that.

This now has a separate java tests target and also checks `make` which tries to use Java2Swift/JavaToSwift (which explodes on linux currently still).

ktoso:
Handled in a different PR which took pieces from here.

PR #42, opened 2024-10-04
Model Swift's type metadata and value witness table for memory layout information

DougGregor:
Extend the Java SwiftKit with the ability to query the memory layout of an arbitrary Swift type given its Swift type metadata pointer (e.g., `Any.Type` in the Swift world). Use this to show the size, stride, and alignment of `Int?` from the Java side, determine the name of a Swift type from its metadata, and produce a [`java.lang.foreign.MemoryLayout`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/foreign/MemoryLayout.html) for any Swift type. Here's some example output for the Swift type `Int?` (mangled name: `SiSg`):

    Memory layout for Swift.Int?:
      size = 9
      stride = 16
      alignment = 8
      Java layout = [8%[9:b1]x7](Swift.Optional<Swift.Int>)

### DIFF
--- a/JavaSwiftKitDemo/src/main/java/org/swift/javakit/SwiftKit.java
+++ b/JavaSwiftKitDemo/src/main/java/org/swift/javakit/SwiftKit.java
@@ -64,6 +64,36 @@ public class SwiftKit {
     }
 
     // ==== ------------------------------------------------------------------------------------------------------------
+    // free
+    /**
+     * Descriptor for the free C runtime function.
+     */
+    public static final FunctionDescriptor free$descriptor = FunctionDescriptor.ofVoid(
+        ValueLayout.ADDRESS
+    );
+
+    /**
+     * Address of the free C runtime function.
+     */
+    public static final MemorySegment free$addr = findOrThrow("free");
+
+    /**
+     * Handle for the free C runtime function.
+     */
+    public static final MethodHandle free$handle = Linker.nativeLinker().downcallHandle(free$addr, free$descriptor);
+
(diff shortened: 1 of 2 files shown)


JavaSwiftKitDemo/src/main/java/org/swift/javakit/SwiftKit.java
```diff
@@ -230,4 +231,121 @@ public static MemorySegment getTypeByMangledNameInEnvironment(String string) {
             throw new AssertionError("should not reach here", ex$);
         }
     }
+
+    /**
+     * The value layout for Swift's Int type, which is a signed type that follows
+     * the size of a pointer (aka C's ptrdiff_t).
+     */
+    public static ValueLayout SWIFT_INT = (ValueLayout.ADDRESS.byteSize() == 4) ?
+        ValueLayout.JAVA_INT : ValueLayout.JAVA_LONG;
```
ktoso, line 271: Oh nice, instead of guessing by platform. Lgtm

ktoso (approved):
Very nice, might be worth adding a test in java already, even tho most fail on CI they should already be working fine on macOS 😊

DougGregor:
Tacked on the real API I wanted, which produces a memory layout for a given Swift type

ktoso (approved):
Very nice, that’ll get us quite far already 🎉

DougGregor:
We're still working on the Java part of the CI; verified it locally.

PR #45, opened 2024-10-07
build: require swift 6 toolchain

ktoso:
Since we require swift 6 anyway, we should reflect this in the
package.swift so it's more obvious when using an older swift to attempt
to build.


ktoso:
Since we require Swift 6 anyway, make it so explicitly in the build.

PR #49, opened 2024-10-07
Improve lookup of JavaClass instances in an environment

DougGregor:
Teach the low-level getJNIClass API to properly check for Java exceptions and turn them into Swift errors. Then add a higher-level API `JavaClass.init(in:)` to look up a specific imported Java class within the given environment.


ktoso (commented):
Lgtm

Issue #12, opened 2024-09-28
Bring more clarity on JDK version requirements


lokesh-tr:
Currently the `jextract-swift` section of the `USER_GUIDE` tells us that it needs JDK 22 to utilise "Project Panama".

Upon further inspection, I realised that the `Java2Swift` Program doesn't compile at all if the JDK version is lesser than 21. This is due to the fact that we hardcoded the `vmArgs.version` to `JNI_VERSION_21` while instantiating a `JavaVirtualMachine` object.

This `JNI_VERSION_21` is a preprocessor constant which is only available in JDK 21 or later.

If this is the intended behaviour, It will be better to update the `USER_GUIDE` to let people know that `Java2Swift` will need a JDK that is 21 or later. However, if this is not the intended behaviour, then replacing the hardcoded JNI version with a JNI version which corresponds to the user's JDK would be better.

-----
I'm not an expert in this context, but making an assumption that a lot of current enterprises (who would like to move their tech stack from JVM to Swift) need not be running their software in JDK 21 or later would be more realistic. Thus, I believe trying to support older JDKs instead of supporting only JDK 21 or later (unless mandatory such as "Project Panama") will be better for this project. Feel free to correct me if I'm wrong.

ktoso:
This is documented more in https://github.com/swiftlang/swift-java/pull/14

Long story short: we'll support both eventually, 22+ with the panama mode, and old versions with just JNI. JavaKit is just JNI and would work with old Java versions.

We need to make the build "smart enough" to survive this eventually.

Today, for ease of development / bootstrapping we're just requiring the latest supported JDK which has Panama.
In the end the required JDK will depend on which feature you're intending to use.

We're not at a phase of "cleanup and publishing" yet, so yes, for now we're just developing with the latest JDK.

lokesh-tr:
> Long story short: we'll support both eventually, 22+ with the panama mode, and old versions with just JNI. JavaKit is just JNI and would work with old Java versions.

I expected this would be the case as well.

> We're not at a phase of "cleanup and publishing" yet, so yes, for now we're just developing with the latest JDK.

I totally understand that, I just think that eventhough we are in early stages of development, the project should be such that, someone can follow the instructions in the Guide and be like "wow, this works" and the JDK requirements not being specified beforehand may not let them reach that point.

[...]

PR #50, opened 2024-10-07
Improve and document the JavaVirtualMachine initializer

DougGregor:
Add options for the class path (so clients don't need to re-learn how to
spell it) and for telling the JVM whether to ignore unrecognized arguments.

While here, reduce the JNI requirements down to JDK/JRE 1.6. We don't use any of the newer APIs. This addresses the functional part of https://github.com/swiftlang/swift-java/issues/12


ktoso (commented):
LGTM :)

PR #52, opened 2024-10-08
Reorganize build: SwiftKit has "swift kit" core, and SwiftKitExample

ktoso:
Further progress on cleaning up the build. We're still having issues with various runtime issues but this gets us to a more understandable separation of "what is a sample" and "what is the library". There's more to be done before we can kill off `make` but this does quite some progress.

- Sample apps "anything with a main()" have their own place; it used to be very mixed up in the same target; together with the example lib
    - `JavaKitSampleApp` there's a sample for javakit which uses the `JavaKitExample` defined swift type
    - `SwiftKitSampleApp` the jextract demo app, uses the `ExampleSwiftLibrary`
    - those JavaKitExample and ExampleSwiftLibrary could use further moving out of `Sources/...` as they're example stuff, not "the library" but I leave at this for now...

Running samples: ./gradlew Samples:SwiftKitSampleApp:run (Samples:Name:run)

- `./gradlew test` does the jextract and whatever the sample apps need
- make is still around; didn't hook up any of the javakit generation
- generated sources no longer in wrong directory; `src/generated/`... as the usual convention in Java land (and not committed)

Screenshot 2024-10-08 at 13 54 30

Tests status:
- macOS
  - gradle test -- OK
  - swift test -- OK
- linux
  - gradle test -- OK
  - swift test -- OK 

but there's some disabled tests on either platforms... We really should get nightly toolchains so we can kill off the `nm` things with https://github.com/swiftlang/swift-java/pull/27 and I'm working to figure that out.


ktoso:
With disabled tests per platform challenges, but at least we have our first green CI run 🥳

PR #55, opened 2024-10-09
Add Fields to Java2Swift Tool

### LINKED ISSUE #22: Generate `@JavaField` in Swift for the public fields of a Java class
When generating Swift wrappers for Java classes (`Java2Swift` direction), we currently ignore the fields. Instead, we should generate `@JavaField` members so that fields can be accessed directly. For example, a Java class like this:

```java
public class MyClass {
  public int value;
}
```

could be generated as:

```swift
@JavaClass("org.mine.MyClass")
struct MyClass {
  @JavaField
  var value: Int32
}
```

jrosen081:
Closes #22 

This PR adds support for exporting java fields into swift. The way I was able to test this was applying the following diff and then running the following commands (I can also add automated tests for this, but wasn't sure what the plan for automated tests for this tool are):
```diff
--- a/Samples/JavaKitSampleApp/src/main/java/com/example/swift/HelloSubclass.java
+++ b/Samples/JavaKitSampleApp/src/main/java/com/example/swift/HelloSubclass.java
@@ -15,7 +15,7 @@
 package com.example.swift;
 
 public class HelloSubclass extends HelloSwift {
-    private String greeting;
+    public String greeting;
 
     public HelloSubclass(String greeting) {
         this.greeting = greeting;
diff --git a/Samples/JavaKitSampleApp/src/main/java/com/example/swift/HelloSwift.java b/Samples/JavaKitSampleApp/src/main/java/com/example/swift/HelloSwift.java
index b4c87f7..3c89e2c 100644
--- a/Samples/JavaKitSampleApp/src/main/java/com/example/swift/HelloSwift.java
+++ b/Samples/JavaKitSampleApp/src/main/java/com/example/swift/HelloSwift.java
@@ -17,7 +17,8 @@ package com.example.swift;
 public class HelloSwift {
     private double value;
     private static double initialValue = 3.14159;
-    private String name = "Java";
+    public String name = "Java";
+    public static String staticName = "Static Java";
 
     static {
         System.loadLibrary("JavaKitExample");
```

And running the following commands from the base folder of the repo:
```
cd Samples/JavaKitSampleApp/src/main/java/com/example/swift/
javac -d ./build *.java
cd build/
jar -cf jar.jar *
swift run Java2Swift com.example.swift.HelloSubclass -o ./ --module-name Java2SwiftTest
cat HelloSubclass.swift
```

This prints out the following swift code (notice the 2 `@JavaField` annotations and not the static field, yet)
```swift
// Auto-generated by Java-to-Swift wrapper generator.
import JavaKit
import JavaRuntime

@JavaClass("com.example.swift.HelloSubclass")
public struct HelloSubclass {
  @JavaField
  public var greeting: String

  @JavaField
  public var name: String

  @JavaMethod
  public init(_ arg0: String, environment: JNIEnvironment)

  @JavaMethod
  public func greet(_ arg0: String)

  @JavaMethod
  public func sayHello(_ arg0: Int32, _ arg1: Int32) -> Int32

  @JavaMethod
  public func throwMessage(_ arg0: String) throws

  @JavaMethod
  public func throwMessageFromSwift(_ arg0: String) throws -> String

  @JavaMethod
  public func toString() -> String

  @JavaMethod
  public func hashCode() -> Int32

  @JavaMethod
  public func notify()

  @JavaMethod
  public func notifyAll()

  @JavaMethod
  public func wait(_ arg0: Int64) throws

  @JavaMethod
  public func wait(_ arg0: Int64, _ arg1: Int32) throws

  @JavaMethod
  public func wait() throws
}
```


Sources/Java2Swift/JavaTranslator.swift
```diff
@@ -422,6 +444,15 @@ extension JavaTranslator {
       public func \(raw: javaMethod.getName())\(raw: genericParameterClause)(\(raw: parametersStr))\(raw: throwsStr)\(raw: resultTypeStr)\(raw: whereClause)
       """
   }
+    
+    func translateField(_ javaField: Field) throws -> DeclSyntax {
```
DougGregor: This looks quite straightforward. Can you de-indent to the 2-space indentation used elsewhere in the file?

DougGregor (approved):
This looks great. The only real missing piece is how we test it... the testing of the Java2Swift path is not great right now, since we basically rely on regenerating its own sources (and not breaking) as "proof" that it didn't break. And unfortunately, none of the classes from the JDK that we build wrappers for include any public fields, so we don't see the effect of this change anywhere in the diff.

Sources/Java2Swift/JavaTranslator.swift
ktoso: FYI: I'll enable formatting validation in CI soon, just that it'll be one massive commit with a bunch of formatting when I do 😓 😅  Holding that off until we land the `nm` removal.

// yeah formatting fix would be nice, but we'll reformat soon.

ktoso (approved):
LGTM, though we won't have CI test this change tbh. Building Java2Swift (`make`) fails right now afair, but it's a nice step.

Sources/Java2Swift/JavaTranslator.swift
jrosen081: formatting updated!

DougGregor:
I just had a wild/willy idea for unit testing this. What if we hook into 

  https://docs.oracle.com/javase/6/docs/api/javax/tools/JavaCompiler.html

in the unit tests so we can compile some Java code expressed as a string literal in Swift, then send that through Java2Swift and "diff" the expected output? If we wire it up correctly, it should make unit-testing changes to Java2Swift straightforward.

ktoso:
Sure, why not I guess, anything that gets us some test coverage tbh :)

PR #57, opened 2024-10-10
Add a test for macro expansion of @JavaClass/@JavaMethod/@JavaField

DougGregor:
Having a unit test here makes iteration faster


ktoso (approved):
Nice, good to start having tests for these, LGTM :)

PR #60, opened 2024-10-10
Fix build with docker issues

### LINKED ISSUE #59: Dockerfile has some issues
I opened this issue to state a few issues I encountered with the Dockerfile and how I fixed them. 

first, `./install_untested_nightly_swift.sh` file name is written wrong on [line 26](https://github.com/swiftlang/swift-java/blob/edee32eec3aae1ef4e727f63255f5ed3311cdefa/docker/Dockerfile#L26C50-L26C51)

After I fixed this it was still not working then I added `COPY install_untested_nightly_swift.sh .` to copy it first. Also the script used `curl` so I added that.

Now the Dockerfile was fixed but the `docker-compose.yaml` file had an issue with the file path

```sh
 + /jdk/bin/java -version
/bin/bash: line 1: /jdk/bin/java: No such file or directory
```
so I changed every `/jdk/bin/java -version` with `java -version`. by using `which java` the bin is actually inside 
`/usr/lib/jvm/default-jdk/bin/java` but I though `java -version` was better.

finally I changed `JAVA_HOME=/jdk` to `JAVA_HOME: /usr/lib/jvm/default-jdk`

AbdAlRahmanGad:
- fix `install_untested_nightly_swift.sh` name in Dockerfile
- add `curl` to the Dockerfile
- fix `java` and `JAVA_HOME` path

closes #59

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,7 +10,8 @@ ARG ubuntu_version
 RUN apt-get update && apt-get install -y \
     locales locales-all \
     make \
-    libc6-dev
+    libc6-dev \
+    curl
 ENV LC_ALL=en_US.UTF-8
 ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US.UTF-8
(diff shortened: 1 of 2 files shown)


AbdAlRahmanGad:
Hi @ktoso, 
Could you please review this patch when your time permits?

ktoso (approved):
Good cleanup thank you!

AbdAlRahmanGad:
You're welcome! Glad I could help.

PR #61, opened 2024-10-10
Add Support for Static Fields to Java2Swift tool

### LINKED ISSUE #39: JavaKit support for static fields (wrapping Java types in Swift)
Right now, there is no way to express a static field in JavaKit. It should probably be represented as

```swift
extension JavaClass<TheSwiftType> {
  @JavaStaticField
  var theStaticField: Int32
}
```

We need to both introduce this macro (like with `@JavaStaticMethod`) and update the JavaToSwift tool to walk static fields of classes to emit these in the `JavaClass` extension, like it currently does for static methods.

jrosen081:
Closes #39 

This adds support to adding static fields to the Java2Swift tool. The issue recommends adding a `@JavaStaticField` macro, but it looks like due to the fact that the subscript created by the `JavaField` macro works for static properties when defined on a `JavaClass`, so I'm not sure a `@JavaStaticField` really buys anything here, so I'm open to removing it.

Same testing as in #55 (since that one also adds a static String).

I also noticed that we had a `staticMethods` array, but were re-requesting all methods in the `Java2Swift` tool when handling static methods, so I updated it to just use the `staticMethods` array to simplify the code a bit.

### DIFF
--- a/Sources/Java2Swift/JavaTranslator.swift
+++ b/Sources/Java2Swift/JavaTranslator.swift
@@ -347,23 +347,28 @@ extension JavaTranslator {
 
     // Format the class declaration.
     classDecl = classDecl.formatted(using: format).cast(DeclSyntax.self)
-      
-    // TODO: Handle static fields in https://github.com/swiftlang/swift-java/issues/39
-
-    if staticMethods.isEmpty {
+    
+    if staticMethods.isEmpty && staticFields.isEmpty {
       return [classDecl]
     }
 
     // Translate static members.
     var staticMembers: [DeclSyntax] = []
-    staticMembers.append(
-      contentsOf: javaClass.getMethods().compactMap {
-        $0.flatMap { method in
-          // Skip the instance methods; they were handled above.
-          if !method.isStatic {
-            return nil
-          }
 
+    staticMembers.append(
+      contentsOf: staticFields.compactMap { field in
+        // Translate each static field.
+        do {
+          return try translateField(field)
+        } catch {
+          logUntranslated("Unable to translate '\(fullName)' field '\(field.getName())': \(error)")
+          return nil
+        }
+      }
+    )
+    
+    staticMembers.append(
+      contentsOf: staticMethods.compactMap { method in
           // Translate each static method.
           do {
             return try translateMethod(
@@ -376,7 +381,6 @@ extension JavaTranslator {
             return nil
           }
         }
-      }
     )
 
     if staticMembers.isEmpty {
@@ -447,7 +451,7 @@ extension JavaTranslator {
     
   func translateField(_ javaField: Field) throws -> DeclSyntax {
     let typeName = try getSwiftTypeNameAsString(javaField.getGenericType()!, outerOptional: true)
-    let fieldAttribute: AttributeSyntax = "@JavaField";
+    let fieldAttribute: AttributeSyntax = javaField.isStatic ? "@JavaStaticField" : "@JavaField";
     return """
       \(fieldAttribute)
       public var \(raw: javaField.getName()): \(raw: typeName)
--- a/Sources/Java2Swift/JavaTranslator.swift
+++ b/Sources/Java2Swift/JavaTranslator.swift
@@ -257,7 +257,7 @@ extension JavaTranslator {
           do {
             return try translateField(field)
           } catch {
-            logUntranslated("Unable to translate '\(fullName)' field '\(field.getName())': \(error)")
+            logUntranslated("Unable to translate '\(fullName)' static field '\(field.getName())': \(error)")
             return nil
           }
         }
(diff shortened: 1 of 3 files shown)


ktoso (approved):
Looks good to me but I'll give @DougGregor a chance to answer the question asked 👍

Sources/JavaKit/Macros.swift
```diff
@@ -93,6 +93,21 @@ public macro JavaInterface(_ fullClassName: String, extends: (any AnyJavaObject.
...
 
+
+/// Attached macro that turns a Swift property into one that accesses a Java static field on the underlying Java object.
+///
+/// The macro must be used within a specific JavaClass instance.
+///
+/// ```swift
+/// @JavaClass("org.swift.example.HelloSwift")
+/// struct HelloSwift {
+///     @JavaField
```
ktoso:
```suggestion
///     @JavaStaticField
```

shouldn't the `var` be `static` as well?

Sources/JavaKitExample/JavaKitExample.swift
```diff
@@ -109,12 +109,12 @@ extension JavaClass<HelloSwift> {
 
 @JavaClass("com.example.swift.HelloSubclass", extends: HelloSwift.self)
 struct HelloSubclass {
-  @JavaField
+  @JavaStaticField
   var greeting: String
```
ktoso: shouldn't we add another static field rather than modify this one which covers a normal property?

Sources/JavaKit/Macros.swift
jrosen081: Oops this was a typo, this needs to be on the `JavaClass<HelloSwift>`

Sources/JavaKitExample/JavaKitExample.swift
jrosen081: Same issue, I put this on the wrong type (but this kind of shows why I think `JavaStaticField` is not needed since this worked when placed on a non static field)

Sources/JavaKit/Macros.swift
jrosen081: updated

Sources/JavaKitExample/JavaKitExample.swift
```diff
@@ -103,7 +103,7 @@ struct HelloSwift {
 }
 
 extension JavaClass<HelloSwift> {
-  @JavaField
+  @JavaStaticField
   var initialValue: Double
```
ktoso, line 107: should this be a static var?

jrosen081: It's an instance in JavaClass similar to how static methods are instance methods on their Java class. I was thinking of how to get this be a static var, but everything needs the JNIEnvironment (from my understanding), which is an instance var on all Java structs

DougGregor: Right. Java static methods/vars become non-static members on the appropriately-specialized `JavaClass` instance.

Sources/Java2Swift/JavaTranslator.swift
```diff
@@ -347,23 +347,28 @@ extension JavaTranslator {
...
-            return nil
-          }
 
+    staticMembers.append(
+      contentsOf: staticFields.compactMap { field in
+        // Translate each static field.
+        do {
+          return try translateField(field)
+        } catch {
+          logUntranslated("Unable to translate '\(fullName)' field '\(field.getName())': \(error)")
```
DougGregor, line 364: Please add "static" into the untranslatedLog message so we know it's a static field

Sources/JavaKitExample/JavaKitExample.swift
DougGregor: For bonus points, one could have the `@JavaStaticField` macro detect the `static` and produce an error diagnosing the problem, but it's not really necessary. Folks shouldn't have to write these macro uses themselves.

DougGregor (approved):
This looks good to me, with the one minor request to the log message to make it easier to figure out when static fields aren't getting translated.

Sources/JavaKitExample/JavaKitExample.swift
jrosen081: Hmmm do we have access to the surrounding context of a macro? I know you did some work there, but I lost where that ended up.

Sources/Java2Swift/JavaTranslator.swift
jrosen081: updated!

Sources/JavaKitExample/JavaKitExample.swift
DougGregor: Yeah, the [`lexicalContext`](https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftSyntaxMacros/MacroExpansionContext.swift#L57-L72) of the macro captures this information. To be very clear, I don't think this diagnostic is something that should block your PR from being merged. I just wanted to note that we can implement these kinds of checks  in the macro implementation if we want.

ktoso:
Nice thank you!

ktoso:
Issue for the follow up : https://github.com/swiftlang/swift-java/issues/65

DougGregor:
https://github.com/swiftlang/swift-java/pull/73 makes it possible to write unit tests for Java2Swift, so long as you can find some class in the JDK that has the kind of thing you want to test.

Issue #43, opened 2024-10-06
swift test crashes on macOS command line


helje5:
```
helge@M2ni swift-java % swift test
Building for debugging...
[21/21] Linking JavaKitPackageTests
Build complete! (4.17s)
...
Test Suite 'All tests' started at 2024-10-06 15:04:37.417.
Test Suite 'All tests' passed at 2024-10-06 15:04:37.419.
	 Executed 0 tests, with 0 failures (0 unexpected) in 0.000 (0.001) seconds
􀟈  Test run started.
􀄵  Testing Library Version: 94 (arm64e-apple-macos13.0)
􀟈  Suite "Nominal type lookup" started.
...
[info][JExtractSwift/Swift2JavaVisitor.swift:53](visitPost(_:)) Completed import: classDecl MySwiftClass  
[info][JExtractSwift/Swift2JavaVisitor.swift:132](visit(_:)) Record method in MySwiftClass 
[info][JExtractSwift/Swift2JavaTranslator.swift:113](postProcessImportedDecls()) Post process imported decls... ["global/funcs": "4", "types": "1"]
[warning][JExtractSwift/Swift2JavaTranslator.swift:124](postProcessImportedDecls()) Unable to find mangled names for imported symbols. Dylib not found: .build/arm64-apple-macosx/debug/libSwiftModule.dylib This method of obtaining symbols is a workaround; it will be removed. 
...
[info][JExtractSwift/Swift2JavaVisitor.swift:53](visitPost(_:)) Completed import: classDecl MySwiftClass  
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0xa) at pc=0x0000000108f13948[info][JExtractSwift/Swift2JavaTranslator.swift:113](postProcessImportedDecls()) Post process imported decls... ["global/funcs": "4", "types": "1"]
, pid=79830, tid=259
#
# JRE version:  (23.0+37) (build )
# Java VM: OpenJDK 64-Bit Server VM (23+37-2369, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# [warning][JExtractSwift/Swift2JavaTranslator.swift:124](postProcessImportedDecls()) Unable to find mangled names for imported symbols. Dylib not found: .build/arm64-apple-macosx/debug/libSwiftModule.dylib This method of obtaining symbols is a workaround; it will be removed. 
[...]

PR #64, opened 2024-10-10
Temporarily switch JavaKitTests target to XCTest

DougGregor:
Command-line testing on macOS is broken for the JavaKitTests due to an interaction between the Java Virtual Machine, a helper test process for swift-testing (`swiftpm-testing-helper`), and some macOS security features. Work around the issue by replacing our use of swift-testing with XCTest, whose helper process doesn't seem to have this problem. Only do so for test targets that need to spin up an instance of the JVM, and keep using swift-testing everywhere we can.

This makes me very sad, but is better than the current solution of disabling these tests when running from the terminal. Once we've found a solution for the underlying issue, we'll revert this change as get back to swift-testing and its much nicer testing output.

Fixes issue https://github.com/swiftlang/swift-java/issues/43 without disabling tests.

### DIFF
--- a/Tests/JExtractSwiftTests/VariableImportTests.swift
+++ b/Tests/JExtractSwiftTests/VariableImportTests.swift
@@ -41,7 +41,7 @@ final class VariableImportTests {
     )
     st.log.logLevel = .error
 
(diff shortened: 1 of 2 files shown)


ktoso:
Huh surprising interaction! Lgtm thanks

PR #69, opened 2024-10-11
Support concurrent use of a shared `JavaVirtualMachine`

DougGregor:
The `JavaVirtualMachine` class was not at all designed with concurrency in mind: it always created a new Java VM under the hood, and was associated with a single JNI environment even though different threads could have different JNI environments.

Move to a different model with a `static shared()` method to create or access a shared `JavaVirtualMachine` instance to encapsulate the (singleton) JVM. If there is an existing JVM, whether created via this Swift library or from some other code, it will be adopted by the shared `JavaVirtualMachine` instance.

Additionally, the `JavaVirtualMachine` no longer stores a JNI environment. Instead, it now provides an `environment` function that queries the JVM to get the JNI environment for the current thread. If the current thread isn't attached to the JVM, it does so and provides the resulting JNI environment.

With all of these in place, we no longer need to force all of our JavaKit tests over to the main actor.


DougGregor:
Merging because I'm building other things on top of this

Issue #68, opened 2024-10-11
Error when running `make`: unknown argument: '-abi-comments-in-module-interface'


AbdAlRahmanGad:
when running `make` I get this error 

```
error: emit-module command failed with exit code 1 (use -v to see invocation)
<unknown>:0: error: unknown argument: '-abi-comments-in-module-interface'
<unknown>:0: error: unknown argument: '-abi-comments-in-module-interface'
make: *** [Makefile:116: generate-JExtract-interface-files] Error 1
```
swift --version
```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```
MacBook Air M1, MacOS 15.0.1

full log:

[log.txt](https://github.com/user-attachments/files/17346450/log.txt)

DougGregor:
You'll need a development toolchain for jextract-swift. Please see the section [Required Swift development toolchains](https://github.com/swiftlang/swift-java/?tab=readme-ov-file#required-swift-development-toolchains) in the README.

DougGregor:
Actually, I'm going to split the makefile targets. The JavaKit direction (using Java from Swift) absolutely does not need the new compiler, and the makefile shouldn't fail. Sorry 'bout that!

DougGregor:
Now it's better, thanks for your patience.

AbdAlRahmanGad:
Thanks for the quick response!

PR #71, opened 2024-10-11
Rework makefile to split JavaKit targets from jextract-swift targets

DougGregor:
The JavaKit targets (Swift calling into Java via JNI) have fewer requirements (for JVM installations and development toolchains) and are a bit easier to get started with. Split those targets out, and turn `make` with no specified target into a little help target.

Fixes issue #68.


ktoso (commented):
Lgtm :)

PR #72, opened 2024-10-11
Regenerate JavaKit Swift sources now that we have support for static fields

DougGregor:


DougGregor:
@jrosen081 FYI, this is the result of re-running the JavaKit generation now that your static field support has landed.

jrosen081 (approved):
This is awesome! Is the plan to always commit these every time we add more support or in the future have the first step to using this be run the `make` (or Gradle when the Makefile gets removed)?

DougGregor:
> This is awesome! Is the plan to always commit these every time we add more support or in the future have the first step to using this be run the `make` (or Gradle when the Makefile gets removed)?

We should commit these every time we add more support, since we need them there to build the Java2Swift tool itself. It's also a nice validation step.

PR #73, opened 2024-10-11
Split the Java2Swift tool out into a library target and add a test harness

DougGregor:
Introduce a test harness so we can run the Java-to-Swift API translator against a Java class and then check the generated syntax. We could definitely do better with the diff'ing of lines to make it a little easier to use, but it's better than no testing.


Package.swift
```diff
@@ -230,13 +230,12 @@ let package = Package(
       ]
     ),
 
-    .executableTarget(
-      name: "Java2Swift",
+    .target(
+      name: "Java2SwiftLib",
```
ktoso, line 234: Not sure about the “lib” suffix tbh, just java2swift is fine

ktoso (approved):
Lgtm tho the lib suffix in target name is weird imho, id leave it off

Package.swift
DougGregor: I'd have to rename the executable target to drop the Lib suffix

PR #74, opened 2024-10-12
java2swift: add non-optional-returning `as` function to `JavaClassMacro`

### LINKED ISSUE #20: Create non-optional-returning `as` functions for superclasses/extended interfaces
`AnyJavaObject` provides [an `as` method](https://github.com/swiftlang/swift-java/blob/main/Sources/JavaKit/JavaObject%2BInheritance.swift#L51) that performs a dynamic cast to another `AnyJavaObject-`conforming type, then returns an optional with either the casted type or `nil` to indicate failure.

When casting to the superclass type or the type of one of the interfaces that a type extends, the result will never be `nil`. Developers performing these upcasts will have to force-unwrap these results. For example:

```swift
let throwable: Throwable = ...
let object: JavaObject = throwable.as(JavaObject.self)!
```

As part of generating the [`Throwable`](https://github.com/swiftlang/swift-java/blob/main/Sources/JavaKit/generated/Throwable.swift) Swift type, we could provide non-optional-returning overloads of `as`, e.g.,

```swift
@JavaClass("java.lang.Throwable")
struct Throwable {
  func as(_: JavaObject.Type) -> JavaObject { ... }
}
```

AbdAlRahmanGad:
closes #20

### DIFF
--- a/Sources/JavaKitMacros/JavaClassMacro.swift
+++ b/Sources/JavaKitMacros/JavaClassMacro.swift
@@ -88,9 +88,9 @@ extension JavaClassMacro: MemberMacro {
       """
 
     let nonOptionalAs: DeclSyntax = """
-      /// It's not checking anything.
-      public func `as`<OtherClass: AnyJavaObject>(_: OtherClass.Type) -> OtherClass {
-          return OtherClass(javaHolder: javaHolder)
+      /// Casting to <\(raw: superclass)> will never be nil because <\(raw: className.split(separator: ".").last!)> extends it.
+      public func `as`(_: \(raw: superclass)) -> \(raw: superclass) {
+          return \(raw: superclass)(javaHolder: javaHolder)
       }
       """
 
--- a/Sources/JavaKitMacros/JavaClassMacro.swift
+++ b/Sources/JavaKitMacros/JavaClassMacro.swift
@@ -88,6 +88,7 @@ extension JavaClassMacro: MemberMacro {
       """
 
     let nonOptionalAs: DeclSyntax = """
+      /// It's not checking anything.
       public func `as`<OtherClass: AnyJavaObject>(_: OtherClass.Type) -> OtherClass {
           return OtherClass(javaHolder: javaHolder)
       }
--- a/Sources/JavaKitMacros/JavaClassMacro.swift
+++ b/Sources/JavaKitMacros/JavaClassMacro.swift
@@ -89,7 +89,7 @@ extension JavaClassMacro: MemberMacro {
 
     let nonOptionalAs: DeclSyntax = """
       /// Casting to ``\(raw: superclass)`` will never be nil because ``\(raw: className.split(separator: ".").last!)`` extends it.
-      public func `as`(_: \(raw: superclass).type) -> \(raw: superclass) {
+      public func `as`(_: \(raw: superclass).Type) -> \(raw: superclass) {
           return \(raw: superclass)(javaHolder: javaHolder)
       }
       """
--- a/Sources/JavaKitMacros/JavaClassMacro.swift
+++ b/Sources/JavaKitMacros/JavaClassMacro.swift
@@ -87,6 +87,12 @@ extension JavaClassMacro: MemberMacro {
       }
       """
 
+    let nonOptionalAs: DeclSyntax = """
+      public func `as`<OtherClass: AnyJavaObject>(_: OtherClass.Type) -> OtherClass {
+          return OtherClass(javaHolder: javaHolder)
+      }
+      """
+
     return [
       fullJavaClassNameMember,
       superclassTypealias,
@@ -94,6 +100,7 @@ extension JavaClassMacro: MemberMacro {
       javaThisMember,
       javaEnvironmentMember,
       initMember,
+      nonOptionalAs,
     ]
   }
 }
--- a/Sources/JavaKitMacros/JavaClassMacro.swift
+++ b/Sources/JavaKitMacros/JavaClassMacro.swift
@@ -88,8 +88,8 @@ extension JavaClassMacro: MemberMacro {
       """
 
     let nonOptionalAs: DeclSyntax = """
-      /// Casting to <\(raw: superclass)> will never be nil because <\(raw: className.split(separator: ".").last!)> extends it.
-      public func `as`(_: \(raw: superclass)) -> \(raw: superclass) {
+      /// Casting to ``\(raw: superclass)`` will never be nil because ``\(raw: className.split(separator: ".").last!)`` extends it.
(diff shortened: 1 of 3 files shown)


Tests/JavaKitMacroTests/JavaClassMacroTests.swift
```diff
@@ -78,6 +78,10 @@ class JavaKitMacroTests: XCTestCase {
             public init(javaHolder: JavaObjectHolder) {
                 self.javaHolder = javaHolder
             }
+
+            public func `as`<OtherClass: AnyJavaObject>(_: OtherClass.Type) -> OtherClass {
```
ktoso: Can we add some docs on the semantics i.e. that it's not checking anything.

ktoso (approved):

Sources/JavaKitMacros/JavaClassMacro.swift
```diff
@@ -87,13 +87,21 @@ extension JavaClassMacro: MemberMacro {
       }
       """
 
+    let nonOptionalAs: DeclSyntax = """
+      /// It's not checking anything.
+      public func `as`<OtherClass: AnyJavaObject>(_: OtherClass.Type) -> OtherClass {
```
ktoso: On second look, this isn't right I think?

Only when casting to interfaces that __we know__ this type implements should this be the non-optional returning one. As implemented right now this just converts "to anything" which isn't safe.

Sources/JavaKitMacros/JavaClassMacro.swift
```diff
@@ -87,13 +87,21 @@ extension JavaClassMacro: MemberMacro {
       }
       """
 
+    let nonOptionalAs: DeclSyntax = """
+      /// It's not checking anything.
```
ktoso: This isn't quite enough.

This should document why the conversion will never fail and the implementation should do so accordingly.

```suggestion
    // Casting to <. the target type .> will never be nil because <. the java type .> <<extends it || or some other reason>>.


```

ktoso (changes requested):

ktoso:
Gave this another look, it's not right at all I think. Only the _known to be safe_ target types should be non-optional. So we need to know which types the java type extends/implements and __for those types__ generate the method.

Sources/JavaKitMacros/JavaClassMacro.swift
AbdAlRahmanGad: So instead of `AnyJavaObject` I should use `\(raw: superclass)`?
Or is there a better way to do it?

ktoso: Yeah, it's about the super class. Idk what about interfaces e.g. if they have default implementations, if such methods are already callable on the subtype in swift -- we can handle this in a separate PR though

Sources/JavaKitMacros/JavaClassMacro.swift
```diff
@@ -87,13 +87,21 @@ extension JavaClassMacro: MemberMacro {
       }
       """
 
+    let nonOptionalAs: DeclSyntax = """
+      /// Casting to <\(raw: superclass)> will never be nil because <\(raw: className.split(separator: ".").last!)> extends it.
+      public func `as`(_: \(raw: superclass)) -> \(raw: superclass) {
```
DougGregor: With the `as` API, one always passes a metatype, e.g,

```swift
obj.as(MyJavaClass.self)
```

So I think the type of the parameter to `as` needs to be `\(raw: superclass).Type` .

Tests/JavaKitMacroTests/JavaClassMacroTests.swift
```diff
@@ -78,6 +78,11 @@ class JavaKitMacroTests: XCTestCase {
             public init(javaHolder: JavaObjectHolder) {
                 self.javaHolder = javaHolder
             }
+
+            /// Casting to <JavaObject> will never be nil because <HelloWorld> extends it.
```
ktoso:
```suggestion
            /// Casting to ``JavaObject`` will never be nil because ``HelloWorld`` extends it.
```

will actually result in proper type links, see here: https://www.swift.org/documentation/docc/linking-to-symbols-and-other-content

Tests/JavaKitMacroTests/JavaClassMacroTests.swift
```diff
@@ -78,6 +78,11 @@ class JavaKitMacroTests: XCTestCase {
             public init(javaHolder: JavaObjectHolder) {
                 self.javaHolder = javaHolder
             }
+
+            /// Casting to ``JavaObject`` will never be nil because ``HelloWorld`` extends it.
+            public func `as`(_: JavaObject.type) -> JavaObject {
```
ktoso:
```suggestion
            public func `as`(_: JavaObject.Type) -> JavaObject {
```

Almost, in Swift this has to be `.Type` (uppercase).

Once that's fixed tests will pass as well so I'll take that edit

Sources/JavaKitMacros/JavaClassMacro.swift
```diff
@@ -87,13 +87,21 @@ extension JavaClassMacro: MemberMacro {
       }
       """
 
+    let nonOptionalAs: DeclSyntax = """
+      /// Casting to ``\(raw: superclass)`` will never be nil because ``\(raw: className.split(separator: ".").last!)`` extends it.
+      public func `as`(_: \(raw: superclass).type) -> \(raw: superclass) {
```
ktoso:
```suggestion
      public func `as`(_: \(raw: superclass).Type) -> \(raw: superclass) {
```

ktoso:
Seems this actually passes on `swift-PR-76905-1539` but is not happy with the nightly, needs some more work -- would you like to continue polishing this up so it passes CI @AbdAlRahmanGad ?

AbdAlRahmanGad:
> Seems this actually passes on `swift-PR-76905-1539` but is not happy with the nightly, needs some more work -- would you like to continue polishing this up so it passes CI @AbdAlRahmanGad ?

Of course!
I just read the failing tests and looks like it's because I need to chaange `as` to \``as`\` in `Macros.swift`

ktoso (approved):

ktoso:
Cool, this is looking good now.

I'd love to see perhaps another PR where we take some type that inherits from some other `class Alpha extends Base` type and make sure that we get the `as(Base.self)` method as well.

ktoso:
Thanks for your contrib!

AbdAlRahmanGad:
> Cool, this is looking good now.
> 
> I'd love to see perhaps another PR where we take some type that inherits from some other `class Alpha extends Base` type and make sure that we get the `as(Base.self)` method as well.

Could you please Guide me to where I should implement this?
Also, Is this to make sure our implementation works right? So we are going to add an example to test it?

PR #76, opened 2024-10-14
java2swift: Add Support for Enum Cases

### LINKED ISSUE #21: Implement Swift wrappers for Java enum types
Java [enum types](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) are encoded as Java classes with various constants for their runtime value. At present, Java2Swift imports these just like any other Java class.

We should try to produce Swift wrappers that either are enums or feel like enums, with support for switches and easy identification of cases.

jrosen081:
Closes #21 
Adds a few nice pieces of support for enums:
1. Adds static properties on the main struct so we can have the static member dot syntax (i.e. `.APRIL` for april static property on Month). 

There are a few things we could improve here related to the static properties that I'm interested to see if we should update:
1. Update the SNAKE_CASE static properties to lower camel case to be more _swifty_
2. Right now to get the environment we have a `try!` which is likely not great. If the programmer already is using the `java-swift` work, they likely have the JVM all set up, so this should be fine, but if not, we might want to remove the static properties or come up with another way of creating the environment statically that is not failing (if that is even possible)
3. We also need to add a force unwrap here since the java field won't compile with a non-optional value

Open Question the I'm unsure of:
- [ ] Should we add an `Equatable` conformance for these using the `equals` function, so that we can get a good implementation of `~=` for switch statements?
- [ ] Should we add an `~=` implementation itself (possibly also using `equals`)?

### DIFF
--- a/Sources/Java2SwiftLib/JavaTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator.swift
@@ -360,7 +360,6 @@ extension JavaTranslator {
       contentsOf: staticFields.compactMap { field in
         // Translate each static field.
         do {
-          // Enum constants are guaranteed to not be optional
           return try translateField(field)
         } catch {
           logUntranslated("Unable to translate '\(fullName)' field '\(field.getName())': \(error)")
@@ -470,10 +469,11 @@ extension JavaTranslator {
 
     let mappingSyntax: DeclSyntax = """
       public var enumValue: \(raw: name)? {
+        let classObj = self.javaClass
         \(raw: enumFields.map {
           // The equals method takes a java object, so we need to cast it here
           """
-          if self.equals(self.javaClass.\($0.getName())?.as(JavaObject.self)) {
+          if self.equals(classObj.\($0.getName())?.as(JavaObject.self)) {
                 return \(name).\($0.getName())
           }
           """
@@ -483,7 +483,26 @@ extension JavaTranslator {
       }
     """
 
-    return [extensionSyntax, mappingSyntax]
+    let initSyntax: DeclSyntax = """
+    public init?(_ enumValue: \(raw: name), environment: JNIEnvironment) throws {
+      let classObj = try JavaClass<Self>(in: environment)
+      switch enumValue {
+    \(raw: enumFields.map {
+      let caseName = $0.getName()
+      return """
+          case .\(caseName):
+            if let \(caseName) = classObj.\(caseName) {
+              self = \(caseName)
+            } else {
+              return nil
+            }
+      """
+    }.joined(separator: "\n"))
+      }
+    }
+    """
+
+    return [extensionSyntax, mappingSyntax, initSyntax]
   }
 
   // Translate a Java parameter list into Swift parameters.
--- a/Sources/Java2SwiftLib/JavaTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator.swift
@@ -239,11 +239,16 @@ extension JavaTranslator {
         $0.flatMap { field in
           if field.isStatic {
             staticFields.append(field)
-            return nil
+
+            if field.isEnumConstant() {
+              return "public static let \(raw: field.getName()) = try! JavaClass<Self>(environment: JavaVirtualMachine.environment()).\(raw: field.getName())"
+            } else {
+              return nil
+            }
           }
           
           do {
-            return try translateField(field)
+            return try translateField(field, isOptional: true)
           } catch {
             logUntranslated("Unable to translate '\(fullName)' static field '\(field.getName())': \(error)")
             return nil
@@ -347,7 +352,8 @@ extension JavaTranslator {
       contentsOf: staticFields.compactMap { field in
         // Translate each static field.
         do {
-          return try translateField(field)
+          // Enum constants are guaranteed to not be optional
+          return try translateField(field, isOptional: !field.isEnumConstant())
         } catch {
           logUntranslated("Unable to translate '\(fullName)' field '\(field.getName())': \(error)")
           return nil
@@ -437,8 +443,8 @@ extension JavaTranslator {
       """
   }
     
-  package func translateField(_ javaField: Field) throws -> DeclSyntax {
-    let typeName = try getSwiftTypeNameAsString(javaField.getGenericType()!, outerOptional: true)
+  package func translateField(_ javaField: Field, isOptional: Bool) throws -> DeclSyntax {
+    let typeName = try getSwiftTypeNameAsString(javaField.getGenericType()!, outerOptional: isOptional)
     let fieldAttribute: AttributeSyntax = javaField.isStatic ? "@JavaStaticField" : "@JavaField";
     let swiftFieldName = javaField.getName().escapedSwiftName
     return """
--- a/Sources/Java2SwiftLib/JavaTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator.swift
@@ -77,6 +77,7 @@ extension JavaTranslator {
   package static let defaultTranslatedClasses: [String: (swiftType: String, swiftModule: String?, isOptional: Bool)] = [
     "java.lang.Class": ("JavaClass", "JavaKit", true),
     "java.lang.String": ("String", "JavaKit", false),
+    "java.lang.Object": ("JavaObject", "JavaKit", true)
   ]
 }
 
@@ -234,6 +235,7 @@ extension JavaTranslator {
     
     // Fields
     var staticFields: [Field] = []
+    var enumConstants: [Field] = []
     members.append(
       contentsOf: javaClass.getFields().compactMap {
         $0.flatMap { field in
@@ -241,10 +243,9 @@ extension JavaTranslator {
             staticFields.append(field)
 
             if field.isEnumConstant() {
-              return "public static let \(raw: field.getName()) = try! JavaClass<Self>(in: JavaVirtualMachine.shared().environment()).\(raw: field.getName())!"
-            } else {
-              return nil
+              enumConstants.append(field)
             }
+            return nil
           }
           
           do {
@@ -256,6 +257,13 @@ extension JavaTranslator {
         }
       }
     )
+
+    if !enumConstants.isEmpty {
+      let enumName = "\(swiftTypeName)Cases"
+      members.append(
+        contentsOf: translateToEnumValue(name: enumName, enumFields: enumConstants)
+      )
+    }
 
     // Constructors
     members.append(
@@ -451,6 +459,31 @@ extension JavaTranslator {
       \(fieldAttribute)
       public var \(raw: swiftFieldName): \(raw: typeName)
       """
+  }
+
+  package func translateToEnumValue(name: String, enumFields: [Field]) -> [DeclSyntax] {
+    let extensionSyntax: DeclSyntax = """
+      public enum \(raw: name): Equatable {
+        \(raw: enumFields.map { "case \($0.getName())" }.joined(separator: "\n"))
+      }
+    """
+
+    let mappingSyntax: DeclSyntax = """
+      public var enumValue: \(raw: name)? {
+        \(raw: enumFields.map {
+          // The equals method takes a java object, so we need to cast it here
+          """
+          if self.equals(self.javaClass.\($0.getName())?.as(JavaObject.self)) {
+                return \(name).\($0.getName())
+          }
+          """
+        }.joined(separator: " else ")) else {
+          return nil
+        }
+      }
+    """
+
+    return [extensionSyntax, mappingSyntax]
   }
 
   // Translate a Java parameter list into Swift parameters.
--- a/Sources/Java2SwiftLib/JavaTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator.swift
@@ -241,14 +241,14 @@ extension JavaTranslator {
             staticFields.append(field)
 
             if field.isEnumConstant() {
-              return "public static let \(raw: field.getName()) = try! JavaClass<Self>(environment: JavaVirtualMachine.environment()).\(raw: field.getName())"
+              return "public static let \(raw: field.getName()) = try! JavaClass<Self>(in: JavaVirtualMachine.shared().environment()).\(raw: field.getName())!"
             } else {
               return nil
             }
           }
           
           do {
-            return try translateField(field, isOptional: true)
+            return try translateField(field)
           } catch {
             logUntranslated("Unable to translate '\(fullName)' static field '\(field.getName())': \(error)")
             return nil
@@ -353,7 +353,7 @@ extension JavaTranslator {
         // Translate each static field.
         do {
           // Enum constants are guaranteed to not be optional
-          return try translateField(field, isOptional: !field.isEnumConstant())
+          return try translateField(field)
         } catch {
           logUntranslated("Unable to translate '\(fullName)' field '\(field.getName())': \(error)")
           return nil
@@ -443,8 +443,8 @@ extension JavaTranslator {
       """
   }
     
-  package func translateField(_ javaField: Field, isOptional: Bool) throws -> DeclSyntax {
-    let typeName = try getSwiftTypeNameAsString(javaField.getGenericType()!, outerOptional: isOptional)
+  package func translateField(_ javaField: Field) throws -> DeclSyntax {
+    let typeName = try getSwiftTypeNameAsString(javaField.getGenericType()!, outerOptional: true)
     let fieldAttribute: AttributeSyntax = javaField.isStatic ? "@JavaStaticField" : "@JavaField";
     let swiftFieldName = javaField.getName().escapedSwiftName
     return """
--- a/Sources/Java2SwiftLib/JavaTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator.swift
@@ -77,7 +77,6 @@ extension JavaTranslator {
   package static let defaultTranslatedClasses: [String: (swiftType: String, swiftModule: String?, isOptional: Bool)] = [
     "java.lang.Class": ("JavaClass", "JavaKit", true),
     "java.lang.String": ("String", "JavaKit", false),
-    "java.lang.Object": ("JavaObject", "JavaKit", true)
   ]
 }
 
--- a/Sources/Java2SwiftLib/JavaTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator.swift
@@ -484,17 +484,16 @@ extension JavaTranslator {
     """
 
     let initSyntax: DeclSyntax = """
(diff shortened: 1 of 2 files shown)


Tests/Java2SwiftTests/Java2SwiftTests.swift
```diff
@@ -84,6 +103,7 @@ func assertTranslatedClass<JavaClassType: AnyJavaObject>(
     """
 
   for expectedChunk in expectedChunks {
+    print(swiftFileText)
```
ktoso: stray print?

jrosen081: oops, it made it much easier for testing that what came out actually built ... but removed it now

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -251,7 +251,12 @@ extension JavaTranslator {
         $0.flatMap { field in
           if field.isStatic {
             staticFields.append(field)
-            return nil
+
+            if field.isEnumConstant() {
+              return "public static let \(raw: field.getName()) = try! JavaClass<Self>(in: JavaVirtualMachine.shared().environment()).\(raw: field.getName())!"
```
DougGregor: If we've going to use `JavaVirtualMachine` here, we need to also make sure to `import JavaKitVM` in the generated code. There's a set of imports in the `JavaTranslator` that you would need to add to.

I do worry a bit about calling `JavaVirtualMachine.shared()` here in something that looks like a constant, and I wonder if we should consider alternative ways to model Java enums.

DougGregor:
@jrosen081 your items (2) and (3) have me wondering if we should try different ways of bringing Java enum constants into Swift. For example, we could collect all of the enum cases and put them into a synthesized Swift enum, e.g.,

```swift
/* assuming this is java.time.Month */
extension Month {
  public enum EnumConstant {
    case JANUARY
    case FEBRUARY
    // ...
  }

  public init(_ constant: EnumConstant, environment: JNIEnvironment) { ... }

  public var enumConstant: EnumConstant { get }
}
```

Then we could add matching operators for switch and such as needed. Perhaps even factor some of these operations out into a protocol to which Java enum classes would be made to conform. We would no longer require the virtual machine instance nor the `try!`.

What do you think? I'm open to other ideas, too!

jrosen081:
> @jrosen081 your items (2) and (3) have me wondering if we should try different ways of bringing Java enum constants into Swift. For example, we could collect all of the enum cases and put them into a synthesized Swift enum, e.g.,
> 
> ```swift
> /* assuming this is java.time.Month */
> extension Month {
>   public enum EnumConstant {
>     case JANUARY
>     case FEBRUARY
>     // ...
>   }
> 
>   public init(_ constant: EnumConstant, environment: JNIEnvironment) { ... }
> 
>   public var enumConstant: EnumConstant { get }
> }
> ```
> 
> Then we could add matching operators for switch and such as needed. Perhaps even factor some of these operations out into a protocol to which Java enum classes would be made to conform. We would no longer require the virtual machine instance nor the `try!`.
> 
> What do you think? I'm open to other ideas, too!

I like it. I think we would need `EnumConstant` to be computed since we might get the value from Java, which would not have a way of creating the synthesized enum in Swift (I don't think). I'll prototype this setup which gives us the benefit of Swift enums without needing to use the environment and force unwrap everywhere

jrosen081:
> > @jrosen081 your items (2) and (3) have me wondering if we should try different ways of bringing Java enum constants into Swift. For example, we could collect all of the enum cases and put them into a synthesized Swift enum, e.g.,
> > ```swift
> > /* assuming this is java.time.Month */
> > extension Month {
> >   public enum EnumConstant {
> >     case JANUARY
> >     case FEBRUARY
> >     // ...
> >   }
> > 
> >   public init(_ constant: EnumConstant, environment: JNIEnvironment) { ... }
> > 
> >   public var enumConstant: EnumConstant { get }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then we could add matching operators for switch and such as needed. Perhaps even factor some of these operations out into a protocol to which Java enum classes would be made to conform. We would no longer require the virtual machine instance nor the `try!`.
> > What do you think? I'm open to other ideas, too!
> 
> I like it. I think we would need `EnumConstant` to be computed since we might get the value from Java, which would not have a way of creating the synthesized enum in Swift (I don't think). I'll prototype this setup which gives us the benefit of Swift enums without needing to use the environment and force unwrap everywhere

New update based on the proposal, my one thought would be to update the `MonthCases` to be camel case, but open to suggestions:
```
import JavaKit
import JavaRuntime

@JavaClass("java.time.Month")
public struct Month {
  public enum MonthCases: Equatable {
    case JANUARY
    case FEBRUARY
    case MARCH
    case APRIL
    case MAY
    case JUNE
    case JULY
    case AUGUST
    case SEPTEMBER
    case OCTOBER
    case NOVEMBER
    case DECEMBER
  }

  public var enumValue: MonthCases? {
    if self.equals(self.javaClass.JANUARY?.as(JavaObject.self)) {
      return MonthCases.JANUARY
    } else if self.equals(self.javaClass.FEBRUARY?.as(JavaObject.self)) {
      return MonthCases.FEBRUARY
    } else if self.equals(self.javaClass.MARCH?.as(JavaObject.self)) {
      return MonthCases.MARCH
    } else if self.equals(self.javaClass.APRIL?.as(JavaObject.self)) {
      return MonthCases.APRIL
    } else if self.equals(self.javaClass.MAY?.as(JavaObject.self)) {
      return MonthCases.MAY
    } else if self.equals(self.javaClass.JUNE?.as(JavaObject.self)) {
      return MonthCases.JUNE
    } else if self.equals(self.javaClass.JULY?.as(JavaObject.self)) {
      return MonthCases.JULY
    } else if self.equals(self.javaClass.AUGUST?.as(JavaObject.self)) {
      return MonthCases.AUGUST
    } else if self.equals(self.javaClass.SEPTEMBER?.as(JavaObject.self)) {
      return MonthCases.SEPTEMBER
    } else if self.equals(self.javaClass.OCTOBER?.as(JavaObject.self)) {
      return MonthCases.OCTOBER
    } else if self.equals(self.javaClass.NOVEMBER?.as(JavaObject.self)) {
      return MonthCases.NOVEMBER
    } else if self.equals(self.javaClass.DECEMBER?.as(JavaObject.self)) {
      return MonthCases.DECEMBER
    } else {
      return nil
    }
  }

  ... other things ...
}
```

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -359,6 +372,7 @@ extension JavaTranslator {
       contentsOf: staticFields.compactMap { field in
         // Translate each static field.
         do {
+          // Enum constants are guaranteed to not be optional
```
DougGregor: This comment is no longer relevant?

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -458,6 +472,31 @@ extension JavaTranslator {
...
+        \(raw: enumFields.map { "case \($0.getName())" }.joined(separator: "\n"))
+      }
+    """
+
+    let mappingSyntax: DeclSyntax = """
+      public var enumValue: \(raw: name)? {
+        \(raw: enumFields.map {
+          // The equals method takes a java object, so we need to cast it here
+          """
+          if self.equals(self.javaClass.\($0.getName())?.as(JavaObject.self)) {
```
DougGregor: Could you put `self.javaClass` into a local variable so we aren't accessing this computed property every time?

DougGregor:
@jrosen081 I definitely like this direction better. Should we also add an initializer that takes a `MonthCases` value and a JNI environment and gets the appropriate instance from the `self.javaClass`? That way we can go in both directions.

I'm mixed on the idea of converting `UPPER_SNAKE_CASE` to `lowerCamelCase`. It will feel a bit nicer in Swift, but might make it harder to find the corresponding Java documentation.

jrosen081:
> @jrosen081 I definitely like this direction better. Should we also add an initializer that takes a `MonthCases` value and a JNI environment and gets the appropriate instance from the `self.javaClass`? That way we can go in both directions.
> 
> I'm mixed on the idea of converting `UPPER_SNAKE_CASE` to `lowerCamelCase`. It will feel a bit nicer in Swift, but might make it harder to find the corresponding Java documentation.

New initializer:
```
  public init?(_ enumValue: MonthCases, environment: JNIEnvironment) throws {
    let classObj = try JavaClass<Self>(in: environment)
    switch enumValue {
      case .JANUARY:
        if let JANUARY = classObj.JANUARY {
          self = JANUARY
        } else {
          return nil
        }
      case .FEBRUARY:
        if let FEBRUARY = classObj.FEBRUARY {
          self = FEBRUARY
        } else {
          return nil
        }
      case .MARCH:
        if let MARCH = classObj.MARCH {
          self = MARCH
        } else {
          return nil
        }
      case .APRIL:
        if let APRIL = classObj.APRIL {
          self = APRIL
        } else {
          return nil
        }
      case .MAY:
        if let MAY = classObj.MAY {
          self = MAY
        } else {
          return nil
        }
      case .JUNE:
        if let JUNE = classObj.JUNE {
          self = JUNE
        } else {
          return nil
        }
      case .JULY:
        if let JULY = classObj.JULY {
          self = JULY
        } else {
          return nil
        }
      case .AUGUST:
        if let AUGUST = classObj.AUGUST {
          self = AUGUST
        } else {
          return nil
        }
      case .SEPTEMBER:
        if let SEPTEMBER = classObj.SEPTEMBER {
          self = SEPTEMBER
        } else {
          return nil
        }
      case .OCTOBER:
        if let OCTOBER = classObj.OCTOBER {
          self = OCTOBER
        } else {
          return nil
        }
      case .NOVEMBER:
        if let NOVEMBER = classObj.NOVEMBER {
          self = NOVEMBER
        } else {
          return nil
        }
      case .DECEMBER:
        if let DECEMBER = classObj.DECEMBER {
          self = DECEMBER
        } else {
          return nil
        }
    }
  }
  ```
It is not ideal imo to be both `throws` and failable, but it's probably best since there are really 2 types of errors here (the JNIEnvironment isn't set up right and the mapped values are optional)

**Comment by DougGregor** [comment]
> > @jrosen081 I definitely like this direction better. Should we also add an initializer that takes a `MonthCases` value and a JNI environment and gets the appropriate instance from the `self.javaClass`? That way we can go in both directions.
> > I'm mixed on the idea of converting `UPPER_SNAKE_CASE` to `lowerCamelCase`. It will feel a bit nicer in Swift, but might make it harder to find the corresponding Java documentation.
> 
> New initializer:
> 
> ```
>   public init?(_ enumValue: MonthCases, environment: JNIEnvironment) throws {
>     let classObj = try JavaClass<Self>(in: environment)
>     switch enumValue {
>       case .JANUARY:
>         if let JANUARY = classObj.JANUARY {
>           self = JANUARY
>         } else {
>           return nil
>         }
>       case .FEBRUARY:
>         if let FEBRUARY = classObj.FEBRUARY {
>           self = FEBRUARY
>         } else {
>           return nil
>         }
>       case .MARCH:
>         if let MARCH = classObj.MARCH {
>           self = MARCH
>         } else {
>           return nil
>         }
>       case .APRIL:
>         if let APRIL = classObj.APRIL {
>           self = APRIL
>         } else {
>           return nil
>         }
>       case .MAY:
>         if let MAY = classObj.MAY {
>           self = MAY
>         } else {
>           return nil
>         }
>       case .JUNE:
>         if let JUNE = classObj.JUNE {
>           self = JUNE
>         } else {
>           return nil
>         }
>       case .JULY:
>         if let JULY = classObj.JULY {
>           self = JULY
>         } else {
>           return nil
>         }
>       case .AUGUST:
>         if let AUGUST = classObj.AUGUST {
>           self = AUGUST
>         } else {
>           return nil
>         }
>       case .SEPTEMBER:
>         if let SEPTEMBER = classObj.SEPTEMBER {
>           self = SEPTEMBER
>         } else {
>           return nil
>         }
>       case .OCTOBER:
>         if let OCTOBER = classObj.OCTOBER {
>           self = OCTOBER
>         } else {
>           return nil
>         }
>       case .NOVEMBER:
>         if let NOVEMBER = classObj.NOVEMBER {
>           self = NOVEMBER
>         } else {
>           return nil
>         }
>       case .DECEMBER:
>         if let DECEMBER = classObj.DECEMBER {
>           self = DECEMBER
>         } else {
>           return nil
>         }
>     }
>   }
> ```
> 
> It is not ideal imo to be both `throws` and failable, but it's probably best since there are really 2 types of errors here (the JNIEnvironment isn't set up right and the mapped values are optional)

Perhaps it should be neither? The `try` is only there for cases where the JVM literally doesn't have the right classes loaded, which feels like it should be fatal in Swift, so I'd be okay with a `try!` there. The switch only fails if there is effectively an ABI "break" on the Java side, where a constant that existed at the time that Java2Swift runs but not when the program runs. Is that something we need to deal with?

Irrespective of our decisions on the above, I'm really liking this direction! Once you're happy with it, I recommend regenerating the java2swift sources with `make javakit-generate`.

File: Sources/Java2SwiftLib/JavaTranslator.swift [hunk 5]
```diff
@@ -458,6 +471,51 @@ extension JavaTranslator {
...
+    """
+
+    let initSyntax: DeclSyntax = """
+    public init?(_ enumValue: \(raw: name), environment: JNIEnvironment) throws {
+      let classObj = try JavaClass<Self>(in: environment)
+      switch enumValue {
+    \(raw: enumFields.map {
+      let caseName = $0.getName()
+      return """
+          case .\(caseName):
```
DougGregor: FYI this use of nested string interpolations triggered a Swift compiler crash in newly-introduced code in the compiler, which I've reduced and filed as https://github.com/swiftlang/swift/issues/77050. To avoid it (rather than waiting on a compiler fix), I think we can remove `caseName` and use `$0.getName()` everywhere.

jrosen081:
> > > @jrosen081 I definitely like this direction better. Should we also add an initializer that takes a `MonthCases` value and a JNI environment and gets the appropriate instance from the `self.javaClass`? That way we can go in both directions.
> > > I'm mixed on the idea of converting `UPPER_SNAKE_CASE` to `lowerCamelCase`. It will feel a bit nicer in Swift, but might make it harder to find the corresponding Java documentation.
> > 
> > 
> > New initializer:
> > ```
> >   public init?(_ enumValue: MonthCases, environment: JNIEnvironment) throws {
> >     let classObj = try JavaClass<Self>(in: environment)
> >     switch enumValue {
> >       case .JANUARY:
> >         if let JANUARY = classObj.JANUARY {
> >           self = JANUARY
> >         } else {
> >           return nil
> >         }
> >       case .FEBRUARY:
> >         if let FEBRUARY = classObj.FEBRUARY {
> >           self = FEBRUARY
> >         } else {
> >           return nil
> >         }
> >       case .MARCH:
> >         if let MARCH = classObj.MARCH {
> >           self = MARCH
> >         } else {
> >           return nil
> >         }
> >       case .APRIL:
> >         if let APRIL = classObj.APRIL {
> >           self = APRIL
> >         } else {
> >           return nil
> >         }
> >       case .MAY:
> >         if let MAY = classObj.MAY {
> >           self = MAY
> >         } else {
> >           return nil
> >         }
> >       case .JUNE:
> >         if let JUNE = classObj.JUNE {
> >           self = JUNE
> >         } else {
> >           return nil
> >         }
> >       case .JULY:
> >         if let JULY = classObj.JULY {
> >           self = JULY
> >         } else {
> >           return nil
> >         }
> >       case .AUGUST:
> >         if let AUGUST = classObj.AUGUST {
> >           self = AUGUST
> >         } else {
> >           return nil
> >         }
> >       case .SEPTEMBER:
> >         if let SEPTEMBER = classObj.SEPTEMBER {
> >           self = SEPTEMBER
> >         } else {
> >           return nil
> >         }
> >       case .OCTOBER:
> >         if let OCTOBER = classObj.OCTOBER {
> >           self = OCTOBER
> >         } else {
> >           return nil
> >         }
> >       case .NOVEMBER:
> >         if let NOVEMBER = classObj.NOVEMBER {
> >           self = NOVEMBER
> >         } else {
> >           return nil
> >         }
> >       case .DECEMBER:
> >         if let DECEMBER = classObj.DECEMBER {
> >           self = DECEMBER
> >         } else {
> >           return nil
> >         }
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is not ideal imo to be both `throws` and failable, but it's probably best since there are really 2 types of errors here (the JNIEnvironment isn't set up right and the mapped values are optional)
> 
> Perhaps it should be neither? The `try` is only there for cases where the JVM literally doesn't have the right classes loaded, which feels like it should be fatal in Swift, so I'd be okay with a `try!` there. The switch only fails if there is effectively an ABI "break" on the Java side, where a constant that existed at the time that Java2Swift runs but not when the program runs. Is that something we need to deal with?

Didn't think of it like that. Went with a `try!` on the environment (if it fails there are bigger issues). For the switch failing, I added a fatalError to explain what is going wrong and how to likely fix it.

> 
> Irrespective of our decisions on the above, I'm really liking this direction! Once you're happy with it, I recommend regenerating the java2swift sources with `make javakit-generate`.

I ran it, but nothing ended up changing (I don't think we are converting any enums right now)

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -77,6 +77,7 @@ extension JavaTranslator {
   package static let defaultTranslatedClasses: [String: (swiftType: String, swiftModule: String?, isOptional: Bool)] = [
     "java.lang.Class": ("JavaClass", "JavaKit", true),
     "java.lang.String": ("String", "JavaKit", false),
+    "java.lang.Object": ("JavaObject", "JavaKit", true)
```
DougGregor: I believe this change shouldn't be necessary. `java.lang.Object` is already listed in `Java2Swift.config` for the `JavaKit` target, so the only way this could be needed is if something is missing a JavaKit dependency.

DougGregor (commented):
I left one small comment. This looks great!

DougGregor:
> > Irrespective of our decisions on the above, I'm really liking this direction! Once you're happy with it, I recommend regenerating the java2swift sources with `make javakit-generate`.
> 
> I ran it, but nothing ended up changing (I don't think we are converting any enums right now)

Okay! I'm sure we'll find some more enums as we bring more of the Java base types into the JavaKit modules.

Sources/Java2SwiftLib/JavaTranslator.swift
DougGregor:
```suggestion
```

ktoso:
LGTM! Looking good, thanks for the contribution! :-)

ktoso (approved):

DougGregor:
@jrosen081 merged, thank you!

Issue #16, opened 2024-09-30
SwiftPM build plugin for producing Swift wrapper types for Java classes (Java2Swift direction)


DougGregor:
The "Java2Swift" tool produces Swift wrapper types (using JavaKit) for Java classes. It's provided as a standalone executable that takes in an input module, some translation manifests, and either a Jar file or a list of class names, and produces Swift source files.

We should create a [SwiftPM build plugin](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/Plugins.md) that runs the tool for us. For example, it should make it easy to declare a SwiftPM target and fill in some of the source files via the generator. For example, the `JavaKitJar` module is [fully generated](https://github.com/swiftlang/swift-java/tree/main/Sources/JavaKitJar/generated), but with a Makefile.

DougGregor:
This was mostly implemented in PR #77 , but we need to make sure that we're picking up the manifest files generated from other targets we depend on. Right now, we only get the manifest files if they are in the "pre-baked" location `generated/<modulename>.swift2java`.

DougGregor:
Done and documented

PR #77, opened 2024-10-15
Add a SwiftPM build plugin to generate Swift wrappers for Java classes

DougGregor:
Introduce the plugin `Java2SwiftPlugin`, which will invoke Java2Swift to generate Swift wrappers for the specified Java classes. Each SwiftPM target that uses this plugin must contain a file named `Java2Swift.config` that specifies options for the Java2Swift tool. This is a JSON file with a top-level dictionary that currently can contain keys for the Java class path (`classPath`) and a dictionary called `classes` whose keys are Java canonical class names (such as `java.util.Vector`) and whose values are the corresponding Swift names. For example:

```swift
{
  "classes" : {
    "java.util.Vector" : "JavaVector"
  }
}

```

will translate the class `java.util.Vector` into the Swift type `JavaVector`. The resulting Swift files will become part of the Swift module that uses this plugin.

I've updated the JavaKitSampleApp to make use of this, albeit in a very narrow way that is not at all compelling. The user guide also needs updating.

Fixes issue #16


Samples/JavaKitSampleApp/Sources/JavaKitExample/Java2Swift.config
```diff
@@ -0,0 +1,5 @@
+{
+  "classes" : {
+    "java.util.Vector" : null
```
ktoso: Vector is a rarely used type that folks generally avoid as it predates the "proper collections framework", I'd recommend avoiding it in examples (it'll look weird to Java developers that we're using that anywhere) and using `java.util.ArrayList` in examples.

--- 

double checked in docs:

> As of the Java 2 platform v1.2, this class was retrofitted to implement the [List](https://docs.oracle.com/javase/8/docs/api/java/util/List.html) interface, making it a member of the [Java Collections Framework](https://docs.oracle.com/javase/8/docs/technotes/guides/collections/index.html). Unlike the new collection implementations, Vector is synchronized. If a thread-safe implementation is not needed, it is recommended to use [ArrayList](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html) in place of Vector.

Package.swift
```diff
@@ -56,6 +56,11 @@ let package = Package(
       targets: ["JavaKit"]
     ),
 
+    .library(
+      name: "JavaRuntime",
+      targets: [ "JavaRuntime"]
```
ktoso:
```suggestion
      targets: ["JavaRuntime"]
```

Need to enable the Swift formatting enforcement soon 🤔

ktoso (approved):
Good start, LGTM, minor nitpick inline

.licenseignore
```diff
@@ -5,6 +5,7 @@
 .swift-format
 .github/*
 *.md
+**/*.config
```
ktoso, line 8: This is to fix the license checking CI job, since it checked and failed on the new file not having a license header

ktoso (approved):

Package.swift
DougGregor: Eh, sure

.licenseignore
DougGregor: Ah, thank you!

Samples/JavaKitSampleApp/Sources/JavaKitExample/Java2Swift.config
DougGregor: I was picking something silly that we wouldn't actually end up wrapping in any of the JavaKit* modules, but yeah... Vector is extra silly. Thanks for the ArrayList suggestion.

ktoso: Yeah, no harm but a bit too silly maybe :) No prob

DougGregor:
I have improvements I'd still like to make (e.g., handling Jar files, documenting this new path), but this is a good start

PR #83, opened 2024-10-15
Rework Java2Swift to operate entirely on Java2Swift.config files

DougGregor:
Instead of having Java2Swift take in a list of classes on the command line, which the plugin synthesizes by reading a configuration file, and then emitting another almost-identical manifest file... rework the command-line interface to Java2Swift to operate entirely on the per-target Java2Swift.config files. This makes the plugin simpler, because it never has to look at build products for its dependencies.

Change the `--jar-file` mode of Java2Swift to emit the configuration file but not translate any classes. This should be used as a tool, not a build step.

Update the user guide to introduce JavaKit through the build tool.


DougGregor:
This completes issue #16

PR #87, opened 2024-10-17
Java2Swift improvements for producing configuration files from Jar files

DougGregor:
Improve the experience of producing `Java2Swift.config` files from Jar files and better cope with layers of Swift-wrapped Java libraries. This includes:

* Filtering out nested classes from Jar files, since we cannot handle them yet
* Combining the class paths from all of the dependent `Java2Swift.config` files with the one being used and providing those to the JVM instance
* Fixing an issue where Java2Swift would generate `func init()` for a Java method named `init`

### DIFF
--- a/Sources/Java2Swift/JavaToSwift.swift
+++ b/Sources/Java2Swift/JavaToSwift.swift
@@ -58,55 +58,118 @@ struct JavaToSwift: ParsableCommand {
   )
   var input: String
 
-  mutating func run() throws {
-    var vmOptions: [String] = []
-    let classpath = classPathWithJarFile
-    if !classpath.isEmpty {
-      vmOptions.append("-cp")
-      vmOptions.append(contentsOf: classpath)
-    }
+  /// Describes what kind of generation action is being performed by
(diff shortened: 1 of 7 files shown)


Sources/Java2Swift/JavaToSwift.swift
```diff
@@ -58,55 +58,118 @@ struct JavaToSwift: ParsableCommand {
...
     }
 
-    let jvm = try JavaVirtualMachine.shared(vmOptions: vmOptions)
-    try run(environment: jvm.environment())
+    // Form a class path from all of our input sources:
+    //   * Command-line option --classpath
+    var classPathPieces: [String] = classpath
+    switch generationMode {
+    case .configuration(jarFile: let jarFile):
+      //   * Jar file (in `-jar-file` mode)
```
ktoso, line 103: I didn't notice this before, `-jar` probably would spell nicer for the option btw.

DougGregor: Sure! PR incoming

PR #90, opened 2024-10-18
JavaKitCollection module

DougGregor:
Introduce a new JavaKitCollection module that wraps up a bunch of the Java collection-related classes for use in Swift


ktoso (approved):
Hmm, yeah a module for collections per se probably makes sense. Even if it doesn't follow the "module = package" separation (that'd be Util then hm).

ktoso:
Gotta add all the generated sources to .licenseignore, like we do:

```
Sources/JavaKit/generated/*
Sources/JavaKitJar/generated/*
Sources/JavaKitNetwork/generated/*
Sources/JavaKitReflection/generated/*
...
```

maybe it works with ** actually?

DougGregor:
> maybe it works with ** actually?

It works, thanks!

DougGregor:
> Hmm, yeah a module for collections per se probably makes sense. Even if it doesn't follow the "module = package" separation (that'd be Util then hm).

Yeah, I'm not sure this kind of splitting actually makes sense vs. just throwing everything into `JavaKit` and being done with it. There are lots more common classes that I keep encountering.

ktoso:
Yeah, we can clean this up later. There’s a bunch of “common types” that one is bound to encounter that could be in the core maybe

PR #98, opened 2024-10-22
cache JDK, speedup downloads and CI time

ktoso:


ktoso:
// We're not using install-java because of base image incompatibilities

Issue #70, opened 2024-10-11
Make the "environment" parameter to `@JavaMethod` initializers optionally-typed and default to `nil`


DougGregor:
Right now, an initializer of an `@JavaClass` type requires an environment parameter, e.g.,

```swift
@JavaClass("org.swift.javakit.examples.MyClass")
struct MyClass {
  @JavaMethod
  init(environment: JNIEnvironment)
}
```

We could allow this to be:

```swift
@JavaClass("org.swift.javakit.examples.MyClass")
struct MyClass {
  @JavaMethod
  init(environment: JNIEnvironment? = nil)
}
```

And, if `nil` is passed, use `JavaVirtualMachine.shared().environment()` (introduced in https://github.com/swiftlang/swift-java/pull/69) as the default value. This requires some work on the `@JavaMethod` macro implementation, and collapsing `JavaKitVM` down into `JavaKit`, but should otherwise be straightforward.

PR #102, opened 2024-10-22
Make the "environment" parameter to @JavaMethod initializers optional

ShenghaiWang:
Closes https://github.com/swiftlang/swift-java/issues/70

Make the "environment" parameter to @JavaMethod initializers optionally-typed and default to nil


Sources/JavaKitMacros/JavaMethodMacro.swift
```diff
@@ -94,7 +94,8 @@ extension JavaMethodMacro: BodyMacro {
 
     return [
       """
-      self = \(raw: tryKeyword) Self.dynamicJavaNewObject(in: environment\(raw: arguments))
+      let _environment = environment == nil ? \(raw: tryKeyword) JavaVirtualMachine.shared().environment() : environment!
```
ktoso: maybe this would be nicer, without resorting to the ! 

```suggestion
      let environment = 
        if let environment { 
          environment
        } else {
          \(raw: tryKeyword) JavaVirtualMachine.shared().environment()
        }
```

ktoso (approved):
Looks ok to me, seems that's what @DougGregor had in mind.

Module wise I'm not sure if the separate VM module gained us anything, either seems fine.

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -483,8 +483,9 @@ extension JavaTranslator {
     """
 
     let initSyntax: DeclSyntax = """
-    public init(_ enumValue: \(raw: name), environment: JNIEnvironment) {
-      let classObj = try! JavaClass<Self>(in: environment)
+    public init(_ enumValue: \(raw: name), environment: JNIEnvironment? = nil) {
+      let _environment = environment == nil ? try! JavaVirtualMachine.shared().environment() : environment!
```
DougGregor: I'd rather do this without the force-unwraps, even if it's more verbose:

```swift
let _environment = if let environment {
  environment
} else {
  try! JavaVirtualMachine.shared().environment()
}
```

Sources/JavaKitMacros/JavaMethodMacro.swift
DougGregor: Ah, I had the same comment earlier ;)

DougGregor (commented):
I would like to remove the unnecessary force-unwraps, but otherwise this looks good to me and thank you!

DougGregor:
> Module wise I'm not sure if the separate VM module gained us anything, either seems fine.

The VM module really didn't seem to be buying us anything, so I'm fine with collapsing it into JavaKit.

ktoso:
LGTM, thanks a lot 👍

PR #103, opened 2024-10-22
workflow experiment with building just once

ktoso:


ktoso:
Ok yeah, build results don't carry over to the other jobs unless part of the cache. We can play around with that but that's not quite like it.

PR #105, opened 2024-10-23
JavaKit: Rework `@ImplementsJava` to be more like `@implements` language feature

DougGregor:
The `@ImplementsJava` macro had an odd formulation where it was used similarly to `@JavaMethod`, but was actually interpreted by the outer `@JavaClass` macro. Completely rework the way `@ImplementsJava` is used to make it more like the `@implements` feature introduced in [SE-0436](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0436-objc-implementation.md), and rename it to `@JavaImplementation` to more closely match the canonical spelling for `@implementation` with a language and make use of the `@Java...` as a common prefix.

In this new design, one places `@JavaImplementation` on an extension of the Swift type, conforms to the corresponding protocol, and marks the implementing methods with `@JavaMethod`. For example:

    @JavaImplementation("org.swift.javakit.HelloSwift")
    extension Hello: HelloNativeMethods {
      @JavaMethod
      func reportStatistics(_ meaning: String, _ numbers: [Double]) -> String {
        let average = numbers.isEmpty ? 0.0 : numbers.reduce(0.0) { $0 + $1 } / Double(numbers.count)
        return "Average of \(meaning) is \(average)"
      }
    }

This also means that we can generate most of the `@JavaClass` definition for a given Java class, and leave the native methods to be implemented by an `@JavaImplementation` extension. The <SwiftType>`NativeMethods` protocol provides declarations for all of the `native` methods that are expected to be implemented in Swift, making it easier to get all of
the signatures right.

Update the JavaKitSampleApp to use the Java compilation plugin to compile the Java files in the target. Automatically feed those into the Java2Swift plugin to generate the Swift representations of the Java classes, skipping the native methods (since they'll be implemented here).

The user guide / tutorial is a bit behind the implementation here as I smooth out more corners.


Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -383,46 +420,74 @@ extension JavaTranslator {
...
     }
 
-    let extDecl =
-      ("""
-    extension JavaClass\(raw: extSpecialization) {
-    \(raw: staticMembers.map { $0.description }.joined(separator: "\n\n"))
+    if !nativeMembers.isEmpty {
+      let protocolDecl: DeclSyntax =
+        """
+        /// Describes the Java `native` methods for \(raw: swiftTypeName).
```
ktoso: worth a?

```suggestion
        /// Describes the Java `native` methods for ``\(raw: swiftTypeName)``.
```

USER_GUIDE.md
```diff
@@ -213,6 +213,69 @@ do {
...
+  "classes" : {
+    "org.swift.javakit.example.HelloSwift" : "Hello",
+  }
+}
+```
+
+Implementations of `native` methods are written in an extension of the Swift type that has been marked with `@JavaImplementation`. The methods themselves must be marked with `@JavaMethod`, indicating that they are available to Java as well. To help ensure that the Swift code implements all of the `native` methods with the right signatures, JavaKit produces a protocol with the Swift type name suffixed by `NativeMethods`. Declare conformance to that protocol and implement its requirements, for example:
+
+```swift
+@JavaImplementation("org.swift.javakit.HelloSwift")
```
ktoso, line 265: Just a thought on the names matching dance here: can we do better than that? This currently needs to match up the java file, the config file, and the swift file. Can the config file eventually become an implementation detail? It seems like a moving piece that folks are bound to forget about when e.g. moving the java type to some other package 🤔 

Perhaps just more good diagnostics would be a way to address this as well.

Occurred to me now while reviewing and comparing with how much of a hassle this is vs integrating Java/JNI/C 🤔

ktoso (approved):
Overall looking good, let's keep polishing this up :) 

We can figure out what the problem with necessitating public was here, it should not be a requirement for `native` AFAIR

USER_GUIDE.md
DougGregor: I think it depends on how much work we're willing to do in the plugin itself. The config file lets us predetermine all of the `.swift` files that will be generated by reading just the one file. If we want to extract that information from the Swift or Java sources, we have to do a lot more digging just to set up the build graph.

The *NativeMethods protocol should help a bit with the names-matching, though. You write the extension to conform to protocol, and the compiler makes sure that all of the names and types match the `native` function. 

Good diagnostics would definitely help here, and we should think more about whether it's okay to generate more of the config file to save steps.

DougGregor:
> We can figure out what the problem with necessitating public was here, it should not be a requirement for `native` AFAIR

Yeah, I think I know the problem here. See https://github.com/swiftlang/swift-java/issues/106

DougGregor:
https://github.com/swiftlang/swift-java/pull/105/commits/e87c3bf56e81850cdb6b312812924fd84deac7e3 addressed the crash in the tests that were causing me trouble on Linux. I suspect the development toolchain might be miscompiling closure captures of parameter packs in some cases.

Issue #82, opened 2024-10-15
SwiftPM command plugin to ingest a Jar file into a SwiftPM target to wrap those Java classes


DougGregor:
The SwiftPM build tool plugin `Java2SwiftPlugin` wraps up the specified set of Java classes so they can be used from Swift. It requires a configuration file containing the mapping from specific Java canonical class names to their Swift type names. However, we'd like to make it easier to get started by importing a Jar file into a Swift target, scanning the Jar file for any public classes and generating the `Java2SwiftPlugin` configuration file.

A SwiftPM [command plugin](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/Plugins.md#command-plugins) would be a reasonable approach here. Essentially, it would take a SwiftPM target name and Jar file and update the package to wrap the public classes from that Jar file. It could be written as:

```
swift package import-java-jar MySwiftTarget MyJavaLib.jar
```

This command would produce two outputs:
1. A file `Sources/MySwiftTarget/Java2Swift.config` that includes the Jar file in the class path and lists the mappings for all of the public classes within that Jar file.
2. An updated `Package.swift` that contains a package dependency on swift-java and applies the Java2SwiftPlugin to the MySwiftTarget.

The goal should be that one can "ingest" a Java library with one command, then `swift build` to get access to that library from Swift.

ktoso:
Note: we likely want to converge all those tasks into the same plugin that has a bunch of sub-commands.

This plugin would likely be a command plugin named `java` or `swift-java`

PR #107, opened 2024-10-23
Generate a Swift Package from a Jar file

ShenghaiWang:
Closes https://github.com/swiftlang/swift-java/issues/82

Please note: Moved some code from `JavaToSwift` to `Java2SwiftLib` in order to reuse them in `JarImporter`

### DIFF
--- a/Plugins/JavaCompilerPlugin/JavaCompilerPlugin.swift
+++ b/Plugins/JavaCompilerPlugin/JavaCompilerPlugin.swift
@@ -58,7 +58,8 @@ struct JavaCompilerBuildToolPlugin: BuildToolPlugin {
           .appending(path: "bin")
           .appending(path: "javac"),
         arguments: javaFiles.map { $0.path(percentEncoded: false) } + [
-          "-d", javaClassFileURL.path()
+          "-d", javaClassFileURL.path(),
+          "-parameters", // keep parameter names, which allows us to emit them in generated Swift decls
         ],
         inputFiles: javaFiles,
         outputFiles: classFiles
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -389,6 +389,19 @@ A number of JavaKit modules provide Swift projections of Java classes and interf
 
 The `Java2Swift` tool can translate any other Java classes into Swift projections. The easiest way to use `Java2Swift` is with the SwiftPM plugin described above. More information about using this tool directly are provided later in this document
 
+#### Improve parameter names of imported Java methods
+When building Java libraries you can pass the `-parameters` option to javac
+in your build system (Gradle, Maven, sbt, etc) in order to retain the parameter names in the resulting byte code.
+
+This way the imported methods will keep their original parameter names, and you'd get e.g.:
+```swift
+// public func hello(String name)
+func hello(_ name: String)
+```
+rather than just `arg0` parameters.
+
+When building Java sources using the JavaCompilerPlugin this option is passed by default.
+
 ### Class objects and static methods
 
 Every `AnyJavaObject` has a property `javaClass` that provides an instance of `JavaClass` specialized to the type. For example, `url.javaClass` will produce an instance of `JavaClass<URL>`. The `JavaClass` instance is a wrapper around a Java class object (`java.lang.Class`) that has two roles in Swift. First, it provides access to all of the APIs on the Java class object. The `JavaKitReflection` library, for example, exposes these APIs and the types they depend on (`Method`,


chengenzhao:
Is it possible to generate Swift Module from a Jar?

Plugins/JarImporterPlugin/JarImporterPlugin.swift
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Please specify the module name and the Java jar file.
+        Usage: swift package import-java-jar MySwiftTarget MyJavaLib.jar
+        """)
+      return
+    }
+    let jarImporter = try context.tool(named: "JarImporter")
+    let sometoolArgs = [
+      "--module-name", mouduleName,
+      "--jar-file", jarFile,
```
ktoso:
```suggestion
      "--jar", jarFile,
```

please :)

Sources/Java2SwiftLib/ConfigFileGenerator.swift
```diff
@@ -0,0 +1,257 @@
...
+  let outputDirectory: String?
+  /// if it's handling jar file
+  let jar: Bool
+  /// Whether we have ensured that the output directory exists.
+  private var createdOutputDirectory: Bool = false
+  private var inNewSwiftPackage: Bool
+
+  package init(moduleName: String,
+              outputDirectory: String? = nil,
+              jar: Bool,
```
ktoso: I'd prefer we avoid just bools like this that are unclear what behavior changes they cause like this. Instead the caller should construct a correct output directory, which is the only thing this was used for.

Sources/Java2SwiftLib/ConfigFileGenerator.swift
```diff
@@ -0,0 +1,257 @@
...
+    translator.addConfiguration(config, forSwiftModule: moduleName)
+
+    // Load all of the requested classes.
+#if false
+    let classLoader = URLClassLoader(
+      [
+        try URL("file://\(classPath)", environment: environment)
+      ],
+      environment: environment
+    )
```
ktoso: Do we need this dead code @DougGregor ? (It's carried over existing code)

Sources/JarImporter/JarImporter.swift
```diff
@@ -0,0 +1,104 @@
...
+      forJarFile: jarFile,
+      classPath: jarFile,
+      environment: jvm.environment())
+
+    // Write out Package.swift and a dummy swift file
+    try generatePackage(moudleName: moduleName)
+  }
+
+  /// Write out the Package.swift file and a dummy swift file
+  func generatePackage(moudleName: String) throws {
```
ktoso: Instead of creating a new package we should add a target in the _current package_.
You can use the package editing commands to do so: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0301-package-editing-commands.md

So we'll do the extraction, emit the config and files, and then add a target to the current project using `swift package add-target <ThatNewJavaTarget>` I think.

Package.swift
```diff
@@ -103,6 +103,14 @@ let package = Package(
         ]
     ),
 
+    // ==== Plugin for generating Swift Package from a Jar file
+    .plugin(
+        name: "JarImporterPlugin",
```
ktoso: Some food for thought -- doesn't hold up this current PR.

I wonder if we should align the words we're using. The two workflows use separate terminology. It's getting confusing remembering all the different names of tools.

In a way this is just like `jextract` but the other way (prefixing with `s` sadly gives us a bad name, so maybe we would` `swift-extract --jar` instead or something?

Alternativly dropping the jextract name would be an option, and just call it all `java2swift` and `swift2java` for the command line tools, then this would be `jar2swift --target Target <jar>` etc maybe?

Sources/JarImporter/JarImporter.swift
```diff
@@ -0,0 +1,104 @@
...
+/// Command-line utility to import jar into Swift.
+@main
+struct JarImporter: ParsableCommand {
+  static var _commandName: String { "JarImporter" }
+
+  @Option(help:"The name of the Swift Package that will be generated.")
+  var moduleName: String
+
+  @Option(help: "The Jar file that needs to be imported. If using a relative path, please make sure there is a same jar file that has the same relationship with the generated package before compiling the package.")
+  var jarFile: String
```
ktoso: Actually, since this tool is only going to work on jars, this would be better expressed as an @Argument `jar2swift --target Name <jar>` makes some more sense than repeating the `--jar` option

Sources/JarImporter/JarImporter.swift
```diff
@@ -0,0 +1,104 @@
+
```
ktoso: stray whitespace

ktoso:
Shaping up well, but needs some more work -- we need to generate a target when importing a jar, not a whole package :) 

Thanks for picking up this task!

Sources/Java2SwiftLib/ConfigFileGenerator.swift
ShenghaiWang: Yeah I agree. I followed the usage of `jar` from `JavaToSwift` and wanted to have minimum change for that part. I will refine both.

Plugins/JarImporterPlugin/JarImporterPlugin.swift
ShenghaiWang: I did this on purpose in order to differentiate from the `jar` usage in the `JavaToSwift`. I will think about it and update :)

Sources/JarImporter/JarImporter.swift
ShenghaiWang: I think a new package would be better for real use cases. How do you think? I could be wrong, let me know.

Sources/JarImporter/JarImporter.swift
ShenghaiWang: Yep. We can update it. It doesn't really matter though as the user will not use this directly.

Sources/JarImporter/JarImporter.swift
DougGregor: I agree with @ktoso. SwiftPM already has package-creation commands (`swift package init`). What I actually expect folks to want is to be able to augment an existing package by adding/updating a target with the appropriate plugins, dependencies, and so on.

ShenghaiWang:
> Is it possible to generate Swift Module from a Jar?

Yes. Will add this.

ShenghaiWang:
> Shaping up well, but needs some more work -- we need to generate a target when importing a jar, not a whole package :)

Yep. I will add this in next push.

Sources/JarImporter/JarImporter.swift
ShenghaiWang: @DougGregor @ktoso 

1. It seems there is no way to add `swiftSettings` using this command. Is it ok we use swift syntax library to add the `swiftSettings`? `swift package add-target <name> [--type <type>] [--dependencies <dependencies> ...] [--url <url>] [--path <path>] [--checksum <checksum>] [--testing-library <testing-library>]` 
2. Do you know the best way to get the path of `swift` that works for different platform? I have one but it seems very lengthy :(
Thanks.

DougGregor: Yes, we can use the swift-syntax library to create edit packages. The Swift Package Manager codebase has some of these "refactoring" operations that use swift-syntax to add targets/dependencies/products/etc. I think what we probably should do is lift those into swift-syntax itself, then add new ones for adding plugins, updating swift settings, and so on. Finally, we can use those from this plugin to perform all of the edits.

ShenghaiWang: Good point. Let me take a look at these repos.

PR #109, opened 2024-10-25
format all swift sources, enable format checks

ktoso:
One of those "it has to be done someday" tasks... enabling formatting checking and formatting all Swift sources.

This is just the `swift format Sources --recursive --in-place` invocation. 

We can tweak the config if we don't like the output but it'll be nice to have consistent formatting.

WDYT?


Sources/ExampleSwiftLibrary/MySwiftLibrary.swift
```diff
@@ -35,7 +35,7 @@ public func globalTakeIntInt(i: Int, j: Int) {
   p("i:\(i), j:\(j)")
 }
 
-public func globalCallMeRunnable(run: () -> ()) {
+public func globalCallMeRunnable(run: () -> Void) {
```
ktoso, line 38: c'mon, seriously swift-format? 🤪

(pretty sure that'll break some things since we checked the () -> () string 😅 )

Sources/JExtractSwift/ImportedDecls.swift
```diff
@@ -285,13 +288,14 @@ public struct ImportedVariable: ImportedDecl, CustomStringConvertible {
...
-      let newValueParam: FunctionParameterSyntax = "_ newValue: \(self.returnType.cCompatibleSwiftType)"
+      let newValueParam: FunctionParameterSyntax =
+        "_ newValue: \(self.returnType.cCompatibleSwiftType)"
       var funcDecl = ImportedFunc(
         parentName: self.parentName,
         identifier: self.identifier,
         returnType: TranslatedType.void,
         parameters: [.init(param: newValueParam, type: self.returnType)])
-      funcDecl.swiftMangledName = self.swiftMangledName + "s" // form mangled name of the getter by adding the suffix
+      funcDecl.swiftMangledName = self.swiftMangledName + "s"  // form mangled name of the getter by adding the suffix
```
ktoso, line 298: weird decision by swift-format here...

ktoso:
Uncovered fun bug in swift-format: swift-format breaks compilation by wrongly reformatting nested multi-line string blocks https://github.com/swiftlang/swift-format/issues/864

Issue #106, opened 2024-10-23
Java2Swift: Investigate using getDeclaredMethods()/getDeclaredConstructors()/etc. to find non-public members


DougGregor:
Right now, the Java2Swift tool uses `java.lang.Class`'s `getMethods()`, `getConstructors()`, etc. that provides only public members. We should investigate whether using the `getDeclaredMethods()` operation would allow us to expose non-public members as well, which could be particularly useful for implementing `native` methods.

PR #110, opened 2024-10-25
Use getDeclaredMethods() so we can see non-public methods of classes

DougGregor:
When we are translating a class that was built as part of the Swift module, use getDeclaredMethods() so we also see non-public methods.

Part of issue #106.


ktoso (approved):
:+1:

DougGregor:
I want to rework the class translator to get more information out of the `getDeclaredXYZ` methods, but I'll merge this now to nudge the needle forward a little bit first.

PR #111, opened 2024-10-25
Minor cleanups to the JavaKit module and samples

DougGregor:
Perform a few minor cleanups and expansions to the JavaKit module:
* Rename `JavaClass.init(in:)` to `JavaClass.init(environment:)` and make the parameter optional, to match other similar APIs
* Autogenerate the `java.lang.Class` binding rather than hand-rolling it
* Import the various `java.lang.*` class wrappers for primitive types (like `Integer` and `Double`)
* Clean up / fix up the JavaKit samples


Samples/JavaSieve/Sources/JavaSieve/main.swift
```diff
@@ -13,11 +13,10 @@
 //===----------------------------------------------------------------------===//
 
 import JavaKit
-import JavaKitVM
 
 let jvm = try JavaVirtualMachine.shared(classPath: ["QuadraticSieve-1.0.jar"])
 do {
-  let sieveClass = try JavaClass<SieveOfEratosthenes>(in: jvm.environment())
+  let sieveClass = try JavaClass<SieveOfEratosthenes>(environment: jvm.environment())
```
ktoso, line 19: +1, abusing the `in:` as label is always a tad too much in Swift libs 😆

PR #115, opened 2024-10-26
Add Support for Nested Classes

### LINKED ISSUE #19: Swift wrappers for nested Java classes
Swift wrapper generation from Java classes (`Java2Swift`) doesn't currently handle nested classes well. It should

jrosen081:
Closes #19 

This adds support for nested classes. There are 2 parts here:
1. Recursively going through when translating classes
2. Using a stack to go through all classes to add to the classes to translate.

This also migrates some places where we were using `getCanonicalName` when we should be using `getName` (to make sure that we use the $)

### DIFF
--- a/Sources/Java2Swift/JavaToSwift.swift
+++ b/Sources/Java2Swift/JavaToSwift.swift
@@ -342,11 +342,6 @@ struct JavaToSwift: ParsableCommand {
         }
       }
 
-      // TODO: For now, skip all nested classes.
-      if entry.getName().contains("$") {
-        continue
-      }
-
       let javaCanonicalName = String(entry.getName().replacing("/", with: ".")
         .dropLast(".class".count))
       configuration.classes[javaCanonicalName] =
--- a/Sources/Java2Swift/JavaToSwift.swift
+++ b/Sources/Java2Swift/JavaToSwift.swift
@@ -229,10 +229,26 @@ struct JavaToSwift: ParsableCommand {
       // Note that we will be translating this Java class, so it is a known class.
       translator.translatedClasses[javaClassName] = (translatedSwiftName, nil, true)
 
-      for internalClass in javaClass.getClasses() {
+      var classes: [(classToImport: JavaClass<JavaObject>?, prefix: String)] = javaClass.getClasses().map { ($0, translatedSwiftName) }
+
+      // Go through all subclasses to find all of the classes to translate
+      while let (internalClass, prefix) = classes.popLast() {
         if let internalClass {
           let (javaName, swiftName) = names(from: internalClass.getCanonicalName())
-          translator.translatedClasses[internalClass.getCanonicalName()] = (swiftName.replacing("$", with: "."), nil, true)
+          // If we have already been through this class, don't go through it again
+          guard translator.translatedClasses[javaName] == nil else { continue }
+          let currentClassName: String
+          // If the prefix is the part of the canonical name, continue (else use the same)
+          // This is needed since `java.lang.Enum.EnumDesc` is one of the declared classes of all of the enums
+          if (internalClass.getCanonicalName().contains(prefix)) {
+            currentClassName = "\(prefix).\(swiftName)"
+          } else {
+            currentClassName = swiftName
+          }
+
+          let currentSanitizedClassName = currentClassName.replacing("$", with: ".")
+          classes.append(contentsOf: internalClass.getClasses().map { ($0, currentSanitizedClassName) })
+          translator.translatedClasses[javaName] = (currentSanitizedClassName, nil, true)
         }
       }
     }
--- a/Sources/Java2Swift/JavaToSwift.swift
+++ b/Sources/Java2Swift/JavaToSwift.swift
@@ -243,7 +243,7 @@ struct JavaToSwift: ParsableCommand {
           if (internalClass.getCanonicalName().contains(prefix)) {
             currentClassName = "\(prefix).\(swiftName)"
           } else {
-            currentClassName = swiftName
+            continue // If we have a class that is not a nested one, ignore it
           }
 
           let currentSanitizedClassName = currentClassName.replacing("$", with: ".")
@@ -256,7 +256,7 @@ struct JavaToSwift: ParsableCommand {
     // Translate all of the Java classes into Swift classes.
     for javaClass in javaClasses {
       translator.startNewFile()
-      let swiftClassDecls = translator.translateClass(javaClass)
+      let swiftClassDecls = try! translator.translateClass(javaClass)
       let importDecls = translator.getImportDecls()
 
       let swiftFileText = """
@@ -294,15 +294,6 @@ struct JavaToSwift: ParsableCommand {
     }
 
     return (javaClassName, swiftName)
-  }
-
-  /// Return the class path augmented with the Jar file, if there is one.
-  var classPathWithJarFile: [String] {
-    if jarFile {
-      return [input] + classpath
-    }
-
-    return classpath
   }
 
   mutating func writeContents(_ contents: String, to filename: String, description: String) throws {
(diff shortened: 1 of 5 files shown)


DougGregor:
Hah! I didn't see your PR and went off and took at stab at this myself (https://github.com/swiftlang/swift-java/pull/118). It looks like we found some of the same things (e.g., the getCanonicalName() -> getName() bit), you handled the Jar reading better, I fixed up the plugin. I'll review yours and rebase mine on top

Sources/Java2Swift/JavaToSwift.swift
```diff
@@ -228,12 +228,27 @@ struct JavaToSwift: ParsableCommand {
...
+          translator.translatedClasses[javaName] = (currentSanitizedClassName, nil, true)
+        }
+      }
     }
 
     // Translate all of the Java classes into Swift classes.
     for javaClass in javaClasses {
       translator.startNewFile()
-      let swiftClassDecls = translator.translateClass(javaClass)
+      let swiftClassDecls = try! translator.translateClass(javaClass)
```
DougGregor: Shouldn't the be a `try` rather than a `try!`?

```suggestion
      let swiftClassDecls = try translator.translateClass(javaClass)
```

Sources/Java2Swift/JavaToSwift.swift
```diff
@@ -247,11 +262,32 @@ struct JavaToSwift: ParsableCommand {
       try writeContents(
         swiftFileText,
         to: swiftFileName,
-        description: "Java class '\(javaClass.getCanonicalName())' translation"
+        description: "Java class '\(javaClass.getName())' translation"
       )
     }
   }
 
+  private func names(from javaClassNameOpt: String) -> (javaClassName: String, swiftName: String) {
```
DougGregor, line 270: Thanks for pulling this out

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -265,7 +274,7 @@ extension JavaTranslator {
     )
 
     if !enumConstants.isEmpty {
-      let enumName = "\(swiftTypeName)Cases"
+      let enumName = "\(swiftTypeName.splitSwiftTypeName().name)Cases"
```
DougGregor: This can be `swiftInnermostTypeName`, right?
```suggestion
      let enumName = "\(swiftInnermostTypeName)Cases"
```

Sources/Java2Swift/JavaToSwift.swift
```diff
@@ -228,12 +228,27 @@ struct JavaToSwift: ParsableCommand {
...
+      while let internalClass = classes.popLast() {
+        if let internalClass {
+          let (javaName, swiftName) = names(from: internalClass.getName())
+          // If we have already been through this class, don't go through it again
+          guard translator.translatedClasses[javaName] == nil else { continue }
+          let currentClassName = swiftName
+          let currentSanitizedClassName = currentClassName.replacing("$", with: ".")
+          classes.append(contentsOf: internalClass.getClasses())
+          translator.translatedClasses[javaName] = (currentSanitizedClassName, nil, true)
+        }
```
DougGregor, line 244: I don't think we should always recurse into nested classes. Rather, I think we should allow specification of nested classes in `Java2Swift.config`, and teach the Jar walker that produces `Java2Swift.config` to walk nested classes.

Sources/Java2SwiftLib/JavaTranslator.swift
```diff
@@ -383,6 +391,20 @@ extension JavaTranslator {
...
+        } catch {
+          logUntranslated("Unable to translate '\(fullName)' subclass '\(clazz.getName())': \(error)")
+          return nil
+        }
+      }
+    }.flatMap(\.self)
+
+    topLevelDecls.append(
+      contentsOf: subClassDecls
+    )
```
DougGregor, line 407: Hmm. I don't think we actually want the recursion here? As noted above, I think nested classes that are translated should be listed in the top-level Java2Swift.config. Then, the top-level loop that calls `translateClass` will also handle nested classes.

Sources/JavaKitJar/generated/Attributes.swift
```diff
@@ -95,3 +98,92 @@ public struct Attributes {
   @JavaMethod
   public func getOrDefault(_ arg0: JavaObject?, _ arg1: JavaObject?) -> JavaObject?
 }
+extension Attributes {
+  @JavaClass("java.util.jar.Attributes$Name")
+  public struct Name {
```
DougGregor, line 103: This is a wonderful improvement, but I think it should be driven by adding

```
"java.util.jar.Attributes$Name" : "Attributes.Name",
```

to this module's `Java2Swift.config`.

Tests/Java2SwiftTests/Java2SwiftTests.swift
```diff
@@ -114,6 +114,36 @@ class Java2SwiftTests: XCTestCase {
       ]
     )
   }
+
+  func testNestedSubclasses() async throws {
```
DougGregor: I grabbed a different set of test classes out of `java.lang`; love 'em both. Could we drop the unnecessary `async` here?

Sources/Java2Swift/JavaToSwift.swift
jrosen081: I was just following what was there before (there was a try! inside, I just moved it out). I'm fine making this a try

Sources/Java2SwiftLib/JavaTranslator.swift
jrosen081: yep missed that

Tests/Java2SwiftTests/Java2SwiftTests.swift
jrosen081: This got updated in the rebase I think (I had this in a partial working state for a few weeks, but couldn't find the time). updated

Sources/Java2Swift/JavaToSwift.swift
jrosen081: I went back and forth on this a bit. I do like the idea of recurring down nested classes, due to the fact that the majority of uses of nested classes (at least in the JDK) are from the original class (i.e. methods that return nested classes from the main class)). I'm open to either, but like this as making a more complete class translation

DougGregor: It does make the result more complete with less work on the config file, I agree. I have two concerns, but now I'm thinking that they aren't that important and can be addressed in the future.

One is that this means there is no way to disable the generation of the Swift import of a nested class: maybe we never need that, but at least for the moment I've avoided importing some Java classes because they trip up the translation in ways I wasn't able to immediately deal with. That said, if it became a problem, we could invent a way to do this.

The other is in how the output is structured. With this automatic recursion, the emitted file `Attributes.swift` now has the `Attributes.Name` type in it as well, i.e., the whole nested class definition is in one file. With the explicit recursion approach, `Attributes.Name` ends up in a separate file `Attributes+Name.swift`. To do that, the config file needs to list out the `Attributes$Name` class as its own entry so that the Java2SwiftPlugin can know all of the files that will be emitted. It's more tidy for Swift, but it's not strictly necessary... and it would be relatively straightforward in the future to support being explicit in this way in addition to the recursion that you've implemented. For example, we could say that if you have an entry in the Java2Swift.config file for a nested class, we skip that nested class in the recursion because it's been handled at the top level already.

That's all a very, very longwinded way of saying that I agree with you now and will likely implement the "explicit" mechanism I described on top of your PR. Let's merge away!

DougGregor: I updated https://github.com/swiftlang/swift-java/pull/118 to build on top of this. The Jar reader now emits nested classes (which is great), so we have to deal with some nested classes being explicitly-specified in `Java2Swift.config` and others that are discovered through the recursion. Thank you for working on nested classes!

PR #117, opened 2024-10-26
Add Support for Final Fields

### LINKED ISSUE #116: Java2Swift: Handle final fields
Right now, we always import java fields (static or not) with both a `get` and a `set`, but if the java field is `final`, I think we should not create the setter so that we don't try to assign it from Swift.

jrosen081:
Closes #116 

This adds support for final fields (so they are only `get`).

2 parts:
1. New update to the macro which defaults to a non-final field
2. Updates translator to define whether a field is final or not


DougGregor (approved):

DougGregor:
Thank you!

PR #118, opened 2024-10-26
Make nested classes explicitly-specifiable and work through some integration issues

DougGregor:
Implement a follow-on to https://github.com/swiftlang/swift-java/pull/115 that improves things in a few ways:
* Allow `Java2Swift.config` to explicitly specify nested classes, or omit them to get them autogenerated. This "moved" the recursion to make it computed by the outer Java2Swift rather than always walking through all nested classes.
* Get the JavaSieve example working again, now with nested types being listed by the Jar file reader
* More tests!
* Deal with an annoying problem when importing Java nested classes named "Type"; it doesn't work well in Swift because `X.Type` already has a meaning.


ktoso (approved):

DougGregor:
We're going to start with https://github.com/swiftlang/swift-java/pull/115 and I'm going to rebase the few needed changes on top.

Issue #84, opened 2024-10-15
Wrap java.util.function classes nicely in Swift


DougGregor:
We should consider wrapping the Java classes from [`java.util.function`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/function/package-summary.html) as part of JavaKit, either the main module or a `JavaKitFunction` module. We should probably extend those wrapper types with `callAsFunction` overloads so they can be called directly from Swift.

ktoso:
This is related to or maybe even the same as https://github.com/swiftlang/swift-java/issues/101

Issue #101, opened 2024-10-22
jextract: support all other functional interfaces (Supplier and friends) 


ktoso:
Similar how we did `Runnable` in https://github.com/swiftlang/swift-java/pull/100 we should be able to import various shapes of functions like (Int) -> Int and others into the various `j.u.f.` interfaces

- BiConsumer<T,U> 
- BiFunction<T,U,R>
- BinaryOperator<T>
- BiPredicate<T,U> 
- BooleanSupplier 
- Consumer<T> 
- DoubleBinaryOperator
- DoubleConsumer 
- DoubleFunction<R>
- DoublePredicate 
- DoubleSupplier 
- DoubleToIntFunction
- DoubleToLongFunction
- DoubleUnaryOperator 
- Function<T,R> 
- IntBinaryOperator 
- IntConsumer 
- IntFunction<R> 
- IntPredicate 
- IntSupplier 
- IntToDoubleFunction 
- IntToLongFunction
- IntUnaryOperator 
- LongBinaryOperator
- LongConsumer
- LongFunction<R> 
- LongPredicate 
- LongSupplier 
- LongToDoubleFunction 
- LongToIntFunction 
- LongUnaryOperator 
- ObjDoubleConsumer<T> 
- ObjIntConsumer<T> 
- ObjLongConsumer<T> 
- Predicate<T>
- Supplier<T> 
- ToDoubleBiFunction<T,U> 
- ToDoubleFunction<T> 
- ToIntBiFunction<T,U> 
- ToIntFunction<T> 
- ToLongBiFunction<T,U> 
- ToLongFunction<T> 
- UnaryOperator<T>   

we can do the non generic ones first

AmrDeveloper:
I am currently working on this 🙏🏻

PR #119, opened 2024-10-27
Generate JavaKitFunction module

iainsmith:
Add JavaKitFunction module & Makefile action as first steps for #84 

## Java
```java
public Predicate<Integer> lessThanTen() {
  Predicate<Integer> predicate = i -> (i < 10);
  return predicate;
}
```

## Swift

```swift
let predicate: JavaPredicate<JavaInteger> = self.lessThanTen()! 
let value = predicate.test(JavaInteger(3).as(JavaObject.self)) // question about removing need for this cast below?
print("Running a predicate from swift 3 < 10 = \(value)")
```

Updated JavaKitExample available in this commit https://github.com/swiftlang/swift-java/commit/7baa992298f423b06145562f4bbf72a381a63ca1

### Questions

- Is this roughly what you had in mind or do you want to go with a different direction?
- For the `callAsFunction` implementation, should we add those by hand or is it worth updating the generator / config?
- Are you happy if we update the existing JavaKitExample to import `JavaKitCollection` and `JavaKitFunction` so these get run through CI? 
- The generated API doesn't enforce the Generic parameter. Should we update the `@JavaMethod` macro to support handling the cast to JavaObject? That would let us move from: 

```diff
@JavaInterface("java.util.function.Predicate")
public struct JavaPredicate<T: AnyJavaObject> {
  @JavaMethod
-  public func test(_ arg0: JavaObject?) -> Bool
+  public func test(_ arg0: T?) -> Bool

// Callsite

let predicate: JavaPredicate<JavaInteger> = self.lessThanTen()! 
- let value = predicate.test(JavaInteger(3).as(JavaObject.self))
+ let value = predicate.test(JavaInteger(3))
```


ktoso:
Looks fine, PR doesn't seem to include changes to samples or a test though.

Would you be interested in helping out in the other direction as well? We need to support all these in jextract as well: https://github.com/swiftlang/swift-java/issues/101

iainsmith:
> PR doesn't seem to include changes to samples or a test though.

- Do you want a new sample or to add a few functions to the existing sample in JavaKitExample/com/example/swift/HelloSwift.java?
- What's the approach for testing generated modules like these, I couldn't see anything for JavaKitCollection?

> We need to support all these in jextract as well: https://github.com/swiftlang/swift-java/issues/101

I haven't been able to get `make jextract-run` to run yet, using either `main-snapshot-2024-10-25` or `6.0-snapshot-2024-10-12`, but I'd be keen to give it a go as time allows.

Samples/JavaKitSampleApp/Sources/JavaKitExample/JavaKitExample.swift
```diff
@@ -44,6 +45,10 @@ extension HelloSwift: HelloSwiftNativeMethods {
     self.name = "a 🗑️-collected language"
     _ = self.sayHelloBack(42)
 
+    let predicate: JavaPredicate<JavaInteger> = self.lessThanTen()!
+    let value = predicate.test(JavaInteger(3).as(JavaObject.self))
+    print("Running a JavaPredicate from swift 3 < 10 = \(value)")
```
iainsmith, line 50: Not a particular compelling example, but illustrates the module working.

ktoso: I think this is fine, just at least some soundness check that it works at all 👍 

I'd put it into a test file in this project and we're good them IMHO.

ktoso (approved):

PR #121, opened 2024-10-28
Java2Swift: Account for covariant overrides so we don't have duplicates

DougGregor:
When we see methods with the same name and parameter types but different
result types, check whether the result types are related and only keep
the more specialized method.

### DIFF
--- a/Sources/JavaKitCollection/generated/ArrayDeque.swift
+++ b/Sources/JavaKitCollection/generated/ArrayDeque.swift
@@ -26,9 +26,6 @@ public struct ArrayDeque<E: AnyJavaObject> {
   public func clone() -> ArrayDeque<JavaObject>?
 
   @JavaMethod
-  public func clone() throws -> JavaObject?
-
-  @JavaMethod
   public func clear()
 
   @JavaMethod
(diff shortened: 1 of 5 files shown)


Sources/JavaKitCollection/generated/ArrayDeque.swift
```diff
@@ -25,9 +25,6 @@ public struct ArrayDeque<E: AnyJavaObject> {
   @JavaMethod
   public func clone() -> ArrayDeque<JavaObject>?
 
-  @JavaMethod
-  public func clone() throws -> JavaObject?
-
```
DougGregor, line 30: This method is subsumed by the covariant method above it, hence its removal.

DougGregor:
It's a little hard to test the effect of this pull request in isolation, because we don't have a good way to test "does not emit this code", but I've marked where in the generated sources we drop a method because it's been subsumed by a covariant override.

ktoso (approved):
Looks good, nice to capture it in a "collector" 👍

Sources/JavaKitCollection/generated/ArrayDeque.swift
ktoso: Gotcha, makes sense +1

PR #127, opened 2024-10-28
JavaClass macro: stop generating javaThis and javaEnvironment members

DougGregor:
Provide these properties via an extension to AnyJavaObject, accessing the holder values, rather than splatting that same code into every @JavaClass.


ktoso:
lgtm

PR #134, opened 2024-10-29
Add placeholder ordo-benchmark and JMH benchmarks

helmetcrest:
There's only one sample benchmark for each system at the moment, so as to keep the PR scope in check.

There are a couple of known issues:
* I've heard that Gradle doesn't reliably rebuild the benchmark project when a change occurs to a dependency. Locally, I've had it skip running benchmarks when running `gradlew jmh`.
* On Linux x86_64, `swift package benchmark` fails with a `SIL verification failed: Two variables with different type but same scope!` exception in the compiler. I can attach some logs if that'd be useful.

I'm not sure if this PR is too early to merge, but I'd very much appreciate pointers on how to address either the above issues, or anything else that seems unusual in my PR. For instance, my formatter keeps trying to fix indents; if that's too noisy, I can manually remove them for now and open a follow-up PR.

I can confirm that everything passes `act pull_request`, at least, but I needed to add a dependency on `jemalloc` in order to build the Swift benchmarking tool.


.github/workflows/pull_request.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install System Dependencies
-        run: apt-get -qq update && apt-get -qq install -y make curl wget
+        run: apt-get -qq update && apt-get -qq install -y make curl wget libjemalloc2 libjemalloc-dev
```
ktoso, line 32: 👍

Package.swift
```diff
@@ -374,5 +375,23 @@ let package = Package(
         .unsafeFlags(["-I\(javaIncludePath)", "-I\(javaPlatformIncludePath)"])
       ]
     ),
+
+    .executableTarget(
```
ktoso: This won't work, because we'll get the following on macOS:

> error: the executable 'JavaApiCallBenchmarks' requires macos 10.15, but depends on the product 'Benchmark' which requires macos 13.0; consider changing the executable 'JavaApiCallBenchmarks' to require macos 13.0 or later, or the product 'Benchmark' to require macos 10.15 or earlier.

The way to do this is to make a directory `Benchmarks/` and make a new `Package.swift` in there, and refer it back into `..` for the library dependency

ktoso: Here's an example how to set it up: https://github.com/apple/swift-nio/blob/main/Benchmarks/Package.swift

Samples/SwiftKitSampleApp/build.gradle
```diff
@@ -16,6 +16,7 @@ import org.swift.swiftkit.gradle.BuildUtils
 
 plugins {
     id("build-logic.java-application-conventions")
+    id "me.champeau.jmh" version "0.7.2"
```
ktoso: match the style please with the `()` :)

Samples/SwiftKitSampleApp/build.gradle
```diff
@@ -31,43 +32,38 @@ java {
...
 
-    // Make the 'main' and 'test' source sets depend on the generated sources
-    main {
-        compileClasspath += sourceSets.generated.output
-        runtimeClasspath += sourceSets.generated.output
-    }
-    test {
-        compileClasspath += sourceSets.main.output
-        runtimeClasspath += sourceSets.main.output
+def jextract = tasks.register("jextract", Exec) {
```
ktoso, line 36: ok thx, I think we'll need to make it into a gradle plugin very soon

Benchmarks/Package.swift
```diff
@@ -0,0 +1,69 @@
+// swift-tools-version: 6.0
+
+import PackageDescription
+
+import class Foundation.FileManager
+import class Foundation.ProcessInfo
+
+// Note: the JAVA_HOME environment variable must be set to point to where
```
helmetcrest, line 8: I unfortunately had to copy this out of `/Package.swift` since it's relied upon for setting include paths in `JavaApiCallBenchmarks`.

Package.swift
helmetcrest: I think I've set it up correctly now -- thanks for the example file! It helped a bunch.

Benchmarks/Package.swift
ktoso: yea we have to copy this around, thx!

Benchmarks/Package.swift
```diff
@@ -0,0 +1,69 @@
...
+  targets: [
+    .executableTarget(
+      name: "JavaApiCallBenchmarks",
+      dependencies: [
+        .product(name: "JavaRuntime", package: "swift-java"),
+        .product(name: "JavaKit", package: "swift-java"),
+        .product(name: "JavaKitNetwork", package: "swift-java"),
+        .product(name: "Benchmark", package: "package-benchmark"),
+      ],
+      path: "Benchmarks/JavaApiCallBenchmarks",
```
ktoso, line 59: Cool structure is looking good now

Makefile
```diff
@@ -133,12 +133,12 @@ jextract-generate: jextract-swift generate-JExtract-interface-files
...
 		--package-name com.example.swift.generated \
 		--swift-module ExampleSwiftLibrary \
-		--output-directory ${SAMPLES_DIR}/SwiftKitSampleApp/src/generated/java \
+		--output-directory ${SAMPLES_DIR}/SwiftKitSampleApp/build/generated/sources/jextract/main \
 		$(BUILD_DIR)/jextract/ExampleSwiftLibrary/MySwiftLibrary.swiftinterface; \
 	swift run jextract-swift \
 		--package-name org.swift.swiftkit.generated \
 		--swift-module SwiftKitSwift \
-		--output-directory ${SAMPLES_DIR}/SwiftKitSampleApp/src/generated/java \
+		--output-directory ${SAMPLES_DIR}/SwiftKitSampleApp/build/generated/sources/jextract/main \
```
ktoso, line 141: makes sense!

ktoso:
This all looks excellent, thank you! 

I only added a small note into the readme how to run the benchmarks :)

ktoso:
I have now included building (not _running_) the benchmarks in CI as well so we don't regress them by accident.

PR #138, opened 2024-10-30
[WIP] JExtractSwiftPlugin

ktoso:
🚧 **early work in progress** 🚧 

This is first steps to make an jextract-swift plugin.

If we're able to pull this off as a swiftpm plugin we would basically have a way to "make me a jar that exposes this swift module, that would be super simple to consume from Java.

Need to figure out how to actually make this work -- it can't be a build plugin, but preBuild also isn't quite enough.

We may have to make it a command plugin, since we need compilation OUTPUT of the module to then act on the swift interfaces.


Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -0,0 +1,80 @@
...
+      "--output-directory", outputDirectory.path(percentEncoded: false),
+      // TODO: "--build-cache-directory", ...
+      //       Since plugins cannot depend on libraries we cannot detect what the output files will be,
+      //       as it depends on the contents of the input files. Therefore we have to implement this as a prebuild plugin.
+      //       We'll have to make up some caching inside the tool so we don't re-parse files which have not changed etc.
+    ]
+    arguments.append(sourceDir)
+    
+    return [
+      .prebuildCommand(
```
ktoso, line 53: this can't work, since we need this to act on swiftinterface files of the module it's acting on. It has to be a "post build" plugin really.

ktoso: TODO: make this into a command plugin people would have to invoke manually `swift package jextract` which would generate the bindings.

We could also then `swift package jextract jar` to produce a jar or even publish it etc.

ktoso:
Before weekend snapshot:

- we now generate swift thunks (!)
- first thunk is to "give me the Any.Type of this type" rather than using swift interfaces
    - these are done via `public func swiftjava_getType_JExtractPluginSampleLib_MyCoolSwiftClass() -> Any /* Any.Type */` generated thunks
- jextract is now a prebuild plugin so we can get seamless build integration


Screenshot 2024-11-01 at 19 16 14


Basically this means you "just" write some swift sources, add the extract plugin to the given target. Add a small task in gradle to trigger swiftpm and this way gradle/java immediately see the wrappers. Upon changing swift -> gradle triggers swift build again and it's all very seamless 🥳

Issue #31, opened 2024-10-01
jextract-swift support for Swift Optional


ktoso:
Look into what we need to do to handle 

```
swift
func get() -> Int?
func take(_: Int?)
```

The exposed Java interface should be using `Optional<T>` OR [OptionalInt](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/OptionalInt.html) and its friends: [OptionalDouble](https://docs.oracle.com/javase/8/docs/api/java/util/OptionalDouble.html), [OptionalLong](https://docs.oracle.com/javase/8/docs/api/java/util/OptionalLong.html)

madsodgaard:
We now have support for this in both JNI and FFM.

PR #139, opened 2024-10-30
jextract import optional types

ktoso:
Half of https://github.com/swiftlang/swift-java/issues/31

```
public func globalGetIntOptional() -> Int?
```

becomes:

```
      public static java.util.OptionalLong globalGetIntOptional()
```

other types of Optional are also supported, including any `T`

Needs more tests for all the cases and also be able to pass them in parameters.


Sources/JExtractSwift/TranslatedType.swift
```diff
@@ -23,10 +23,30 @@ extension Swift2JavaVisitor {
...
+          originalSwiftType: optionalType.wrappedType, // FIXME: just optionalType?
+          cCompatibleSwiftType: "OptionalLong",
+          cCompatibleJavaMemoryLayout: .heapObject,
+          javaType: .javaUtilOptionalLong
+        )
+      } else if let translatedWrappedType = try? cCompatibleType(for: optionalType.wrappedType) {
+        return TranslatedType(
+          cCompatibleConvention: .direct,
+          originalSwiftType: optionalType.wrappedType, // FIXME: just optionalType?
+          cCompatibleSwiftType: "Optional<TODO>",
```
ktoso, line 43: FIXME

ktoso:
On purpose pushed a partial WIP, I want to get back to this once we have an jextract plugin

ktoso:
This will get reworked a bit with a fresh start

Issue #132, opened 2024-10-29
Java2Swift: Move from Swift structs over to Swift classes


DougGregor:
Right now, Java classes (and interfaces) are imported as Swift `struct` declarations, which means we don't get sub typing relationships. Let's see if we can move over to using Swift `class` declarations instead so the Java class hierarchy is reflected in the Swift class hierarchy. There are lots of issues to consider here.

PR #140, opened 2024-10-31
JavaKit: Macro and translator support for mapping Java classes to Swift classes

DougGregor:
Java classes are currently translated in Swift structs, and we use is/as
functions to cast up and down the hierarchy. Start down an alternative
path where we translate Java classes into Swift classes, making proper use
of inheritance in Swift to reflect inheritance in Java.

This step updates the @javaclass macro implementation to check whether
it is being applied to a class (rather than a struct) and adjust its
behavior accordingly:

* The superclass is extracted from the inheritance clause of the class
rather than the "extends" argument to the macro.
* The "full Java class name" static member becomes a "class" member.
For everything other than JavaObject itself, it is an override.
* The `javaHolder` property is only emitted into JavaObject, nowhere else.
* The `init(javaHolder:)` initializer becomes required. For everything
other than JavaObject, it's implemented as a super.init call.
* The `JavaSuperclass` typealias is no longer emitted.

All of this is keyed off the subject of the attribute being a "class"
rather than a "struct", so we keep the existing struct versions working.

Relatedly, extend the translator with an option to map Java classes to
Swift classes. This involves a number of changes:

* Emitting each as an "open class" rather than "public struct"
* Emitting the superclass into the inheritance clause rather than the "extends" argument
* Emitting methods as "open" rather than "public", and "override" when we are overriding a method declared in a superclass
* Emitting initializers as "convenience" since we still always build a Java object behind the scenes
* Only emit methods and fields declared in the class, not inherited ones

This option is only currently only enabled in test cases while we stage in this functionality.

This is for issue https://github.com/swiftlang/swift-java/issues/132.


ktoso:
Hm, definitely promising direction. I like that the object holder is in the super class, that makes sense. `open` also makes sense though maybe I'm unnecessarily worried about actually using this ability to subclass those 😅 

So far looking good :)

Tests/Java2SwiftTests/Java2SwiftTests.swift
```diff
@@ -314,7 +314,7 @@ class Java2SwiftTests: XCTestCase {
         """,
         """
           @JavaMethod
-          public convenience init(environment: JNIEnvironment? = nil)
+          @_nonoverride public convenience init(environment: JNIEnvironment? = nil)
```
ktoso, line 317: hah sneaky 👍

ktoso:
Looks good 👍

ktoso (approved):

PR #144, opened 2024-10-31
Draft PR: tracking changes for Android JVM integration

lhoward:


ktoso:
Let us know when ready to review; I see this included the `.idea` folder, so please make sure to .gitignore that as well 👍

lhoward:
New to Android and Kotlin, so still figuring out what is ignorable :) Shall fix.

DougGregor:
This is great! The changes to the common code are quite narrow, which is quite encouraging.

Sources/JavaRuntime/AndroidSupport.cpp
```diff
@@ -34,7 +34,7 @@ static JavaRuntime_GetCreatedJavaVMs_fn JavaRuntime_GetCreatedJavaVMs;
 static void *JavaRuntime_dlhandle;
 
 __attribute__((constructor)) static void JavaRuntime_init(void) {
-  JavaRuntime_dlhandle = dlopen("libnativehelper.so", RTLD_NOW | RTLD_GLOBAL);
+  JavaRuntime_dlhandle = dlopen("libnativehelper.so", RTLD_NOW | RTLD_LOCAL);
```
DougGregor: Question about this! Should we link the JavaRuntime target to `libnativehelper.so` when building for Android rather than doing the `dlopen` here?

lhoward: Hmm, I'd assumed this wasn't possible because the Android NDK only provides the `.so`, and typically on Linux `ld` will only find a versioned `.so`. But maybe this isn't the case on Android. I'll check and get back.

lhoward: Actually I may have that around the wrong way, you need the unversioned `.so` at `ld` time, and the versioned one at runtime. So yes, checking I can link against it now.

lhoward: OK, this is possible, but requires setting the NDK target to >=31:

```
        .unsafeFlags(["--target=aarch64-unknown-linux-android31"], .when(platforms: [.android])),
```

Obviously I don't want to encode the target architecture, do you know offhand if this is surfaced?

lhoward: * Actually this is a bit harder than it seems, for reasons I'll explain in a comment later.

lhoward:
Regarding linking rather than dynamically loading `libnativehelper.so`, we'll need upstream Swift Android SDK support (finagolfin/swift-android-sdk#191) for building against arbitrary Android API levels. If we force the API level to one that includes `libnativehelper.so` then the Swift runtime libraries are not found.

finagolfin:
My new 6.1 SDK bundle supports building against all Android APIs between 24-35, try it out and let me know how it works.

ktoso:
With https://github.com/swiftlang/swift-java/pull/401 merged I'm going to close this one; We did merge some minimal support

Issue #143, opened 2024-10-31
Android JVM does not allow NewGlobalRef whilst exception pending


lhoward:
Relates to Android branch [here](https://github.com/swiftlang/swift-java/compare/main...PADL:swift-java:lhoward/android).

Android JNI does not allow `NewGlobalRef()` to be called whilst an exception is pending, which is causing the following error when trying to raise an exception from the JVM side:

```
JNI DETECTED ERROR IN APPLICATION: JNI NewGlobalRef called with pending exception java.lang.Exception: I am an error
   at void com.example.swift.HelloSwift.throwMessage(java.lang.String) (HelloSwift.kt:58)
   at int com.example.swift.HelloSwift.sayHello(int, int) (HelloSwift.kt:-2)
   at java.lang.String com.example.swift.HelloSwift.getGreeting() (HelloSwift.kt:26)
   at java.lang.String com.example.swift.HelloSwift.access$getGreeting(com.example.swift.HelloSwift) (HelloSwift.kt:17)
   at void com.example.swift.HelloSwift$onCreate$1$1$1.invoke(androidx.compose.foundation.layout.PaddingValues, androidx.compose.runtime.Composer, int) (HelloSwift.kt:68)
```

`NewGlobalRef()` is called by `JavaObjectHolder`.

I haven't done any further investigation.

lhoward:
This appears to fix issue:

```diff
diff --git a/Sources/JavaKit/Exceptions/ExceptionHandling.swift b/Sources/JavaKit/Exceptions/ExceptionHandling.swift
index efd98e9..cf063d0 100644
--- a/Sources/JavaKit/Exceptions/ExceptionHandling.swift
+++ b/Sources/JavaKit/Exceptions/ExceptionHandling.swift
@@ -20,9 +20,8 @@ extension JNIEnvironment {
 
     // Check whether a Java exception occurred.
     if let exception = interface.ExceptionOccurred(self) {
-      let throwable = Throwable(javaThis: exception, environment: self)
       interface.ExceptionClear(self)
-      throw throwable
+      throw Throwable(javaThis: exception, environment: self)
     }
```

PR #145, opened 2024-10-31
JavaKit: clear exception before initializing Throwable

lhoward:
Initializing Throwable with an exception calls the JNI method NewGlobalRef, which on Android cannot be called while an exception is pending. [1] Clear the exception first before initializing the Throwable.

[1] https://developer.android.com/training/articles/perf-jni.html#exceptions_1


ktoso (approved):
Looks ok to me 👍

lhoward:
Closes #143

PR #148, opened 2024-11-02
Java2Swift: Swift does not support overriding with covariant array result types

DougGregor:


lhoward:
works for me, thank you!

Issue #136, opened 2024-10-30
javakit: improve error messages when class missing from classpath


ktoso:
Today it seems we just:

```
Error: org/gradle/tooling/GradleConnector
```

which you can guess what it means but it'll be better if we spell out "Class not found" etc

lhoward:
Also seeing the same thing. Can I assume this means class not found? Do I need to explicitly add the current directory to `classPath` when mapping a class that's defined within a Swift target? Can a single Swift module wrap both classes from an external JAR and those that are located within it?

ktoso:
You'll have to build a "proper" classpath, that's an adventure in of itself but I'm sure we can figure it out.

Most of the time such message is a class not found, but can't guarantee it's just that, that's what this issue is about, making those errors better

lhoward:
OK, so maybe I need to fix this (GH) issue to fix my actual issue :)

Should Swift packages wrapping external JARs be in distinct packages from those that have Java classes (with native methods) inside them?

lhoward:
Can you give me a hint as to where this error message is emitted from? `git grep` offers no hints.

ktoso:
> Should Swift packages wrapping external JARs be in distinct packages from those that have Java classes (with native methods) inside them?

I don't think we have a complete opinion formed here yet. Can you explain in more detail the question. There's various "wrappings" that make sense, for example "a jar that contains java code including native dylib that it calls into (either using JNI or FMM (jextract-swift)" is definitely a thing, mixed language projects eventually too... but this isn't figured out yet. What specifically are you asking about here?

ktoso:
> Can you give me a hint as to where this error message is emitted from? git grep offers no hints.
That's part of this issue I'm afraid 😅 It might be during a `FindClass` call perhaps? Idk if the Java runtime does the print here or who does, we don't seem to have such `Error: ...` print at least.

You could also launch your jvm with `lldb -- java -cp ... <main>` and this way you could maybe find where this happens by stepping through a sample app if nothing else is helping.

lhoward:
> > Should Swift packages wrapping external JARs be in distinct packages from those that have Java classes (with native methods) inside them?
> 
> I don't think we have a complete opinion formed here yet. Can you explain in more detail the question. There's various "wrappings" that make sense, for example "a jar that contains java code including native dylib that it calls into (either using JNI or FMM (jextract-swift)" is definitely a thing, mixed language projects eventually too... but this isn't figured out yet. What specifically are you asking about here?

[...]

PR #150, opened 2024-11-03
JavaKit: check each classPath element exists before starting VM

lhoward:
Partial fix for #136: validate each classPath element by attempting to open it before starting VM. It would be more efficient to stat() it but SystemPackage lacks an abstraction for checking if a file exists, and we don't want to import Foundation.

### DIFF
--- a/Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
+++ b/Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
@@ -12,6 +12,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
 typealias JavaVMPointer = UnsafeMutablePointer<JavaVM?>
 
 public final class JavaVirtualMachine: @unchecked Sendable {
@@ -54,6 +60,12 @@ public final class JavaVirtualMachine: @unchecked Sendable {
     // Construct the complete list of VM options.
     var allVMOptions: [String] = []
     if !classPath.isEmpty {
+      let fileManager = FileManager.default
+      for path in classPath {
+        if !fileManager.fileExists(atPath: path) {
+          throw JavaKitError.classPathEntryNotFound(entry: path, classPath: classPath)
+        }
+      }
       let colonSeparatedClassPath = classPath.joined(separator: ":")
       allVMOptions.append("-Djava.class.path=\(colonSeparatedClassPath)")
     }
@@ -267,5 +279,9 @@ extension JavaVirtualMachine {
       default: self = .unknown(error)
       }
     }
+  }
+
+  enum JavaKitError: Error {
+    case classPathEntryNotFound(entry: String, classPath: [String])
   }
 }


lhoward:
Todo: check open-ing a directory is portable (specifically on Windows).

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
```diff
@@ -54,6 +56,14 @@ public final class JavaVirtualMachine: @unchecked Sendable {
     // Construct the complete list of VM options.
     var allVMOptions: [String] = []
     if !classPath.isEmpty {
+      for path in classPath {
+        do {
+          let fd = try FileDescriptor.open(path, .readOnly)
+          try? fd.close()
```
ktoso: Can't we do this without a new dependency? I don't love FileManager but it does have such API: `FileManager.default.fileExists(atPath: path)` so we con't have to include another dependency (yet).

https://developer.apple.com/documentation/foundation/filemanager/1410277-fileexists

Package.swift
```diff
@@ -161,7 +162,12 @@ let package = Package(
     ),
     .target(
       name: "JavaKit",
-      dependencies: ["JavaRuntime", "JavaKitMacros", "JavaTypes"],
+      dependencies: [
+        "JavaRuntime",
+        "JavaKitMacros",
+        "JavaTypes",
+        .product(name: "SystemPackage", package: "swift-system")
```
ktoso: see comment below about using foundation instead of new dependency

lhoward:
For sure, just thought avoiding Foundation was preferable. This won't pull in ICU?

lhoward:
OK, use FoundationEssentials where possible.

lhoward:
```
  enum JavaKitError: Error {
    case classPathNotFound(String)
  } 
```

maybe we move VMError under this?

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
```diff
@@ -12,6 +12,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
```
ktoso, line 19: Yup that's the way, thank you

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
```diff
@@ -54,6 +60,12 @@ public final class JavaVirtualMachine: @unchecked Sendable {
     // Construct the complete list of VM options.
     var allVMOptions: [String] = []
     if !classPath.isEmpty {
+      let fileManager = FileManager.default
+      for path in classPath {
+        if !fileManager.fileExists(atPath: path) {
+          throw JavaKitError.classPathNotFound(path)
```
ktoso:
```suggestion
          throw JavaKitError.classPathSegmentNotFound(notFound: path, classpath: classPath)
```

It's not that the classpath isn't found, just that one of its segments/paths isnt, shall we rename it like this?

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
```diff
@@ -268,4 +280,8 @@ extension JavaVirtualMachine {
       }
     }
   }
+
+  enum JavaKitError: Error {
+    case classPathNotFound(String)
```
ktoso:
```suggestion
    case classPathSegmentNotFound(notFound: String, classpath: [String])
```

ktoso (approved):

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
lhoward: Yes, let's just check `segment` is the right word, I'll look at the JDK source to see if it suggests anything (unless you know this for sure?)

lhoward: I think `entry` is used in the JDK.

ktoso: Good catch, sounds good!

lhoward: And just to confirm you don't want to make `VMError` a case of `JavaKitError`?

ktoso: Either is fine tbh, whichever you prefer. I don’t think we expect people to handle them specifically

ktoso:
Thanks for dealing with the nitpicks :)

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
```diff
@@ -268,4 +280,8 @@ extension JavaVirtualMachine {
       }
     }
   }
+
+  enum JavaKitError: Error {
+    case classPathEntryNotFound(entry: String, classPath: String)
```
ktoso:
```suggestion
    case classPathEntryNotFound(entry: String, classPath: [String])
```

ktoso: Fixing compilation

ktoso:
If you do the "maintainers can edit" on the right I can more easily fix simple stuff like that :) There's a simple compilation error, see proposed fix.

Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
lhoward: Oh, it was supposed to be `colonSeparatedClassPath` – must have dropped that diff somehow. But I suppose the colon separation is an implementation detail. I'll do as you suggest above.

lhoward:
> If you do the "maintainers can edit" on the right I can more easily fix simple stuff like that :) There's a simple compilation error, see proposed fix.

Hmm, weirdly I can't seem to find this checkbox...

ktoso:
Hmm, don't see something like that? (pic from a different PR that I made) Screenshot 2024-11-05 at 12 32 08

ktoso:
Either way, thank you merged :)

lhoward:
> Hmm, don't see something like that? (pic from a different PR that I made) Screenshot 2024-11-05 at 12 32 08

I think it might be because my fork was in my company repo (PADL) rather than personal. It's a bit hard to parse [1].

[1] https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#

ktoso:
Huh interesting, didn't realize this limitation.

Issue #154, opened 2024-11-06
JavaKit: Class loader hints for Android


lhoward:
Android JVM class resolution is tricky, as which classes are resolvable depends not only on the JNI environment but also the stack frame.

From https://developer.android.com/training/articles/perf-jni:

>You can get into trouble if you create a thread yourself (perhaps by calling pthread_create and then attaching it with AttachCurrentThread). Now there are no stack frames from your application. If you call FindClass from this thread, the JavaVM will start in the "system" class loader instead of the one associated with your application, so attempts to find app-specific classes will fail.
>
>There are a few ways to work around this:
>
>* Do your FindClass lookups once, in JNI_OnLoad, and cache the class references for later use. Any FindClass calls made as part of executing JNI_OnLoad will use the class loader associated with the function that called System.loadLibrary (this is a special rule, provided to make library initialization more convenient). If your app code is loading the library, FindClass will use the correct class loader.
>* Pass an instance of the class into the functions that need it, by declaring your native method to take a Class argument and then passing Foo.class in.
>* Cache a reference to the ClassLoader object somewhere handy, and issue loadClass calls directly. This requires some effort.

Doing `FindClass()` lookups in `JNI_OnLoad` ~~isn't really viable unless JavaKit caches classes and we can ’poison’ the cache by resolving any classes we will need there. It's~~ is not ideal though as it requires a priori knowledge of all the application classes (I suppose, this could be done with a macro). Conversely, passing an instance of the class into functions that need it doesn't make for a great API. Neither of these is particularly ergonomic because `getJNIClass()` can't be overridden (it's in an extension), so one ends up needing to reimplement `dynamicJavaNewObjectInstance()`.

Caching a reference to the ClassLoader object might be the best approach, this can be done in `JNI_OnLoad` (as long as you know the name of at least one application class), and then perhaps it could be an additional optional argument to initialisers. Or `JNIEnvironment` becomes a class to which class loaders can be attached.

Anyway, just flagging this as food for thought.

lhoward:
As a test, I cached the class loader in `JNI_OnLoad`, and implemented my own `FindClass()` that used it. (I thought I might be able to swizzle the JNI env, which whilst ugly would be quite convenient, but no dice – not writable memory.)

This gets me a few steps further but plenty of things like `as()` are broken. So this is a bit of a tricky one. I think this is going to come up in any non-trivial Android application because you're likely going to need to allocate Java objects from the Swift side.

lhoward:
[...]

PR #159, opened 2024-11-07
JavaKit: add CustomJavaClassLoaderConstructible protocol

lhoward:
Swift projections of Java classes can conform to
CustomJavaClassLoaderConstructible, to provide a custom class loader to be used when constructing new instances.

Fixes: #154


Tests/Java2SwiftTests/Java2SwiftTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import JavaKit
+@testable import JavaKit
```
DougGregor: Could we use `@_spi(Testing)` here instead of `@testable`?

DougGregor (commented):
This is a great addition, thank you!

DougGregor:
> Bikeshed: `CustomJavaClassLoaderConstructible` or `CustomJavaClassLoaderInitializable`?

I'd prefer to avoid `Constructible`, since that's not really a thing in Swift. How about just `CustomJavaClassLoader`, as in, this type has a custom Java class loader?

DougGregor:
@lhoward anything else you want to do here before we merge?

lhoward:
> @lhoward anything else you want to do here before we merge?

I'm happy for you to merge, unless you want tests (which I am still working on).

DougGregor:
I do like tests. Let's wait until you have some tests in place.

lhoward:
I'm a bit caught up with other stuff today, I'll endeavour to get onto tests next week, feel free to merge (or not) in the interim.

DougGregor:
Since this is also needed in the other PR, I'll go ahead and merge.

PR #161, opened 2024-11-08
ExampleSwiftLibrary: adjust imports for Windows

compnerd:
Guard the `Darwin` import to when the module is available rather than assuming it to be the default. Add a case for Windows.


ktoso (approved):

ktoso (approved):

ktoso:
Nightly snapshot toolchain regressed the build for us, will fix it.

PR #163, opened 2024-11-08
build: adjust the linker flags for JavaKit

compnerd:
`-rpath` is not a universally valid flag for the linker. Adjust the linker flags to add the library search path on Windows and elide the `-rpath` flag. It is unclear why there is no `server` subdirectory under `lib` when using the Oracle JDK 23, but this allows us to find the necessary library.


Package.swift
```diff
@@ -174,8 +174,10 @@ let package = Package(
             "-rpath",
             "-Xlinker",
             "\(javaHome)/lib/server",
-          ]
+          ],
+          .when(platforms: [.linux, .macOS])
         ),
+        .unsafeFlags(["-L\(javaHome)/lib"], .when(platforms: [.windows])),
```
ktoso:
```suggestion
        .unsafeFlags(
          [
            "-L\(javaHome)/lib"
          ], 
          .when(platforms: [.windows])),
```

ktoso (approved):
lgtm, just formatting it the same way as above snippet

PR #164, opened 2024-11-08
build: elide the JavaKitMacroTests

compnerd:
The way that SPM builds executable tests is not supportable on Windows. We end up with duplicate definitions of `main` which is a hard error on Windows. Disable this test suite to allow us to build.


ktoso:
Closing for now, seems ot be a draft not revisited in a while. We can revisit later

Issue #152, opened 2024-11-04
JavaCompiler:  need flexible argument passing


lhoward:
For example, the Android `javac` defaults to targeting Java SE 21 which isn't compatible with Kotlin. We need to be able to affect the `--target` command line option to `javac` (and by extension other options).

ktoso:
> really I want to pass this as an option to plugin in Package.swift.
>  
> [forums.swift.org/t/swift-package-buildtool-plugin-cannot-accept-arguments-nor-environment-variables/59046](https://forums.swift.org/t/swift-package-buildtool-plugin-cannot-accept-arguments-nor-environment-variables/59046) suggests that it's not possible?

It's a known limitation; we have to work around it with config files, I suggest you add config options to the existing config files.

We'll converge into a single `swiftjava.config` file soon, that would contain config for all the parts.

So for now just add things into the Java2Swift.config

lhoward:
> > really I want to pass this as an option to plugin in Package.swift.
> > [forums.swift.org/t/swift-package-buildtool-plugin-cannot-accept-arguments-nor-environment-variables/59046](https://forums.swift.org/t/swift-package-buildtool-plugin-cannot-accept-arguments-nor-environment-variables/59046) suggests that it's not possible?
> 
> It's a known limitation; we have to work around it with config files, I suggest you add config options to the existing config files.
> 
> We'll converge into a single `swiftjava.config` file soon, that would contain config for all the parts.
> 
> So for now just add things into the Java2Swift.config

OK, this sounds good.

lhoward:
OK with `targetCompatibility` and `sourceCompatibility` keys with integer values?

PR #166, opened 2024-11-08
JavaKit: add source/target versions to Java2Swift.config

lhoward:
Fixes: #152

### DIFF
--- a/Plugins/JavaCompilerPlugin/Configuration.swift
+++ b/Plugins/JavaCompilerPlugin/Configuration.swift
@@ -12,34 +12,40 @@
 //
 //===----------------------------------------------------------------------===//
 
-package typealias JavaVersion = Int
+typealias JavaVersion = Int
 
 /// Configuration for the Java2Swift translation tool, provided on a per-target
 /// basis.
 ///
-/// Note: there is a copy of this struct in the Java2Swift plugin. They
+/// Note: there is a copy of this struct in the Java2Swift library. They
 /// must be kept in sync.
-package struct Configuration: Codable {
+struct Configuration: Codable {
   /// The Java class path that should be passed along to the Java2Swift tool.
-  package var classPath: String? = nil
+  var classPath: String? = nil
 
   /// The Java classes that should be translated to Swift. The keys are
   /// canonical Java class names (e.g., java.util.Vector) and the values are
   /// the corresponding Swift names (e.g., JavaVector).
-  package var classes: [String: String] = [:]
+  var classes: [String: String] = [:]
 
-  package var sourceCompatibility: JavaVersion?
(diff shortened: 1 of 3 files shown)


Plugins/JavaCompilerPlugin/Configuration.swift
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+typealias JavaVersion = Int
+
+/// Configuration for the Java2Swift translation tool, provided on a per-target
+/// basis.
+///
+/// Note: there is a copy of this struct in the Java2Swift library. They
```
ktoso, line 20: Since we are forced into this because plugins can’t have dependencies, I wonder if we need to set up a ci task to check those are indeed the same… we can do that 🤔 

But maybe later on, as the project becomes more stable.

lhoward: Yeah, or maybe a makefile target to copy them from a SSOT.

ktoso: I’d love to not have to use make in this build but I guess for such stuff maybe we will still… just because we already have swift gradle AND make so it’d be nice to just have two builds to worry about…

but i think it’d be okey to use make just for some simple automation but let actual “building” to swiftpm and gradle. So your idea here may be good :)

ktoso: Either way, right now the copy is fine

ktoso:
Sorry for the delay, I'll get CI green tomorrow and merge all such pending PRs. Since we use nightly toolchain it regressed us accidentally

Plugins/JavaCompilerPlugin/JavaCompilerPlugin.swift
```diff
@@ -32,6 +32,13 @@ struct JavaCompilerBuildToolPlugin: BuildToolPlugin {
     // so we cannot eliminate this deprecation warning.
     let sourceDir = target.directory.string
 
+    // The name of the configuration file JavaKit.config from the target for
+    // which we are generating Swift wrappers for Java classes.
+    let configFile = URL(filePath: sourceDir)
+      .appending(path: "Java2Swift.config")
+    let configData = try Data(contentsOf: configFile)
+    let config = try JSONDecoder().decode(Configuration.self, from: configData)
```
lhoward: Hmm, we probably want to make the config file optional for JavaCompiler. Let me fix this.

ktoso: makes sense yeah

Issue #172, opened 2024-11-09
JavaKit: thread detach interactions with implicitly created threads


lhoward:
Considering the not uncommon case where an application uses structured concurrency and some or all of the executors use libdispatch. The application does not manage thread creation and destruction explicitly. (Naturally are other reasons to use libdispatch, thread pools, etc, but this case is noteworthy as it is generally imported by the language runtime itself.)

JNI's `AttachCurrentThread()` is called implicitly by JavaKit when retrieving the environment (see also #157) but `DetachCurrentThread()` must currently be called explicitly via its `JavaVirtualMachine` wrapper. Failing to detach the thread can cause bad things to happen (leaking local references, thread monitors hanging, etc).

I suspect, on attach, we should register a dummy TSD with a cleanup handler that will detach the current thread from the JVM, on thread exit or cancellation. Whilst this would require the introduction of some non-portable code I think it speaks to leaving things the ’way we found them’.

(Interestingly side note – I see libdispatch has `_dispatch_install_thread_detach_callback()` for Android only, anyone know the provenance of this?)

ktoso:
> (Interestingly side note – I see libdispatch has _dispatch_install_thread_detach_callback() for Android only, anyone know the provenance of this?)

Okey this may save the day for Android specifically... I don't know much about it though but if it's there in corelibs dispatch we can use it I think. I'll ask around a bit.

lhoward:
> > (Interestingly side note – I see libdispatch has _dispatch_install_thread_detach_callback() for Android only, anyone know the provenance of this?)
> 
> Okey this may save the day for Android specifically... I don't know much about it though but if it's there in corelibs dispatch we can use it I think. I'll ask around a bit.

I wonder it may have been used for some thing Apple-private as I can't find any references in the open source Foundation.

In any case, if we have the opportunity to call `_dispatch_install_thread_detach_callback()`, we also have the opportunity to call `pthread_setspecific()` (or equivalent), and presumably we should favour the more portable API.

ktoso:
Hmm, yeah I think we can do that with just the callbacks to pthread_setspecific, but I'll check in if there's better reasons for the detach callback or not.

lhoward:
Yeah, `_dispatch_install_thread_detach_callback()` could be called at program startup I suppose, it's more like `__cxa_thread_atexit()`. But I don't think that particularly helps us given JavaKit attaches on demand.

[...]

Issue #157, opened 2024-11-07
Passing objects between threads


lhoward:
`javaMethodLookup()` uses the environment attached to the object, probably it should use the ambient (thread-specific) environment.

```
java_vm_ext.cc:599] JNI DETECTED ERROR IN APPLICATION: thread Thread[23,tid=1783,Native,Thread*=0xb40000718be7eb40,peer=0x1449d000,"Thread-4"] using JNIEnv* from thread Thread[1,tid=1631,Native,Thread*=0xb40000718be307b0,peer=0x72c1eb68,"main"]
                                                                                                    java_vm_ext.cc:599]     in call to GetObjectClass
```

lhoward:
It's not a huge issue because I can schedule things on the main thread but, would be nice to not need to do this.

lhoward:
NB: we can also do something Android-specific if necessary, as Android guarantees only one JVM per process.

lhoward:
I just did this for now – https://github.com/PADL/swift-java/commit/2d8c5955d73dce199a343872f05c3efc63e8518c – but obviously it's a workaround. I think I need upstream input as to what the right solution is, whether we remove `javaEnvironment` completely and always use the ambient environment.

PR #173, opened 2024-11-09
Detach from any threads we have attached to.

lhoward:
Fixes: #172

### DIFF
--- a/Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
+++ b/Sources/JavaKit/JavaKitVM/JavaVirtualMachine.swift
@@ -24,6 +24,11 @@ public final class JavaVirtualMachine: @unchecked Sendable {
   /// The JNI version that we depend on.
   static let jniVersion = JNI_VERSION_1_6
 
+  /// Thread-local storage to detach from thread on exit
+  private static let destroyTLS = ThreadLocalStorage { _ in
+    try? JavaVirtualMachine.shared().detachCurrentThread()
+  }
+
   /// The Java virtual machine instance.
   private let jvm: JavaVMPointer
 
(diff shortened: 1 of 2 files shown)


ktoso:
Oh boy, this will be quite some trouble outside of simple examples (where we'll have to deal with the fact that `Task{}` and friends exist)... because if we can't make JNI calls from a thread that did not register, technically after every suspension point we may be on a different thread, so we may need to keep trying to register threads.

We also have no control over dispatch deciding to start/stop threads, so we may leave lingering resources since we'd never detach some of the threads hm.

I think in the long term we're going to need to write an executor, make it the global and main executor and there we'd have to control threads and this way we'd be able to detach them consistently... 🤔 We are considering offering APIs to customize the global default executors so this may be possible nicely in the future, but that's not until next Swift release hm.

Meanwhile I'll think about your idea here.

lhoward:
This relates to #157 – if we remove the affinity between projected objects and environments, then we can always attach a thread on demand (and register a detach destructor ). I'm not sure what the right thing to do is when there are multiple VMs, but the JNI spec is quite clear that environments are pre-thread (FWIW, Android only ever has a single VM).

lhoward:
Given the necessary hooks in the Swift runtime (I haven't looked too closely), we could definitely extend our [AExecutor](https://github.com/PADL/AndroidLooper/blob/main/Sources/AndroidLooper/AExecutor.swift) to be a concurrent executor. But whilst idiomatic, a platform-agnostic approach that uses a thread pool which attaches/detaches should be fine.

The elephant in the room here though is going to be third-party libraries that use libdispatch instead of Swift concurrency. (I've already had to carefully audit the dependencies in my application to ensure that nothing calls `DispatchQueue.main.async()` as the main queue is never started on Android.) As long as it is not too expensive, I think a solution that attaches on demand and supports the continued use of libdispatch is desirable.

ktoso:
Heh main.async() is a pretty funny wrinkle. Having that said, if we had proper Swift concurrency adopting libraries they'd do `MainActor` _and_ then we're in the middle of allowing the executor for that actor -- so we'd be able to then replace it with some more meaningful thread on android, and not have to ban using it. This won't be in Swift 6.0 though, but is something we're investigating

lhoward:
CI error appears unrelated

```
<unknown>:0: error: unknown argument: '-abi-comments-in-module-interface'
<unknown>:0: error: unknown argument: '-abi-comments-in-module-interface'
```

ktoso:
Gotta update the branch, can you do that please :)

ktoso (approved):
I think this is okey for the time being, if we find better ways of doing this we can adjust of course.

ktoso:
Thanks for the contrib!

PR #174, opened 2024-11-09
Fail Macro Expansion if JavaStaticField is used in non-class

### LINKED ISSUE #65: JavaKit's @JavaStaticField could detect that it should not be used on static field
Followup to https://github.com/swiftlang/swift-java/pull/61

> For bonus points, one could have the @JavaStaticField macro detect the static and produce an error diagnosing the problem, but it's not really necessary. Folks shouldn't have to write these macro uses themselves.

jrosen081:
Closes #65 

This adds support for failing macro expansion if `JavaStaticField` is used in the incorrect context.

One thing I was thinking is we can add similar logic for the `JavaStaticMethod` to fail in that context. That will fail in compilation, but it could be nicer to fail expansion vs fail in an implementation detail (of a function not existing on the type)

### DIFF
--- a/Sources/JavaKitMacros/JavaFieldMacro.swift
+++ b/Sources/JavaKitMacros/JavaFieldMacro.swift
@@ -33,6 +33,11 @@ extension JavaFieldMacro: AccessorMacro {
       return []
     }
 
+    let isStatic = node.attributeName.trimmedDescription == "JavaStaticMethod"
+    guard !isStatic || isInStaticContext(context: context) else {
+      throw MacroExpansionErrorMessage("Cannot use @JavaStaticMethod outside of a JavaClass instance")
+    }
+
     // Dig out the Java field name, if provided. Otherwise, use the name as written.
     let fieldName =
       if case .argumentList(let arguments) = node.arguments,
@@ -78,5 +83,19 @@ extension JavaFieldMacro: AccessorMacro {
     }
 
     return accessors
+  }
+
+  private static func isInStaticContext(context: some MacroExpansionContext) -> Bool {
+    for lexicalContext in context.lexicalContext {
+      if let classSyntax = lexicalContext.as(StructDeclSyntax.self) {
+        if classSyntax.name.trimmedDescription.starts(with: "JavaClass") {
+          return true
+        } else {
+          return false
+        }
+      }
+    }
+
+    return false
   }
 }
--- a/Sources/JavaKitMacros/JavaFieldMacro.swift
+++ b/Sources/JavaKitMacros/JavaFieldMacro.swift
@@ -33,9 +33,9 @@ extension JavaFieldMacro: AccessorMacro {
       return []
     }
 
-    let isStatic = node.attributeName.trimmedDescription == "JavaStaticMethod"
+    let isStatic = node.attributeName.trimmedDescription == "JavaStaticField"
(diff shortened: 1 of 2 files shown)


Sources/JavaKitMacros/JavaFieldMacro.swift
```diff
@@ -79,4 +84,18 @@ extension JavaFieldMacro: AccessorMacro {
...
   }
+
+  private static func isInStaticContext(context: some MacroExpansionContext) -> Bool {
+    for lexicalContext in context.lexicalContext {
+      if let classSyntax = lexicalContext.as(ClassDeclSyntax.self) {
+        if classSyntax.name.trimmedDescription.starts(with: "JavaClass<") {
+          return true
+        } else {
+          return false
+        }
```
ktoso: that's just a `return classSyntax.name.trimmedDescription.starts(with: "JavaClass<")` right?

Sources/JavaKitMacros/JavaFieldMacro.swift
```diff
@@ -79,4 +84,18 @@ extension JavaFieldMacro: AccessorMacro {
 
     return accessors
   }
+
+  private static func isInStaticContext(context: some MacroExpansionContext) -> Bool {
```
ktoso:
```suggestion
  private static func isInJavaClassContext(context: some MacroExpansionContext) -> Bool {
```

maybe? Not to be confused in being in some actual `static` etc.

ktoso (commented):
looks good, minor comment

ktoso (approved):

ktoso:
Tests failed, something isn't quite right:

```

/__w/swift-java/swift-java/Sources/JavaKit/generated/JavaBoolean.swift:31:3: error: Cannot use @JavaStaticField outside of a JavaClass instance (from macro 'JavaStaticField')
29 | }
30 | extension JavaClass<JavaBoolean> {
31 |   @JavaStaticField(isFinal: true)
   |   `- error: Cannot use @JavaStaticField outside of a JavaClass instance (from macro 'JavaStaticField')
32 |   public var TRUE: JavaBoolean!
33 | 
```

jrosen081:
> Tests failed, something isn't quite right:
> 
> 
> 
> ```
> 
> 
> 
> /__w/swift-java/swift-java/Sources/JavaKit/generated/JavaBoolean.swift:31:3: error: Cannot use @JavaStaticField outside of a JavaClass instance (from macro 'JavaStaticField')
> 
> 29 | }
> 
> 30 | extension JavaClass<JavaBoolean> {
> 
> 31 |   @JavaStaticField(isFinal: true)
> 
>    |   `- error: Cannot use @JavaStaticField outside of a JavaClass instance (from macro 'JavaStaticField')
> 
> 32 |   public var TRUE: JavaBoolean!
> 
> 33 | 
> 
> ```

Oh my. I'm checking for classdecl, but I think I need to check for extensiondecl. I'll fix this up soon

Sources/JavaKitMacros/JavaFieldMacro.swift
jrosen081: updated to use java class in this. I can see how static could be confusing since it meant static in the `java` form not the `swift`

ktoso (approved):
Nice, thank you very much!

Issue #177, opened 2024-11-14
Use "swiftc -print-target-info" to get the java library paths where to find swiftCore.so/dylib


ktoso:
Since with Swiftly installations the paths depend on selected version so we must use this specific command to find the right paths to use if swiftly is used:

The output is:

```
{
  "compilerVersion": "Swift version 6.0.2 (swift-6.0.2-RELEASE)",
  "target": {
    "triple": "x86_64-unknown-linux-gnu",
    "unversionedTriple": "x86_64-unknown-linux-gnu",
    "moduleTriple": "x86_64-unknown-linux-gnu",
    "compatibilityLibraries": [ ],
    "librariesRequireRPath": false
  },
  "paths": {
    "runtimeLibraryPaths": [
      "/home/ktoso/.local/share/swiftly/toolchains/6.0.2/usr/lib/swift/linux"
    ],
    "runtimeLibraryImportPaths": [
      "/home/ktoso/.local/share/swiftly/toolchains/6.0.2/usr/lib/swift/linux",
      "/home/ktoso/.local/share/swiftly/toolchains/6.0.2/usr/lib/swift/linux/x86_64"
    ],
    "runtimeResourcePath": "/home/ktoso/.local/share/swiftly/toolchains/6.0.2/usr/lib/swift"
  }
}
```

ktoso:
I'll do this next

PR #178, opened 2024-11-14
use -print-target-info to get swift library paths reliably

ktoso:
resolves https://github.com/swiftlang/swift-java/issues/177

we have this code duplicated in some places due to project setup reasons, i'll fix those too


DougGregor (approved):

ktoso:
I did not manage to kill off the duplication but at least the logic is the better one now

PR #179, opened 2024-11-15
introduce some initial docc files

ktoso:
Seeing if we can start using docc a bit


ktoso:
TODO: consider the preview of https://forums.swift.org/t/a-preview-of-doccs-support-for-combined-documentation/74373

PR #182, opened 2024-11-19
Add support for failing build when multiple types are mapped to the same type

### LINKED ISSUE #176: Java2Swift: Diagnose when two Java classes are mapped to the same Swift type
It's possible to have a Java2Swift configuration file where multiple Java classes (say, from different packages) mapped to the same Swift type name in the same module. This will cause hard-to-diagnose build failures about there being multiple producers for the same object file.

We should detect this case in Java2Swift itself and produce a reasonable diagnostic with a suggestion of how to fix the problem.

jrosen081:
Closes #176 

Adds validation to the configurations created to make sure that two types are not mapped to the same type in the same module. The output looks as following:
```
Error: The following Java classes were mapped to the same Swift type:
  Swift Type: Testing.ProcessBuilder, Java Types: java.lang.Math, java.lang.ProcessBuilder
```

The configuration used was:
```json
{
  "classes": {
    "java.lang.ProcessBuilder": "ProcessBuilder",
    "java.lang.Math": "ProcessBuilder"
  }
}
```

### DIFF
--- a/Sources/Java2SwiftLib/JavaTranslator+Validation.swift
+++ b/Sources/Java2SwiftLib/JavaTranslator+Validation.swift
@@ -1,0 +1,72 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+package extension JavaTranslator {
+  struct SwiftTypeName: Hashable {
+    let swiftType: String
+    let swiftModule: String?
+
+    package init(swiftType: String, swiftModule: String?) {
+      self.swiftType = swiftType
+      self.swiftModule = swiftModule
+    }
+  }
+
+  struct SwiftToJavaMapping: Equatable {
+    let swiftType: SwiftTypeName
+    let javaTypes: [String]
+
+    package init(swiftType: SwiftTypeName, javaTypes: [String]) {
+      self.swiftType = swiftType
+      self.javaTypes = javaTypes
+    }
+  }
+
+  enum ValidationError: Error, CustomStringConvertible {
+    case multipleClassesMappedToSameName(swiftToJavaMapping: [SwiftToJavaMapping])
+
+    package var description: String {
+      switch self {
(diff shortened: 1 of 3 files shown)


Sources/Java2SwiftLib/JavaTranslator+Validation.swift
```diff
@@ -0,0 +1,72 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
ktoso: do we actually use any Foundation in here? Avoid import if not necessary

Sources/Java2SwiftLib/JavaTranslator+Validation.swift
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  enum ValidationError: Error, CustomStringConvertible {
+    case multipleClassesMappedToSameName(swiftToJavaMapping: [SwiftToJavaMapping])
+
+    package var description: String {
+      switch self {
+      case .multipleClassesMappedToSameName(let swiftToJavaMapping):
+              """
+              The following Java classes were mapped to the same Swift type:
```
ktoso:
```suggestion
              The following Java classes were mapped to the same Swift type name:
```

include "name" maybe? It made it sound like we mapped to an existing type but the problem is we didn't form a type for it at all since the conflict

Sources/Java2SwiftLib/JavaTranslator+Validation.swift
```diff
@@ -0,0 +1,72 @@
...
+                \(swiftToJavaMapping.map { mapping in
+                  "Swift Type: \(mapping.swiftType.swiftModule ?? "").\(mapping.swiftType.swiftType), Java Types: \(mapping.javaTypes.sorted().joined(separator: ", "))"
```
ktoso:
```suggestion
                  "Swift Type: '\(mapping.swiftType.swiftModule ?? "")'.\(mapping.swiftType.swiftType), Java Types: \(mapping.javaTypes.sorted().map({ "'\($0)'" })joined(separator: ", "))"
```

I'd split out the creating of the names into separate lines, this is a bit busy inside the string interpolation

ktoso (approved):
Minor nit but looks good :-)

Sources/Java2SwiftLib/JavaTranslator+Validation.swift
jrosen081: updated and made a helper function to split it out even more

DougGregor:
Thank you!

PR #188, opened 2024-11-25
Add Support for Threadsafe and NotThreadSafe annotations

### LINKED ISSUE #129: Consider @ThreadSafe when importing Java types
While "dormant" some project do use the JSR-305 annotations, and one of them is https://www.javadoc.io/doc/com.google.code.findbugs/jsr305/latest/javax/annotation/concurrent/ThreadSafe.html

This is something we could use to annotate such imported Java types as `Sendable`.

Others are:

Name | Description
-- | --
GuardedBy | The field or method to which this annotation is applied can only be accessed when holding a particular lock, which may be a built-in (synchronization) lock, or may be an explicit Lock.
Immutable | The class to which this annotation is applied is immutable.
NotThreadSafe | The class to which this annotation is applied is not thread-safe.
ThreadSafe | The class to which this annotation is applied is thread-safe.

jrosen081:
Closes #129 

This adds support for threadsafe, immutable, and notthreadsafe annotations. I was able to test this locally making jars, which I can attach if needed, but couldn't do any automated testing due to the fact that these types are not built in to Java, so I couldn't use the normal translator tests. I'm open to hearing if there are any good ways of doing automated testing for this (we could possibly build a jar in CI with the given dependencies and use that, but it's probably not great to be doing that during unit tests, especially since I think we will need to download dependencies if we want to use the real annotations in the ticket)

### DIFF
--- a/Samples/JavaKitSampleApp/Sources/JavaKitExample/com/example/swift/ThreadSafeHelperClass.java
+++ b/Samples/JavaKitSampleApp/Sources/JavaKitExample/com/example/swift/ThreadSafeHelperClass.java
@@ -1,0 +1,6 @@
+package com.example.swift;
+
+@ThreadSafe
+public class ThreadSafeHelperClass {
+    public ThreadSafeHelperClass() { }
+}
--- a/Samples/JavaKitSampleApp/Sources/JavaKitExample/com/example/swift/ThreadSafeHelperClass.java
+++ b/Samples/JavaKitSampleApp/Sources/JavaKitExample/com/example/swift/ThreadSafeHelperClass.java
@@ -1,3 +1,17 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
 package com.example.swift;
 
 @ThreadSafe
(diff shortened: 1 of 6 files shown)


ktoso:
In a Java test modules you could literarily define an annotation with those names - it doesn't have to be the "real jar". Do you want to give that a short or should I try to streamline it a bit more? Idk the exact steps without looking into it, let me know if you'd need a hand here

jrosen081:
> In a Java test modules you could literarily define an annotation with those names - it doesn't have to be the "real jar". Do you want to give that a short or should I try to streamline it a bit more? Idk the exact steps without looking into it, let me know if you'd need a hand here.
> 

I'll look into it later this week and let you know if I can figure out the automated test way.

> 
> 
> Alternatively you could do this on one of the sample apps -- we CI verify them all, so declare the annotation there and see that some type indeed got the Sendable conformance by assigning it to some `var: x: Sendable` etc?

Oh didn't know that this was done in CI. If I can't get the first part right, I can do this part

ktoso:
We actually rely on the samples a lot for correctness so yeah, that'd definitely be a viable way to verify things as well :)

Samples/JavaKitSampleApp/Sources/JavaKitExample/com/example/swift/ThreadSafeHelperClass.java
```diff
@@ -0,0 +1,6 @@
+package com.example.swift;
```
ktoso, line 15: missing license header?

ktoso (approved):
Looks good, thanks for including the snippet

jrosen081:
> We actually rely on the samples a lot for correctness so yeah, that'd definitely be a viable way to verify things as well :)

I updated one of the Samples to check for ThreadSafe. This still has the hole where there isn't a way to test something is _not_ Sendable (which NotThreadSafe does). I can look later in the week for other testing of not sendable depending on what you think

Samples/JavaKitSampleApp/Sources/JavaKitExample/com/example/swift/ThreadSafeHelperClass.java
jrosen081: oops updated

ktoso:
That's okey, thanks :)

PR #189, opened 2024-11-27
Adds support for optional mapping

### LINKED ISSUE #92: Java2Swift: Consider translating java.util.Optional into Swift.Optional
[`java.util.Optional`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html) is very similar in spirit to Swift optionals. We should determine whether there is a good way to automatically translate `java.util.Optional<T>` to `T?` in Swift, correctly.

jrosen081:
Closes #92 

Adds support for methods and variables with optional. I went for an approach of making a wrapper func/var with the `Optional` suffix, but am open to other ideas. Also fixed up the Makefile so i could use it


Sources/Java2SwiftLib/StringExtras.swift
```diff
@@ -39,4 +39,18 @@ extension String {
...
     return replacingOccurrences(of: String(character), with: String(replacement))
   }
+
+  public func optionalWrappedType() -> String? {
+    if self == "JavaOptionalInt!" {
+      return "Int32?"
+    } else if self == "JavaOptionalDouble!" {
+      return "Double?"
+    } else if self == "JavaOptionalLong!" {
+      return "Int64?"
```
ktoso, line 47: nitpick but maybe order as Int32, Int64 and then Double, grouping them closer together a bit

Sources/Java2SwiftLib/StringExtras.swift
```diff
@@ -39,4 +39,18 @@ extension String {
...
+    } else if hasPrefix("JavaOptional<") {
+      return "\(String(self[index(startIndex, offsetBy: 13)..<lastIndex(of: ">")!]))?"
```
ktoso: the "13" is a bit magic number, can we .count on the string we're skipping instead please?

Sources/JavaKit/Optional+JavaOptional.swift
```diff
@@ -0,0 +1,81 @@
...
+  init(javaOptional: JavaOptionalInt?) {
+    if let javaOptional {
+      self = javaOptional.isPresent() ? javaOptional.getAsInt() : nil
+    } else {
+      self = nil
+    }
+  }
+}
+
+extension Optional where Wrapped == Int64 {
```
ktoso: Should we also produce those for `UInt` versions? Java doesn't have unsigned but may be convenient for Swift hm

Sources/JavaKit/generated/JavaOptionalLong.swift
```diff
@@ -0,0 +1,44 @@
...
+  public func ofOptional(_ arg0: Int64) -> Int64? {
+    Optional(javaOptional: of(arg0))
+  }
+
+  @JavaStaticMethod
+  public func empty() -> JavaOptionalLong!
+
+  public func emptyOptional() -> Int64? {
+    Optional(javaOptional: empty())
+  }
```
ktoso, line 43: Not sure I understand where the ofOptional and emptyOptional came from? I know this is generated but couldn't quite wrap my head around what the intent of these is?

Sources/JavaKit/generated/JavaOptional.swift
```diff
@@ -0,0 +1,51 @@
...
+  @JavaMethod
+  open func isPresent() -> Bool
+
+  @JavaMethod
+  open func orElse(_ arg0: JavaObject?) -> JavaObject!
+
+  @JavaMethod
+  open func orElseThrow() -> JavaObject!
+}
+extension JavaClass {
```
ktoso, line 30: Does this look right?

Sources/Java2SwiftLib/JavaClassTranslator.swift
```diff
@@ -582,10 +612,33 @@ extension JavaClassTranslator {
...
+
+    if let optionalType = swiftTypeName.optionalWrappedType() {
+      let setter = if !javaField.isFinal {
+      """
+      
+        set {
+          \(swiftFieldName) = newValue?.toJavaOptional()
+        }
+      """
+      } else { "" }
```
ktoso: The formatting here is somewhat confusing with the else not being indented, what does swift-format suggest here (we don't require it in CI since we have an issue with it hanging on CI... but we need to fix that).

Sources/Java2SwiftLib/JavaClassTranslator.swift
```diff
@@ -582,10 +612,33 @@ extension JavaClassTranslator {
...
+      \(fieldAttribute)(isFinal: \(raw: javaField.isFinal))
+      public var \(raw: swiftFieldName): \(raw: typeName)
+
+      
+      public var \(raw: swiftFieldName)Optional: \(raw: optionalType) {
+        get {
+          Optional(javaOptional: \(raw: swiftFieldName))
+        }\(raw: setter)
+      }
+      """
```
ktoso, line 637: I think going the opposite way maybe is more typical? It is "normal" to expose the  Swift optional and maybe have the java optional type property as `thingJava` or some `thing$raw` or `thing$java`? Just ideas 🤔

Samples/JavaKitSampleApp/Sources/JavaKitExample/JavaKitExample.swift
```diff
@@ -77,7 +77,12 @@ extension HelloSwift: HelloSwiftNativeMethods {
     }
 
     // Make sure that the thread safe class is sendable
-    let threadSafe: Sendable = ThreadSafeHelperClass(environment: javaEnvironment)
+    let helper = ThreadSafeHelperClass(environment: javaEnvironment)
+    let threadSafe: Sendable = helper
+
+    let text: String? = helper.textOptional
```
ktoso: since we start having a lot of things here maybe put this into `void testOptionals()` and have a bunch of checks there, including checking the numeric ones?

ktoso (commented):
Some minor comments, looks like a good step

ktoso:
A bunch of compile errors discovered by the Sample, please check

Sources/JavaKit/generated/JavaOptionalLong.swift
jrosen081: This is a fun one:
Any method that returns a Java optional (or friends) now has a second method that returns a swift optional and does wrapping. A fun consequence of this is that all optionals have static methods to create themselves ... which then have a helper generated to make a swift optional. It's a fun bit of recursion-ish stuff

Sources/Java2SwiftLib/JavaClassTranslator.swift
jrosen081: I went the opposite way first but then ran into a roadblock since the JavaMethod macro doesn't take in a name to use as the Java method name (which can be done, just felt weird as part of this pr). This means that it _needs_ to have the exact Java name and then it felt weird to not have parity between fields and methods with optional mapping

Sources/JavaKit/generated/JavaOptional.swift
jrosen081: Looks to be how we import static methods/functions for classes that have generic parameters. Seems to be the same way we do it for `HashMap` since the logic for the generic gets put into the function itself. Unless I misunderstood what you meant?

ktoso: Hmmm, I see (note: I've not spend much time with JavaKit but more focused on the opposite direction, so thanks for helping review this 😄)... 

Yeah that makes sense I guess, I wonder if this will become a problem in the long term with names clashing -- lots of types like to have a `of(_:)` or some form of `create(_:)` but maybe it'll work out okey, could be a lot of overloads eventually hm

Sources/Java2SwiftLib/JavaClassTranslator.swift
ktoso: We can go with this way if you get it to build tbh, thanks for the background info

Sources/JavaKit/Optional+JavaOptional.swift
jrosen081: Would these map to JavaOptional<UInt32> or did you have something else in mind? There is no built-in type, but it could be nice (we can map the boxed types if we don't already)

jrosen081: I don't see any unsigned long wrapper types (could have missed them in the docs), so I worry about mapping these as we would be changing the signedness of them

jrosen081:
> A bunch of compile errors discovered by the Sample, please check

Updated the code to get it running correctly with more samples!

Sources/JavaKit/Optional+JavaOptional.swift
ktoso: Ok let's let those be :)

ktoso (approved):

ktoso:
Very nice, thanks again :-)

Issue #169, opened 2024-11-09
[Swift-6] Get rid of type shadowing warnings in generated sources


ktoso:
We have a lot of noise thanks to these: 

```
 98 | extension TreeMap {
 99 |   @JavaClass("java.util.AbstractMap$SimpleImmutableEntry")
100 |   open class SimpleImmutableEntry<K: AnyJavaObject, V: AnyJavaObject>: JavaObject {
    |                                                     `- warning: generic parameter 'V' shadows generic parameter from outer scope with the same name; this is an error in the Swift 6 language mode
```

it'd be nice to avoid them, e.g. make those V1 or something

AbdAlRahmanGad:
Hi @ktoso, 
Could you point me to which file I should implement this change inside?

ktoso:
Most of this is in JavaTranslator however some of it may be spread around that Java2SwiftLib module.

We should somehow keep track that a `V` is already used in an outer scope and then use a V1 for example or different name

PR #200, opened 2024-12-11
get rid of some of the warnings in Java2Swift generated sources: generics

AbdAlRahmanGad:
Tries to solve #169

I got rid of some of the warning but I'm now stuck.

The following classes were translated appropriately
```
JavaOptional.swift
HashMap.swift
HashSet.swift
JavaSet.swift
List.swift
```

These classes were generated with a different order of methods and variables, I don't know why that happened.
```
JavaClass.swift
JavaString.swift
BitSet.swift
```
This class generates a bad translation which is not a valid class, I don't know why that happedned.
```
URLClassLoader.swift
```

### DIFF
--- a/Sources/Java2SwiftLib/JavaClassTranslator.swift
+++ b/Sources/Java2SwiftLib/JavaClassTranslator.swift
@@ -104,10 +104,23 @@ struct JavaClassTranslator {
     }
 
     let genericParameters = javaTypeParameters.map { param in
-      "\(param.getName()): AnyJavaObject"
+      let name = generateGenericParameterName(param.getName())
+      translator.usedGenericParameterNames.insert(param.getName())
+      return "\(name): AnyJavaObject"
     }
 
     return "<\(genericParameters.joined(separator: ", "))>"
+  }
+
+  /// If the name already exists, append a 1 to it.
+  /// For example, if name = "T" and it already exists, the result will be "T1".
+  private func generateGenericParameterName(_ baseName: String) -> String {
+    var name = baseName
+    let index = 1
+    if translator.usedGenericParameterNames.contains(name) {
+      name = "\(baseName)\(index)"
+    }
+    return name
   }
 
   /// Prepare translation for the given Java class (or interface).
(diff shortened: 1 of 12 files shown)


Sources/Java2SwiftLib/JavaClassTranslator.swift
```diff
@@ -104,12 +104,25 @@ struct JavaClassTranslator {
...
 
+  /// If the name already exists, append a 1 to it.
+  /// For example, if name = "T" and it already exists, the result will be "T1".
+  private func generateGenericParameterName(_ baseName: String) -> String {
+    var name = baseName
+    let index = 1
+    if translator.usedGenericParameterNames.contains(name) {
+      name = "\(baseName)\(index)"
+    }
+    return name
```
ktoso, line 123: This would probably have to be more contextual "in this context (of a function or type) there already was a T used so make it a T1". Especially with types nested in other types where there's a T outside and then another T etc...

ktoso (changes requested):
Yeah that's roughly the idea, other than pursuing it more and trying what will work out in the end I'm not sure I have much hints to provide here. You'll have to work out how to avoid the conflicts and just think through edge cases you uncover

AbdAlRahmanGad:
I tried running `java-generate` on a fresh copy of the main branch, but the issue still persists.

Even without my changes, these files result in a different order of methods and variables:  
```
JavaClass.swift
JavaString.swift
BitSet.swift
```

And this file generates a non-valid class:  
```
URLClassLoader.swift
```

Should I file an issue on this?

AbdAlRahmanGad:
> Huh I see, then yeah please file an issue and ofc help on resolving it would also be welcome but otherwise we'll find some time :)

I'd really like to help! But for now and until mid-January, I will be busy with college projects and final exams.

As of the state of this PR, should I resume working on it when the issue is resolved?

ktoso:
This has been inactive for a while so I'll close, but we'd welcome any new PRs :)

PR #210, opened 2025-01-23
Initial implementation of a new lowering from a Swift func to `@_cdecl` func

DougGregor:
Start implementing a more complete and formal lowering from an arbitrary Swift function to a `@_cdecl`-compatible thunk that calls that function. This is set up in stages more akin to what you'd see in a compiler:

1. Resolve the syntax trees for the function into a more semantic representation, for example resolving type names to nominal type declarations. This includes a simplistic implementation of a symbol table so we can resolve arbitrary type names.
2. Lower the semantic representation of each function parameter (including self). How we do this varies based on type:
  * Value types (struct / enum) are passed indirectly via Unsafe(Mutable)RawPointer, using a mutable pointer when the corresponding parameter is inout.
  * Class / actor types are passed directly via UnsafeRawPointer.
  * Swift types that map to primitive types (like Swift.Int32) in passed directly, no translation required.
  * Tuple types are recursively "exploded" into multiple parameters.
  * Unsafe*BufferPointer types are "exploded" into a pointer and count.
  * Typed unsafe pointers types are passed via their raw versions.
3. Lower returns similarly to parameters, using indirect mutable parameters for the return when we can't return directly.
4. Render the lowered declaration into a FunctionDeclSyntax node, which can be modified by the client.

At present, we aren't rendering the bodies of these thunks, which need to effectively undo the transformations described in (3) and (4). For example, reconstituting a tuple from disparate arguments, appropriately re-typing and loading from indirectly-passed value types, and so on. The description of the lowered signature is intended to provide sufficient information for doing so, but will likely require tweaking.

At present, this new code is not integrated in the main code path for jextract-swift. Once it hits feature parity, we'll enable it.

### DIFF
--- a/Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
+++ b/Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
@@ -13,11 +13,12 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import JavaTypes
 
 /// Represents a value of a `java.lang.foreign.Self` that we want to render in generated Java code.
 ///
 /// This type may gain further methods for adjusting target layout, byte order, names etc.
-public struct ForeignValueLayout: CustomStringConvertible {
+public struct ForeignValueLayout: CustomStringConvertible, Equatable {
   var inlineComment: String?
   var value: String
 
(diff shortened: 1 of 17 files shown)


DougGregor:
I'm going to merge this and continue iterating.

PR #211, opened 2025-01-27
jextract: Indirect returns and more value type support

ktoso:
This includes:

- handling indirect return in initializers of value types
- destroy() done in a SwiftArena on value types 
- some other fixes in calling our thunks -- we no longer need the Self.Type in initializers since we started doing thunks for inits.
- more safety against use-after free with the arena's marking java wrappers as destroyed.


Sources/JExtractSwift/ImportedDecls.swift
```diff
@@ -73,6 +74,34 @@ public enum NominalTypeKind {
   case `class`
   case `enum`
   case `struct`
+  case `void`  // TODO: NOT NOMINAL, BUT...
```
ktoso, line 78: I'll work on removing this `NominalTypeKind` so we replace it with `SwiftNominalTypeDeclaration.Kind` but we'll have to do a bit of a dance to pull it off -- perhaps make it optional somehow or enum over "nominal or not nominal" etc

Sources/JExtractSwift/ImportedDecls.swift
```diff
@@ -44,18 +44,19 @@ public struct ImportedNominalType: ImportedDecl {
     TranslatedType(
       cCompatibleConvention: .direct,
       originalSwiftType: "\(raw: swiftTypeName)",
+      originalSwiftTypeKind: self.kind,
```
ktoso, line 47: TODO: we'll want to remove this and instead pass around a SwiftNominalTypeDeclaration instead, and base the name and kind off from that.

Sources/JExtractSwift/TranslatedType.swift
```diff
@@ -274,6 +294,15 @@ enum CCompatibleJavaMemoryLayout: Hashable {
   case cFunction
 }
 
+enum SwiftTypeKind {
```
ktoso, line 297: TODO: Working on removing this and converge with the other `SwiftNominalTypeDeclaration.Kind` in a follow up PR

PR #212, opened 2025-01-30
Implement function bodies for lowered @_cdecl thunks

DougGregor:
Improve the handling of lowering Swift declarations down to C thunks to more clearly model the mapping between the C parameters and the Swift parameters, and start generating code within the body of these C thunks. Provide tests and fixes for lowering of methods of structs and classes, including mutating methods, as well as inout parameters, direct and indirect returns, and so on.

### DIFF
--- a/Sources/JExtractSwift/Swift2JavaTranslator+FunctionLowering.swift
+++ b/Sources/JExtractSwift/Swift2JavaTranslator+FunctionLowering.swift
@@ -127,8 +127,28 @@ extension Swift2JavaTranslator {
     parameterName: String
   ) throws -> LoweredParameters {
     switch type {
-    case .function, .metatype, .optional:
+    case .function, .optional:
       throw LoweringError.unhandledType(type)
+
+    case .metatype(let instanceType):
+      return LoweredParameters(
+        cdeclToOriginal: .unsafeCastPointer(
+          .passDirectly(parameterName),
+          swiftType: instanceType
+        ),
+        cdeclParameters: [
+          SwiftParameter(
+            convention: .byValue,
+            parameterName: parameterName,
+            type: .nominal(
+              SwiftNominalType(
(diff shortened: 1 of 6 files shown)


Sources/JExtractSwift/Swift2JavaTranslator+FunctionLowering.swift
```diff
@@ -145,8 +168,12 @@ extension Swift2JavaTranslator {
       let mutable = (convention == .inout)
       let loweringStep: LoweringStep
       switch nominal.nominalTypeDecl.kind {
-      case .actor, .class: loweringStep = .passDirectly(parameterName)
-      case .enum, .struct, .protocol: loweringStep = .passIndirectly(parameterName)
+      case .actor, .class:
+        loweringStep = 
+          .unsafeCastPointer(.passDirectly(parameterName), swiftType: type)
```
ktoso, line 173: Oh nice, I like this idea of constructing what we need to do like that

Sources/JExtractSwift/Swift2JavaTranslator+FunctionLowering.swift
```diff
@@ -375,3 +484,74 @@ public struct LoweredFunctionSignature: Equatable {
   var parameters: [LoweredParameters]
   var result: LoweredParameters
 }
+
+extension LoweredFunctionSignature {
+  /// Produce the `@_cdecl` thunk for this lowered function signature that will
+  /// call into the original function.
+  @_spi(Testing)
+  public func cdeclThunk(cName: String, inputFunction: FunctionDeclSyntax) -> FunctionDeclSyntax {
```
ktoso, line 492: Very nice, thanks! 👍

ktoso (approved):
Really cool work here, looking good

PR #219, opened 2025-04-10
Refactor to converge many command line tools into a single "swift-java"

ktoso:
I think it's a good time to converge all the commands we have. Developers don't need to concern themselfes about tool names -- just what they want to achieve. It should not matter where a fetch dependencies or generates swift or java is in our heads, for developers it's all just "swift java interop" so we can offer a single tool `swift-java` that has various subcommands to do the different tasks.

This should also help reduce duplication a bit.

Also using this PR to check if the CI is healthy, we've not touched this repo since a while.

And we'll do subcommands instead.


ktoso:
Will redo this given the refactorings recently

Issue #209, opened 2025-01-23
jextract must handle "sibling" types while extracting


ktoso:
For example file 1:

```
struct Hello {}
```

file 2:

```
struct HelloAgain {
  init(hello: Hello) {}
}
```

if we visit them in the order Hello, then HelloAgain things work fine.

If we happen to visit them `HelloAgain` first, then we'll reject importing the HelloAgain type because:

```
[info][JExtractSwift/Swift2JavaVisitor.swift:219](visit(_:)) Unable to import initializer due to unknown(IdentifierTypeSyntax
╰─name: identifier("Hello"))
```

PR #222, opened 2025-04-15
[jextract] Prepare all inputs before analyze()

rintaro:
Register all the files to `Swift2JavaTranslator` (and `NominalTypeResolution`) and `analye()` all the files at once.

`Swift2JavaVisitor.visit(_: ExtentionDeclSyntax)` requires all the nominal types are prepared in the `nominalResolution`. Otherwise it's just ignored.

Also previously, writing files was performed for each file iteration but actually, it is not a per-source file operation. Instead, it only writes the current state of the translator, so multiple calls were redundant.

### DIFF
--- a/Sources/JExtractSwift/Swift2Java.swift
+++ b/Sources/JExtractSwift/Swift2Java.swift
@@ -74,16 +74,23 @@ public struct SwiftToJava: ParsableCommand {
       }
     }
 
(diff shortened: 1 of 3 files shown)


DougGregor (approved):
Ah, thank you! Might this fix https://github.com/swiftlang/swift-java/issues/209?

ktoso (approved):
Nice, thank you

PR #227, opened 2025-05-05
[WIP] Allow returning new java objects from JNI downcall

ktoso:
This isn't really a solution but just showcases the issue.

We cannot just return a newly created object without keeping the ref alive as we'll get a null on the java side otherwise.

The solution will be managing refs and checking if we escaped them or not proably...


Sources/JavaKit/JavaObjectHolder.swift
```diff
@@ -38,6 +39,6 @@ public class JavaObjectHolder {
   }
 
   deinit {
-    self.forget()
+    // self.forget()
```
ktoso, line 42: fixme, not real solution ofc.

colemancda:
I believe this is related to the behavior I was seeing https://github.com/swiftlang/swift-java/issues/272

ktoso:
Going to close this as we'll revisit properly soon

PR #234, opened 2025-05-27
Add JavaIO and CSV example

ktoso:
Adds some types from java.io and an example that consumes commons-csv


Samples/JavaDependencySampleApp/Sources/JavaDependencySample/main.swift
```diff
@@ -42,4 +43,13 @@ let ext = try! FilenameUtilsClass.getExtension(path)
 print("org.apache.commons.io.FilenameUtils.getExtension = \(ext)")
 precondition(ext == "exe")
 
+let CSCFormatClass = try JavaClass<CSVFormat>()
```
t089:
```suggestion
let CSVFormatClass = try JavaClass<CSVFormat>()
```

ktoso: Thanks :)

PR #236, opened 2025-05-29
[jextract] Rework translation

rintaro:
* Utilize `SwiftFunctionSignature` and the cdecl-lowering  facilities throughout the code base. 
  * `SwiftFunctionSignature` is created from the `DeclSyntax` in `Swift2JavaVisitor`.
  * `LoweredFunctionSignature` describing `@cdecl` thunk, is created from `SwiftFunctionSignature`.
  * `TranslatedFunctionSignature` describing generated Java API, is created from `LoweredFunctionSignature`. (Swift2JavaTranslator+JavaTranslation.swift)
  * `ImportedFunc` is now basically just a wrapper of `TranslatedFunctionSignature` with the `name`.  
* Remove `ImportedVariable`, instead variables are described as `ImportedFunc` as accessors.
* Now supports methods returning imported type values. E.g. `func factory(arg: Int) -> MyClass` such methods require `SwiftArena` passed-in.
* Built-in lowerings (e.g. `toCString(String)` for `String` -> C string conversion) are now implemented in JavaKit.
* Stop emitting `MyClass::apiName$descriptor()` method etc. as they were not used.
* Use the `@cdecl` thunk name as the function descriptor class name, for simplicity.
* Getter and setter accessors are now completely separate API. No more `HANDLE_GET` and `HANDLE_SET` etc. Instead descriptor class is split to `$get` or `$set` suffixed name.

<details>
<summary>Actual output of MyClass.java ▼</summary>

```java
// Generated by jextract-swift
// Swift module: MySwiftLibrary

package com.example.swift;

import org.swift.swiftkit.*;
import org.swift.swiftkit.SwiftKit;
import org.swift.swiftkit.util.*;
import java.lang.foreign.*;
import java.lang.invoke.*;
import java.util.Arrays;
import java.util.stream.Collectors;
import java.util.concurrent.atomic.*;
import java.nio.charset.StandardCharsets;

public final class MySwiftClass extends SwiftValue implements SwiftHeapObject {
  static final String LIB_NAME = "MySwiftLibrary";
  static final Arena LIBRARY_ARENA = Arena.ofAuto();
  @SuppressWarnings("unused")
  private static final boolean INITIALIZED_LIBS = initializeLibs();
  static boolean initializeLibs() {
      System.loadLibrary(SwiftKit.STDLIB_DYLIB_NAME);
      System.loadLibrary("SwiftKitSwift");
      System.loadLibrary(LIB_NAME);
      return true;
  }
  
  public static final SwiftAnyType TYPE_METADATA =
      new SwiftAnyType(SwiftKit.swiftjava.getType("MySwiftLibrary", "MySwiftClass"));
  public final SwiftAnyType $swiftType() {
      return TYPE_METADATA;
  }
  
  private static final GroupLayout $LAYOUT = (GroupLayout) SwiftValueWitnessTable.layoutOfSwiftType(TYPE_METADATA.$memorySegment());
  public static final GroupLayout $LAYOUT() {
      return $LAYOUT;
  }
  public final GroupLayout $layout() {
      return $LAYOUT;
  }
  
  public MySwiftClass(MemorySegment segment, SwiftArena arena) {
    super(segment, arena);
  }
  
  // ==== --------------------------------------------------
  // MySwiftClass.init
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_init_len_cap {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* len: */SwiftValueLayout.SWIFT_INT,
      /* cap: */SwiftValueLayout.SWIFT_INT,
      /* _result: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_init_len_cap");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printInitializerDowncallConstructors(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:320
  /**
   * Create an instance of {@code MySwiftClass}.
   *
   * {@snippet lang=swift :
   * public init(len: Int, cap: Int)
   * }
   */
  public MySwiftClass(long len, long cap) {
    this(len, cap, SwiftArena.ofAuto());
  } // printInitializerDowncallConstructor(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:424
  /**
   * Create an instance of {@code MySwiftClass}.
   *
   * {@snippet lang=swift :
   * public init(len: Int, cap: Int)
   * }
   */
  public MySwiftClass(long len, long cap, SwiftArena swiftArena$) {
    super(() -> {
      var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_init_len_cap.HANDLE;
      try {
        MemorySegment _result = swiftArena$.allocate(MySwiftClass.$LAYOUT());
        if (SwiftKit.TRACE_DOWNCALLS) {
            SwiftKit.traceDowncall(len, cap, _result);
        }
        mh$.invokeExact(len, cap, _result);
        return _result;
      } catch (Throwable ex$) {
        throw new AssertionError("should not reach here", ex$);
      }
    }, swiftArena$);
  } // printInitializerDowncallConstructor(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:424
  
  // ==== --------------------------------------------------
  // getter:MySwiftClass.len
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_len$get {
    public static final FunctionDescriptor DESC = FunctionDescriptor.of(
      /* -> */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_len$get");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public var len: Int
   * }
   */
  public long getLen() {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_len$get.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(this.$memorySegment());
      }
      return (long)mh$.invokeExact(this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // setter:MySwiftClass.len
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_len$set {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* newValue: */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_len$set");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public var len: Int
   * }
   */
  public void setLen(long newValue) {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_len$set.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(newValue, this.$memorySegment());
      }
      mh$.invokeExact(newValue, this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // getter:MySwiftClass.cap
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_cap$get {
    public static final FunctionDescriptor DESC = FunctionDescriptor.of(
      /* -> */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_cap$get");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public var cap: Int
   * }
   */
  public long getCap() {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_cap$get.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(this.$memorySegment());
      }
      return (long)mh$.invokeExact(this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // setter:MySwiftClass.cap
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_cap$set {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* newValue: */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_cap$set");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public var cap: Int
   * }
   */
  public void setCap(long newValue) {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_cap$set.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(newValue, this.$memorySegment());
      }
      mh$.invokeExact(newValue, this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // getter:MySwiftClass.counter
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_counter$get {
    public static final FunctionDescriptor DESC = FunctionDescriptor.of(
      /* -> */SwiftValueLayout.SWIFT_INT32,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_counter$get");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public var counter: Int32
   * }
   */
  public int getCounter() {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_counter$get.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(this.$memorySegment());
      }
      return (int)mh$.invokeExact(this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // setter:MySwiftClass.counter
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_counter$set {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* newValue: */SwiftValueLayout.SWIFT_INT32,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_counter$set");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public var counter: Int32
   * }
   */
  public void setCounter(int newValue) {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_counter$set.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(newValue, this.$memorySegment());
      }
      mh$.invokeExact(newValue, this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // MySwiftClass.factory
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_factory_len_cap {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* len: */SwiftValueLayout.SWIFT_INT,
      /* cap: */SwiftValueLayout.SWIFT_INT,
      /* _result: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_factory_len_cap");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public static func factory(len: Int, cap: Int) -> MySwiftClass
   * }
   */
  public static MySwiftClass factory(long len, long cap) {
    return factory(len, cap, SwiftArena.ofAuto());
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public static func factory(len: Int, cap: Int) -> MySwiftClass
   * }
   */
  public static MySwiftClass factory(long len, long cap, SwiftArena swiftArena$) {
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_factory_len_cap.HANDLE;
    try {
      MemorySegment _result = swiftArena$.allocate(MySwiftClass.$LAYOUT());
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(len, cap, _result);
      }
      return new MySwiftClass(_result, swiftArena$);
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // MySwiftClass.voidMethod
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_voidMethod {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_voidMethod");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public func voidMethod()
   * }
   */
  public void voidMethod() {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_voidMethod.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(this.$memorySegment());
      }
      mh$.invokeExact(this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // MySwiftClass.takeIntMethod
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_takeIntMethod_i {
    public static final FunctionDescriptor DESC = FunctionDescriptor.ofVoid(
      /* i: */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_takeIntMethod_i");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public func takeIntMethod(i: Int)
   * }
   */
  public void takeIntMethod(long i) {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_takeIntMethod_i.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(i, this.$memorySegment());
      }
      mh$.invokeExact(i, this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // MySwiftClass.echoIntMethod
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_echoIntMethod_i {
    public static final FunctionDescriptor DESC = FunctionDescriptor.of(
      /* -> */SwiftValueLayout.SWIFT_INT,
      /* i: */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_echoIntMethod_i");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public func echoIntMethod(i: Int) -> Int
   * }
   */
  public long echoIntMethod(long i) {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_echoIntMethod_i.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(i, this.$memorySegment());
      }
      return (long)mh$.invokeExact(i, this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // MySwiftClass.makeIntMethod
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_makeIntMethod {
    public static final FunctionDescriptor DESC = FunctionDescriptor.of(
      /* -> */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_makeIntMethod");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public func makeIntMethod() -> Int
   * }
   */
  public long makeIntMethod() {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_makeIntMethod.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(this.$memorySegment());
      }
      return (long)mh$.invokeExact(this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  
  // ==== --------------------------------------------------
  // MySwiftClass.makeRandomIntMethod
  
  private static class swiftjava_MySwiftLibrary_MySwiftClass_makeRandomIntMethod {
    public static final FunctionDescriptor DESC = FunctionDescriptor.of(
      /* -> */SwiftValueLayout.SWIFT_INT,
      /* self: */SwiftValueLayout.SWIFT_POINTER
    );
    public static final MemorySegment ADDR =
      MySwiftLibrary.findOrThrow("swiftjava_MySwiftLibrary_MySwiftClass_makeRandomIntMethod");
    public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(ADDR, DESC);
  } // printFunctionDowncallMethods(_:_:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:338
  /**
   * Downcall to Swift:
   * {@snippet lang=swift :
   * public func makeRandomIntMethod() -> Int
   * }
   */
  public long makeRandomIntMethod() {
    $ensureAlive();
    var mh$ = swiftjava_MySwiftLibrary_MySwiftClass_makeRandomIntMethod.HANDLE;
    try {
      if (SwiftKit.TRACE_DOWNCALLS) {
          SwiftKit.traceDowncall(this.$memorySegment());
      }
      return (long)mh$.invokeExact(this.$memorySegment());
    } catch (Throwable ex$) {
      throw new AssertionError("should not reach here", ex$);
    }
  } // printFuncDowncallMethod(_:_:isAutoArenaWrapper:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:481
  @Override
  public String toString() {
      return getClass().getSimpleName()
          + "("
          + SwiftKit.nameOfSwiftType($swiftType().$memorySegment(), true)
          + ")@"
          + $memorySegment();
  }
} // printNominal(_:_:body:) @ JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift:194
```

</details>


SwiftKit/src/main/java/org/swift/swiftkit/SwiftKit.java
```diff
@@ -206,7 +207,7 @@ public static long retainCount(MemorySegment object) {
     }
 
     public static long retainCount(SwiftHeapObject object) {
-        return retainCount(object.$memorySegment());
+        return retainCount(object.$instanceDataAddress());
```
ktoso: maybe just `$instance`? Data is a bit misleading for a class I suppose

SwiftKit/src/main/java/org/swift/swiftkit/SwiftInstanceCleanup.java
```diff
@@ -22,51 +22,6 @@
...
-        this.selfType = selfType;
-    }
-
-    @Override
-    public void run() throws UnexpectedRetainCountException {
-        // Verify we're only destroying an object that's indeed not retained by anyone else:
-        long retainedCount = SwiftKit.retainCount(selfPointer);
-        if (retainedCount > 1) {
-            throw new UnexpectedRetainCountException(selfPointer, retainedCount, 1);
-        }
```
rintaro: Since `SwiftHeapObjectCleanup` is NOT used anymore, `arena_releaseClassOnClose_class_leaked` test is now failing, but this check was not okay because somewhere in the Swift world may still owns the instance.

ktoso: Okey, let's go with that.

A bit unsure about it since the intent was to eventually give developers control over if we want to allow an object to escape a confined arena scope or not but we can always bring back that mode if we'd like to. Let's go with what you're proposing here 👍

SwiftKit/src/main/java/org/swift/swiftkit/SwiftValueWitnessTable.java
```diff
@@ -220,12 +220,8 @@ public static void destroy(SwiftAnyType type, MemorySegment object) {
 
         var mh = destroy.handle(type);
 
-        try (var arena = Arena.ofConfined()) {
-            // we need to make a pointer to the self pointer when calling witness table functions:
-            MemorySegment indirect = arena.allocate(SwiftValueLayout.SWIFT_POINTER); // TODO: remove this and just have classes have this always anyway
-            MemorySegmentUtils.setSwiftPointerAddress(indirect, object);
-
-            mh.invokeExact(indirect, wtable);
```
ktoso: I think you mentioned this last week but I remember this being done on purpose since there's an indirection here... @DougGregor can you remember this bit?

This was that the VWT destroy wanted an indirect pointer AFAIR, wasn't it?

rintaro: IIUC, this was correct for HeapObject being the address pointed the instance. But for value types (now including reference types), the address points the value (which is,  for reference types, a storage for the pointer to the instance). What we should destroy is that value I think.

Sources/JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift
```diff
@@ -0,0 +1,717 @@
+//===----------------------------------------------------------------------===//
```
ktoso: Re filename: I don't think we should be calling the Java side "thunks", if we're looking for names we've been calling them accessors I guess. I'd like to keep that distinction

SwiftKit/src/main/java/org/swift/swiftkit/SwiftValueWitnessTable.java
ktoso: Reading through this more I think the removal of `SwiftHeapObject` use is pushing it too far. A swift class is _not_ a "swift value". Swift value types are `SwiftValue`, and we want to know about that in the Java side so that we know that we may need to `copy()` a value potentially etc.

Even if the cleanups we end up doing the same way, we can then push that into SwiftInstance, but making a class a `SwiftValue` seems incorrect to me, can we share the cleanup logic by doing it with SwiftInstances instead?

rintaro: I'm not sure I understand your concerns correctly.

> we want to know about that in the Java side so that we know that we may need to copy() a value potentially etc.

You mean we might want to do like this in Java:
```java
var obj1 = new MySwiftThing(arena);
var obj_copied = obj1.$copy(arena);
```
By treating reference types as a "value" I _think_ we should be able to do it _without_ knowing whether it's a `struct` or `class`:
By like:
```java
  public MySwiftThing $copy(SwiftArena swiftArena$) {
    var segment = SwiftValueWitnessTable.initializeWithCopy(
      /*type*/this.$swiftType(),
      /*dst*/swiftArena$.allocate(this.$layout()),
      /*src*/this.$memorySegment()
    );
    return new MySwiftClass(segment, swiftArena$)
  }
```
But maybe I'm wrong. 

> Even if the cleanups we end up doing the same way, we can then push that into SwiftInstance, but making a class a SwiftValue seems incorrect to me, can we share the cleanup logic by doing it with SwiftInstances instead?

I agree using Swift**Value** might be confusing. Is your concern here just the naming? I'd be happy to make `SwiftInstance` an abstract class just like my current `SwiftValue`.

Sources/JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift
rintaro: Hmm, "accessor" sounds super confusing with variable accessor for me 😅 

https://github.com/openjdk/jextract/blob/master/doc/GUIDE.md
> The jextract tool parses header (.h) files of native libraries, and generates Java code, called **bindings**

It looks like "bindings" is the official term?

SwiftKit/src/main/java/org/swift/swiftkit/SwiftValueWitnessTable.java
ktoso: Hmm, that's an interesting idea that technically we could implement that copy... I don't know if it is meaningful in any useful way to copy a reference type like that 🤔 But let's say we can, so that's fine.

I'm concerned we'll lose typesafety in e.g. `SwiftKit.retain(SwiftHeapObject)`, which today you can't accidentally pass in a swift value type instance to. So even if the copy and initialization implementations remain the same I do think we want the `SwiftHeapObject` to remain in order to model in types that we're dealing with a class instance.

> I agree using SwiftValue might be confusing. Is your concern here just the naming? I'd be happy to make SwiftInstance an abstract class just like my current SwiftValue

Yeah primarily it's a concern about conflating classes with value types. I think what I'm proposing is:

- keep all the logic changes about cleanup, init, that you're proposing
- keep the `SwiftHeapObject` and make classes extracted `extends/implements SwiftHeapObject`
- whichever logic we can implement in common, destroys, copying etc, implement against `SwiftInstance`.

WDYT? So it's just about keeping the type so it's clear what it is, but if we can push more implementation to work with `SwiftInstance` let's do that.

Sources/JExtractSwift/Swift2JavaTranslator+JavaThunkPrinting.swift
ktoso: That sounds good! Let's start using "bindings" consistently then for the Java generated side 👍

SwiftKit/src/main/java/org/swift/swiftkit/SwiftValueWitnessTable.java
ktoso: Okey so the `SwiftValueWitnessTable.initializeWithCopy` on a class would literarily just do a _retain_... If we'll end up needing to expose a "copy" to end users for sure I would not want to expose a method called copy that just retains as that's a bit too confusing to users from Java, so either way I do think we need that type distinction between class/structs on the java side I think.

rintaro: Yeah, it's just retain for reference types (and copy the pointer to the destination), and I _thought_ that's fine.
But I think now I understand your concern; for Java users, `copy` sounds actually cloning the instance into a difference instance, okay.

ktoso: Yeah, that's the concern. If it's our internal use that's fine, but exposing it to users wouldn't be ok.

DougGregor: VWT operations always work on pointers to the "value". For heap objects, that's a pointer to the pointer. To deal with Swift types generically through the VWT, we'll need this extra indirection. I agree that we might not want to expose it to users that way, but that's how our generic lower-level layer will need to work.

rintaro: To make sure I understand, what I'm proposing in this PR is that `$memorySegment` is now always a pointer to the "value". I.e. in the Swift thunk, both for `struct` and `class` the result are returned indirectly and initialized like:
```swift
@cdecl("swiftjava_MyModule_MyClass_init")
public func swiftjava_MyModule_MyClass_init(_ _result: UnsafeRawPointer) {
  _result.assumingMemoryBound(to: MyClass.self).initialize(to: MyClass())
}
```
In Java side, that `_result` is the `$memorySegment`.  So I believe we should pass that `$memorySegment` to the VWT functions without extra indirection, because it's already is a "a pointer to the pointer".
Is my understanding correct?

DougGregor: I agree with your approach in this PR that `$memorySegment` is should always a pointer to a "value". Your understanding sounds correct to me.

ktoso: Sounds good to me, unifying those to the same initialization pattern

SwiftKit/src/main/java/org/swift/swiftkit/SwiftInstance.java
```diff
@@ -17,36 +17,87 @@
 import java.lang.foreign.GroupLayout;
 import java.lang.foreign.MemorySegment;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
 
-public interface SwiftInstance {
+public abstract class SwiftInstance {
```
rintaro: Now `SwiftInstance` is the base class for every imported Swift type. `SwiftValue` and `SwiftHeapObject` are interfaces mostly just for a marker at this point.

ktoso: Nice, thanks!

Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
```diff
@@ -26,7 +26,7 @@ public struct ForeignValueLayout: CustomStringConvertible, Equatable {
 
   public init(inlineComment: String? = nil, javaConstant: String) {
     self.inlineComment = inlineComment
-    self.value = javaConstant
+    self.value = "SwiftValueLayout.\(javaConstant)"
```
ktoso, line 26: 👍

Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
```diff
@@ -57,12 +57,12 @@ public struct ForeignValueLayout: CustomStringConvertible, Equatable {
       result.append("/*\(inlineComment)*/")
     }
 
-    result.append("SwiftValueLayout.\(value)")
+    result.append(value)
 
     // When the type is some custom type, e.g. another Swift struct that we imported,
-    // we need to import its layout. We do this by calling $layout() on it.
+    // we need to import its layout. We do this by calling $LAYOUT() on it.
```
ktoso: is this because openjdk seems to use uppercase for these?

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -0,0 +1,358 @@
...
+  ) {
+    printer.printSeparator(decl.displayName)
+
+    printJavaBindingDescriptorClass(&printer, decl)
+
+    // Render the "make the downcall" functions.
+    printFuncDowncallMethod(&printer, decl)
+  }
+
+  /// Printer FFM Java binding descriptors for the imported Swift API.
```
ktoso:
```suggestion
  /// Print FFM Java binding descriptors for the imported Swift API.
```

Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
rintaro: There is `$layout()` method as an instance method. I added `$LAYOUT()` as a static method because this is rendered as `ClassName.$LAYOUT()`. Maybe we could just use `$LAYOUT` static field, but I wasn't sure it's OK to make it `public`.

ktoso: `public static final ... LAYOUT` field is fine I think, it'd help differentiate the field and method. TBD if we need both, but we can decide this later. Uppercase like that does look a bit weird so I'd go for a constant which then looks "expected" being uppercase.

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -0,0 +1,358 @@
...
+      printer.indent()
+    } else {
+      printer.print("FunctionDescriptor.of(")
+      printer.indent()
+      printer.print("/* -> */", .continue)
+      printer.print(cFunc.resultType.foreignValueLayout, .parameterNewlineSeparator(isEmptyParam))
+    }
+
+    for (param, isLast) in cFunc.parameters.withIsLast {
+      printer.print("/* \(param.name ?? "_"): */", .continue)
```
ktoso, line 81: Minor / doesn't have to be in this PR: I think we had an effectiveParamLabel somewhere?

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -0,0 +1,358 @@
...
+    assert(decl.translatedSignature.requiresSwiftArena, "constructor always require the SwiftArena")
+    paramDecls += ", SwiftArena swiftArena$"
+
+    printer.printBraceBlock(
+      """
+      /**
+       * Create an instance of {@code \(className)}.
+       *
+       * {@snippet lang=swift :
+       * \(decl.signatureString)
```
ktoso, line 111: Follow up:

Silly but I think now if a signature contains `/**/` because we take the raw syntax with the trivia that would break rendering of the source file because the */ will end the comment.

We probably should sanitize that and add a test for it?

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -0,0 +1,358 @@
...
+  /// Print the calling body that forwards all the parameters to the `methodName`,
+  /// with adding `SwiftArena.ofAuto()` at the end.
+  public func printFuncDowncallMethod(
+    _ printer: inout CodePrinter,
+    _ decl: ImportedFunc) {
+    let methodName: String = switch decl.kind {
+    case .getter: "get\(decl.name.toCamelCase)"
+    case .setter: "set\(decl.name.toCamelCase)"
+    case .function: decl.name
+    case .initializer: fatalError("unreachable")
```
ktoso: give this a better reason string why we crash?

Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
rintaro: Sounds good. Updated. While I am here, I removed `needsMemoryLayoutCall` from the type and added  `.$LAYOUT` in the constructor.

Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
```diff
@@ -22,18 +21,16 @@ public struct ForeignValueLayout: CustomStringConvertible, Equatable {
...
+    self.value = "SwiftValueLayout.\(javaConstant)"
   }
 
   public init(inlineComment: String? = nil, customType: String) {
     self.inlineComment = inlineComment
-    self.value = customType
-    self.needsMemoryLayoutCall = true
+    // When the type is some custom type, e.g. another Swift struct that we imported,
+    // we need to import its layout. We do this by referring $LAYOUT on it.
+    self.value = "\(customType).$LAYOUT"
```
ktoso, line 33: thanks for updating that to a property

Sources/JExtractSwift/Swift2JavaVisitor.swift
```diff
@@ -171,92 +160,86 @@ final class Swift2JavaVisitor: SyntaxVisitor {
...
+    let varName = "\(binding.pattern.trimmed)"
 
     self.log.debug("Import variable: \(node.kind) '\(node.qualifiedNameForDebug)'")
 
-    let returnTy: TypeSyntax
-    if let typeAnnotation = binding.typeAnnotation {
-      returnTy = typeAnnotation.type
-    } else {
-      returnTy = "Swift.Void"
+    func importAccessor(kind: SwiftAPIKind) throws {
```
ktoso, line 167: yay!

Sources/JExtractSwift/Swift2JavaVisitor.swift
```diff
@@ -289,23 +272,3 @@ extension DeclSyntaxProtocol where Self: WithModifiersSyntax & WithAttributesSyn
     return true
   }
 }
-
-private let mangledNameCommentPrefix = "MANGLED NAME: "
-
-extension SyntaxProtocol {
-  /// Look in the comment text prior to the node to find a mangled name
-  /// identified by "// MANGLED NAME: ".
-  var mangledNameFromComment: String? {
```
ktoso, line 298: yup we dont need this anymore, thank you

Sources/JExtractSwift/SwiftThunkTranslator.swift
```diff
@@ -38,27 +136,23 @@ struct SwiftThunkTranslator {
...
-    //      if let acc = v.accessorFunc(kind: .get) {
-    //        decls.append(contentsOf: render(forFunc: acc))
-    //      }
-    //      if let acc = v.accessorFunc(kind: .set) {
-    //        decls.append(contentsOf: render(forFunc: acc))
-    //      }
-    //    }
+    for decl in nominal.methods {
+      decls.append(contentsOf: render(forFunc: decl))
+    }
```
ktoso, line 155: yay, variables - thank you

ktoso (approved):
Looks all fine to me, thank you @rintaro.

Just few minor comments, one of which you already fixed, which can be done anytime in the future. LGTM, please merge at will 👍 

FYI @madsodgaard -- with this merged, it'll be a good time for you to introduce the protocol to pick between ffm and jni modes 👍

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
rintaro: Added FIXME on `ImportedFunc.signatureString`

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
rintaro: This actually never be `nil`. `cFunc` is created from `LoweredFunctionSignature` created by `lowerFunctionSignature()` which actually gives a name to every parameter.

Sources/JExtractSwift/JavaConstants/ForeignValueLayouts.swift
ktoso: Nice, thanks 👍

PR #237, opened 2025-05-30
[jextract] Remove NominalTypeResolution

rintaro:
This is a part of #236 plus some simple test cases for `SwiftLookupTable`.

`NominalTypeResolution` was used only used for populating the `SwiftLookupTable` and resolving the extended type name for extensions. However `SwiftLookupTable` already has that capabilities (with some modifications)


Sources/JExtractSwift/SwiftTypes/SwiftNominalTypeDeclaration.swift
```diff
@@ -14,9 +14,13 @@
 
 import SwiftSyntax
 
+///// A syntax node for a nominal type declaration.
+@_spi(Testing)
+public typealias NominalTypeDeclSyntaxNode = any DeclGroupSyntax & NamedDeclSyntax & WithAttributesSyntax & WithModifiersSyntax
```
ktoso, line 19: hah interesting combo :)

ktoso (approved):
LGTM, nice!

PR #238, opened 2025-05-30
[jextract] Improve Cdecl lowering

rintaro:
Part of #236 with some test cases.

Cdecl lowering improvements without integrating into the actual workflow.

* Newly introduce `lowerResult()` (splitted from `lowerParameter()` because result lowering can be differnt from parameter lowering.
* Remove `ConversionStep.init(cdeclToSwift:)/.init(swiftToCdecl:)`. Instead embed the logic in `lowerParameter()`/`lowerResult()` so that we can implement type lowering and the actual conversion at the same place.
* Support 'String', 'Void' and '() -> Void) types
* Implement getter / setter lowering
* Fix conversion for tuple returns (see: `lowerTupleReturns()` test)
* Reference types are now returned indirectly, as the logic are being unifed. Thanks to this we don't need manual `retain()` before returning instances, which wasn't implemented correctly.


Sources/JExtractSwift/ConversionStep.swift
```diff
@@ -103,38 +76,30 @@ enum ConversionStep: Equatable {
 
   /// Convert the conversion step into an expression with the given
   /// value as the placeholder value in the expression.
-  func asExprSyntax(isSelf: Bool, placeholder: String) -> ExprSyntax {
+  func asExprSyntax(placeholder: String, bodyItems: inout [CodeBlockItemSyntax]) -> ExprSyntax? {
```
rintaro, line 79: Using `bodyItems`, `ConversionStep` can now introduce side effects, for example it can store the inner step into a variable, then return the `DeclReferenceExprSyntax` to it.

Sources/JExtractSwift/CDeclLowering/Swift2JavaTranslator+FunctionLowering.swift
```diff
@@ -47,16 +46,57 @@ extension Swift2JavaTranslator {
       symbolTable: symbolTable
     )
 
-    return try lowerFunctionSignature(signature)
+    return try CdeclLowering(swiftStdlibTypes: swiftStdlibTypes).lowerFunctionSignature(signature, apiKind: .initializer)
+  }
+
+  /// Lower the given variable decl to a C-compatible entrypoint,
+  /// providing all of the mappings between the parameter and result types
+  /// of the original function and its `@_cdecl` counterpart.
```
DougGregor, line 54: Total nit, but there's no parameter or result type for a variable, just the type itself.

Sources/JExtractSwift/SwiftTypes/SwiftFunctionSignature.swift
```diff
@@ -38,34 +43,17 @@ enum SwiftSelfParameter: Equatable {
...
 extension SwiftFunctionSignature {
   init(
     _ node: InitializerDeclSyntax,
     enclosingType: SwiftType?,
     symbolTable: SwiftSymbolTable
   ) throws {
+    // Prohibit generics for now.
+    if let generics = node.genericParameterClause {
+      throw SwiftFunctionTranslationError.generic(generics)
+    }
```
DougGregor, line 55: Thank you for making it defensive like this

Sources/JExtractSwift/SwiftTypes/SwiftFunctionSignature.swift
```diff
@@ -163,6 +157,101 @@ extension SwiftFunctionSignature {
...
+        var hasGetter = false
+        var hasSetter = false
+
+        for accessor in accessors {
+          switch accessor.accessorSpecifier {
+          case .keyword(.get), .keyword(._read), .keyword(.unsafeAddress):
+            hasGetter = true
+          case .keyword(.set), .keyword(._modify), .keyword(.unsafeMutableAddress):
+            hasSetter = true
+          default: // Ignore willSet/didSet and unknown accessors.
```
DougGregor, line 239: willSet/didSet imply that we have both getter and setter, don't they?

Sources/JExtractSwift/SwiftTypes/SwiftFunctionSignature.swift
```diff
@@ -163,6 +157,101 @@ extension SwiftFunctionSignature {
...
+            break
+          }
+        }
+
+        switch (hasGetter, hasSetter) {
+        case (true, true): return [.get, .set]
+        case (true, false): return [.get]
+        case (false, true): return [.set]
```
DougGregor, line 247: Technically, this case is unreachable in well-formed code

Tests/JExtractSwiftTests/FunctionLoweringTests.swift
```diff
@@ -41,14 +40,30 @@ final class FunctionLoweringTests {
...
 
+  @Test("Lowering String") func loweringString() throws {
+    try assertLoweredFunction(
+      """
+      func takeString(str: String) {}
+      """,
+      expectedCDecl: """
+      @_cdecl("c_takeString")
+      public func c_takeString(_ str: UnsafePointer<Int8>) {
+        takeString(str: String(cString: str))
```
DougGregor, line 59: Very useful, thank you!

DougGregor (approved):
This looks fantastic, thank you!

ktoso (approved):
Looks great! Lgtm

Sources/JExtractSwift/SwiftTypes/SwiftFunctionSignature.swift
rintaro: Ohh, I didn't know setter only variables are prohibited in Swift 😅

PR #240, opened 2025-06-01
[JExtract] Unify mechanisms between value types and reference types

rintaro:
Part of #236 with minimal `Sources/JExtractSwift` changes. All the changes for it are going to be replaced with the remaining changes in #236 . This PR is mostly for reviewing `SwiftKit` changes.

Unify memory and instance management mechanism between value types (e.g. `string` or `enum`) and reference types (e.g. `class` and `actor`). Now all imported nominal types are allocated using the value witness table, are returned indirectly, and are destroyed in the same manner.

`SwiftInstance` is now the abstract base class for all the imported types including reference types. Concrete types can simply construct it by calling `super(memorySegment, arena)`.

### DIFF
--- a/Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
+++ b/Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
@@ -290,11 +290,8 @@ extension Swift2JavaTranslator {
       parentProtocol = "SwiftValue"
     }
 
-    printer.printTypeDecl("public final class \(decl.javaClassName) implements \(parentProtocol)") {
+    printer.printTypeDecl("public final class \(decl.javaClassName) extends SwiftInstance implements \(parentProtocol)") {
       printer in
-      // ==== Storage of the class
-      printClassSelfProperty(&printer, decl)
-      printStatusFlagsField(&printer, decl)
 
       // Constants
       printClassConstants(printer: &printer)
@@ -400,35 +397,6 @@ extension Swift2JavaTranslator {
     )
   }
 
-  /// Print a property where we can store the "self" pointer of a class.
-  private func printClassSelfProperty(_ printer: inout CodePrinter, _ decl: ImportedNominalType) {
-    printer.print(
-      """
-      // Pointer to the referred to class instance's "self".
-      private final MemorySegment selfMemorySegment;
-
-      public final MemorySegment $memorySegment() {
-        return this.selfMemorySegment;
-      }
-      """
-    )
-  }
-
-  private func printStatusFlagsField(_ printer: inout CodePrinter, _ decl: ImportedNominalType) {
-    printer.print(
-      """
-      // TODO: make this a flagset integer and/or use a field updater
-      /** Used to track additional state of the underlying object, e.g. if it was explicitly destroyed. */
-      private final AtomicBoolean $state$destroyed = new AtomicBoolean(false);
-
-      @Override
-      public final AtomicBoolean $statusDestroyedFlag() {
-        return this.$state$destroyed;
-      }
-      """
-    )
-  }
-
   private func printClassMemoryLayout(_ printer: inout CodePrinter, _ decl: ImportedNominalType) {
     printer.print(
       """
@@ -472,29 +440,10 @@ extension Swift2JavaTranslator {
       \(decl.renderCommentSnippet ?? " *")
        */
       public \(parentName.unqualifiedJavaTypeName)(\(renderJavaParamDecls(decl, paramPassingStyle: .wrapper))) {
-        this(/*arena=*/null, \(renderForwardJavaParams(decl, paramPassingStyle: .wrapper)));
+        this(SwiftArena.ofAuto(), \(renderForwardJavaParams(decl, paramPassingStyle: .wrapper)));
       }
       """
     )
-
-    let initializeMemorySegment =
-      if let parent = decl.parent,
-        parent.isReferenceType
-      {
-        """
-        this.selfMemorySegment = (MemorySegment) mh$.invokeExact(
-          \(renderForwardJavaParams(decl, paramPassingStyle: nil))
-          );
-        """
-      } else {
-        """
-        this.selfMemorySegment = arena.allocate($layout());
-        mh$.invokeExact(
-          \(renderForwardJavaParams(decl, paramPassingStyle: nil)),
-          /* indirect return buffer */this.selfMemorySegment
-        );
-        """
-      }
 
     printer.print(
       """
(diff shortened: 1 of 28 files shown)


SwiftKit/src/main/java/org/swift/swiftkit/SwiftValueWitnessTable.java
```diff
@@ -285,15 +281,8 @@ public static MemorySegment initializeWithCopy(SwiftAnyType type, MemorySegment
...
-            // we need to make a pointer to the self pointer when calling witness table functions:
-            MemorySegment indirectDest = arena.allocate(SwiftValueLayout.SWIFT_POINTER);
-            MemorySegmentUtils.setSwiftPointerAddress(indirectDest, dest);
-            MemorySegment indirectSrc = arena.allocate(SwiftValueLayout.SWIFT_POINTER);
-            MemorySegmentUtils.setSwiftPointerAddress(indirectSrc, src);
-
-            var returnedDest = (MemorySegment) mh.invokeExact(indirectDest, indirectSrc, wtable);
-            return returnedDest;
+        try {
+            return (MemorySegment) mh.invokeExact(dest, src, wtable);
```
rintaro, line 285: This `initializeWithCopy` change was not included in #236. But I realized this has the same issue with `destroy`. No need for extra indirection anymore.

ktoso: Makes sense

Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
```diff
@@ -290,11 +290,8 @@ extension Swift2JavaTranslator {
       parentProtocol = "SwiftValue"
     }
 
-    printer.printTypeDecl("public final class \(decl.javaClassName) implements \(parentProtocol)") {
+    printer.printTypeDecl("public final class \(decl.javaClassName) extends SwiftInstance implements \(parentProtocol)") {
```
ktoso, line 293: Ok 👍

Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
```diff
@@ -472,30 +440,11 @@ extension Swift2JavaTranslator {
       \(decl.renderCommentSnippet ?? " *")
        */
       public \(parentName.unqualifiedJavaTypeName)(\(renderJavaParamDecls(decl, paramPassingStyle: .wrapper))) {
-        this(/*arena=*/null, \(renderForwardJavaParams(decl, paramPassingStyle: .wrapper)));
```
ktoso, line 475: I wonder if we should remove the constructors without an arena passed in actually… fine to keep in this pr, wdyt tho?

SwiftKit/src/main/java/org/swift/swiftkit/SwiftAnyType.java
```diff
@@ -34,18 +34,18 @@ public SwiftAnyType(MemorySegment memorySegment) {
...
-        }
-        this.memorySegment = type.get().memorySegment;
-    }
+//    public SwiftAnyType(SwiftHeapObject object) {
+//        if (object.$layout().name().isEmpty()) {
+//            throw new IllegalArgumentException("SwiftHeapObject must have a mangled name in order to obtain its SwiftType.");
+//        }
+//
+//        String mangledName = object.$layout().name().get();
+//        var type = SwiftKit.getTypeByMangledNameInEnvironment(mangledName);
```
rintaro: I don't feel we need this function anymore. Every instance has `$swiftType()`. I don't even think `object.$layout().name()` is a mangled name (i.e. it's initialized with `.withName(SwiftKit.nameOfSwiftType(typeMetadata, true));`)

Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
rintaro: I'm fine with removing this constructor as we don't want users to use `ofAuto()` arena as much as possible. 
Also mixing different arena in a session sounds scary 😅 E.g.
```Java
try(var arena = SwiftArena.ofConfined()) {
  var obj = new MyClass(arena);
  var value = new MyStruct(); // Accidental use of "auto" arena.
  obj.process(value); // Probably fine, but...
}
```
This is actually probably fine, but still it's not ideal.
Removing this convenience constructor prevents this kind of issues.

ktoso: Yeah, it’ll be more consistent to just always pass arenas, might save us weird headaches down the line.

you can remove here or we’ll do later, your call

rintaro: Let's do it in this PR 👍

Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
```diff
@@ -504,21 +440,24 @@ extension Swift2JavaTranslator {
        *
       \(decl.renderCommentSnippet ?? " *")
        */
-      public \(parentName.unqualifiedJavaTypeName)(SwiftArena arena, \(renderJavaParamDecls(decl, paramPassingStyle: .wrapper))) {
-        var mh$ = \(descClassIdentifier).HANDLE;
-        try {
+      public \(parentName.unqualifiedJavaTypeName)(\(renderJavaParamDecls(decl, paramPassingStyle: .wrapper)), SwiftArena arena) {
```
rintaro, line 443: I'm 90% sure that `arena` parameter at the end is more common practice.
Although I haven't been able to find any wrapper constructor examples,
* [SymbolLookup::libraryLookup(String,Arena)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/foreign/SymbolLookup.html#libraryLookup(java.lang.String,java.lang.foreign.Arena))
* [Linker::upcallStub(MethodHandle,FunctionDescriptor,Arena,Linker.Option...)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/foreign/Linker.html#upcallStub(java.lang.invoke.MethodHandle,java.lang.foreign.FunctionDescriptor,java.lang.foreign.Arena,java.lang.foreign.Linker.Option...)
)
* Generated function pointer binding `allocate` https://github.com/openjdk/jextract/blob/master/samples/libjimage/JImageFile.java#L65-L71

SwiftKit/src/main/java/org/swift/swiftkit/SwiftAnyType.java
rintaro: Removed

Sources/JExtractSwift/Swift2JavaTranslator+Printing.swift
ktoso: Last sounds ok me, makes the calls more similar to their swift equivalents and only the arena added on.

SwiftKit/src/main/java/org/swift/swiftkit/SwiftAnyType.java
ktoso: Yeah let's remove, it is from the time when we were still calling directly into initializers from the java side I think

ktoso (approved):

PR #241, opened 2025-06-01
Only enable terminal colors when supported

rintaro:
Without this change, test failure messages in Xcode console is extremely hard to read.


ktoso (approved):

ktoso:
gotcha, yeah I'm mostly from terminal all day... thanks for making this nicer for Xcode users :)

PR #246, opened 2025-06-05
[JExtract] Static 'call' method in binding descriptor classes

rintaro:
* Add a static `call` method to each binding descriptor class to handle the actual downcall.
* Refactor wrapper methods to delegate to the binding descriptor's `call` method.
* Clearly separate responsibilities: each binding descriptor class now encapsulates the complete lowered Cdecl thunk, while wrapper methods focus on Java-to-Cdecl type conversion.


ktoso (approved):
Looks good to me, you're right that centralizing the func description like that is nice 👍

PR #248, opened 2025-06-05
Converge command line tools into one: swift-java

ktoso:
Work in progress.

We want to have one command line tool to rule them all™ :-)


ktoso:
This should be complete but I need to give it a rebase now

ktoso:
This just moves code around so we have one command line tool: `swift-java` that has various modes.

There will be follow up PRs further cleaning up and making it easier to use each of the modes. Merging so we can keep iterating and rebase other PRs waiting.

ktoso:
post-merge review welcome, there will be more cleanups incoming here. Some commands need to be renamed still and I think we need to make them commands `swift-java resolve` rather than `--fetch` etc.

PR #249, opened 2025-06-05
Introduce protocol-based structure for generating code

madsodgaard:
This goal of this PR is to introduce a protocol `Swift2JavaGenerator`, which does all the code generation. This will allow us to provide multiple generation modes, such as FFM or JNI.

Most of the stuff from `Swift2JavaTranslator+...` has been moved into `FFMSwift2JavaGenerator`.


madsodgaard:
@rintaro I could use a bit of your thoughts here. Currently, it seems like the visitor depends on translated types, specifically this function https://github.com/swiftlang/swift-java/blob/99fc99c7b06f0cd28a229f842fc9f2b57bc3b4e4/Sources/JExtractSwift/Swift2JavaTranslator%2BJavaTranslation.swift#L18
used in places like
https://github.com/swiftlang/swift-java/blob/99fc99c7b06f0cd28a229f842fc9f2b57bc3b4e4/Sources/JExtractSwift/Swift2JavaVisitor.swift#L230


This stuff seems FFM specific, so should really live in the FFM implementation of the protocol and the visitor should not depend on it. It seems like the property `translatedSignature` is only actually used in the translation phase, not the analysis phase. So my initial thought was to postpone this translation to the actual FFM phase and not during the visiting phase. I just wanted to confirm with you whether you think this is a good idea, or if you have any better ideas?

Sources/JExtractSwift/Swift2JavaGenerator.swift
```diff
@@ -0,0 +1,3 @@
+protocol Swift2JavaGenerator {
+  func generate() throws
```
madsodgaard: It seems kind of weird that this protocol just has an "empty" function with no arguments. But, I did that to prevent having to pass `AnalysisResult` to a bunch of sub-functions in `print...` So instead the implementations can receive them as initializer arguments.

Sources/JExtractSwift/Swift2JavaTranslator.swift
```diff
@@ -88,6 +88,13 @@ extension Swift2JavaTranslator {
   /// a checked truncation operation at the Java/Swift board.
   var javaPrimitiveForSwiftInt: JavaType { .long }
 
+  var result: AnalysisResult {
```
madsodgaard: Perhaps this should just be the return value of `analyze()`?

rintaro:
I'm totally fine with delaying FFM `translate` after the "analysis" phase, which will mean the `AnalysisResult` (i.e. `ImportedFunc`) would only hold `SwiftFunctionSignature` at most, not `TranslatedFunctionSignature`, right?

There are a couple of reasons I didn't do that:
* Wasn't sure sharing the visitor would be the way to go. It's just 250 lines of code. JNI generator might want to create its own visitor with its own needs. We could still share  `SwiftSymbolTable` and `SwiftFunctionSignature`.
* Wanted to keep the `TranslatedFunctionSignature` somewhere around. Although I don't think it's used multiple times _at this point_, it's possible that we want to use it multiple places in the generator/printer. 

That being said, I now feel sharing `analysis` result is a good move. To resolve the 2nd bullet, probably `FFMSwift2JavaGenerator` can hold the cached `TranslatedFunctionSignature` result for each `ImportedFunc`, (just like the `thunkNameRegistry`) I.e.
```swift
var translatedSignatures: [ImportedFunc: TranslatedFunctionSignature]
```

Sources/JExtractSwift/Swift2JavaGenerator.swift
rintaro: It's not clear for me how this protocol will be used. The main flow can be just like:
```swift
  try translator.analyze()
  
  if shouldGenerateFFM {
    FFMSwift2JavaGenerator(analysis, ...).generate()
  }
  
  if shouldGenrateJNI {
    JNISwift2JavaGenerator(analysis, ...).generate()
  }
```

Could you explain why we'd want this protocol?

ktoso: Yeah that’s pretty equivalent. The protocol being just the generate method seemed nicer to me, and we’d swap out what we generate with but it’s equivalent to that if statement. Since there’s not much to share either approach is fine i think, modeling wise i like the protocol but we could do the if

ktoso:
Cool, that’s the outcome I was hoping for here — when we delay the translated signature the analysis can remain detached of the output, and FFM and JNI can do their own translation/lowering based on that result 👍

rintaro:
Okay, I am moving the FFM translation part to after `analyze()`. https://github.com/swiftlang/swift-java/pull/252

@madsodgaard Sorry for the conflicts, but I believe this makes JNI additions easier.

ktoso:
Thank you Rintaro! :)

madsodgaard:
@rintaro @ktoso I think this is ready for a first pass. For now, I have just kept the protocol, but as @rintaro pointed out, we don't actually do any dynamic swapping it in that way, but I'll just keep it around for now.

Otherwise, most of the stuff is just moving the previous translator stuff into `FFMSwift2JavaGenerator`, as well as a new `FFM` folder. The tests still assume we always test FFM, but we can perhaps look at how to tackle more generalized ways of testing both JNI and FFM once we get further and actually start generating JNI.

Sources/JExtractSwift/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -16,27 +16,87 @@ import JavaTypes
...
+    self.javaPackage = javaPackage
+    self.swiftOutputDirectory = swiftOutputDirectory
+    self.javaOutputDirectory = javaOutputDirectory
+    self.symbolTable = translator.symbolTable
+    self.swiftStdlibTypes = translator.swiftStdlibTypes
+  }
+
+  func generate() throws {
+    try writeExportedJavaSources()
+    try writeSwiftThunkSources()
```
rintaro: Just curious, any reason you changed the order? The current main writes the Swift thunk files first.

Sources/JExtractSwift/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -16,27 +16,87 @@ import JavaTypes
...
+  let analysis: AnalysisResult
+  let swiftModuleName: String
+  let javaPackage: String
+  let swiftOutputDirectory: String
+  let javaOutputDirectory: String
+  let swiftStdlibTypes: SwiftStandardLibraryTypes
+  let symbolTable: SwiftSymbolTable
+
+  var javaPackagePath: String {
+    javaPackage.replacingOccurrences(of: ".", with: "/")
```
rintaro: (Not in this PR, but we should consistently use `PATH_SEPARATOR`, or more preferably `swift-system`'s `FilePath`)

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
```
rintaro, line 5: Does this affect Java files too? Currently handwritten Java files in this repo use 4 spaces indentation.

rintaro (commented):
The code moving part looks fine to me. I'll defer the command-line option part to @ktoso as it will affect #248 

(As for the test failures, the workflow still uses older compiler which doesn't support trailing commas in parametar clauses)

Sources/JExtractSwift/FFM/FFMSwift2JavaGenerator.swift
madsodgaard: hm no, I can change it back if you'd prefer that?

.editorconfig
madsodgaard: probably does. I've modified it to only include .swift files

ktoso: 2 spaces is fine, we cans witch over to that consistently tbh.

The formatter plugin used to hang in github actions so we don't have it enabled right now, but we should revisit this again.

Sources/JExtractSwift/FFM/FFMSwift2JavaGenerator.swift
ktoso: It shouldn't really matter but let's not do changes when we don't need to do them -- can you bring back the old order here?

ktoso:
The license header failure is about Sources/JExtractSwift/GenerationMode.swift and other files missing the header we have in other files, can you add these please? :)

ktoso:
I merged the tools rename, I can do the conflict resolution or leave it you -- let's catch up on chat :)

Sources/JExtractSwift/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -16,27 +16,87 @@ import JavaTypes
 import SwiftSyntax
 import SwiftSyntaxBuilder
 
+package class FFMSwift2JavaGenerator: Swift2JavaGenerator {
+  let log = Logger(label: "ffm-generator", logLevel: .info)
```
ktoso: Minor: Probably need to make it a var and copy the log level from translator.log.logLevel during init `self.log.logLevel = translator.log.logLevel`

Sources/JExtractSwift/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -16,27 +16,87 @@ import JavaTypes
...
+  /// Every imported public type becomes a public class in its own file in Java.
+  package func writeExportedJavaSources(printer: inout CodePrinter) throws {
+    for (_, ty) in analysis.importedTypes.sorted(by: { (lhs, rhs) in lhs.key < rhs.key }) {
       let filename = "\(ty.swiftNominal.name).java"
       log.info("Printing contents: \(filename)")
       printImportedNominal(&printer, ty)
 
       if let outputFile = try printer.writeContents(
-        outputDirectory: outputDirectory,
+        outputDirectory: javaOutputDirectory,
```
ktoso: Good! Btw I think we should start consistently prefixing things with java or swift; We used to have a bit of a mix, we'll now start converging on everything being prefixed with swift/java depending on which bit it applies to (like moduleName` I think we need to make into `swiftModule` etc.

Sources/JExtractSwift/Swift2Java.swift
```diff
@@ -45,14 +45,18 @@ public struct SwiftToJava: ParsableCommand {
   @Option(name: .shortAndLong, help: "Configure the level of lots that should be printed")
   var logLevel: Logger.Level = .info
 
+  @Option(
+    name: .long,
+    help: "The mode of generation to use for the output files.")
+  var mode: GenerationMode = .ffm
```
ktoso: `--mode` sounds good to me 👍

Sources/JExtractSwift/Swift2JavaTranslator.swift
ktoso: Yeah, I think that'll be nicer.

The translator still is stateful since it may keep names for resolution etc but it'll be nicer to return the result like that 👍

ktoso (approved):

rintaro:
Not sure why https://github.com/swiftlang/swift-java/pull/258 doesn't work 🤔 Maybe rebasing required?

> ```
> HEAD is now at 7b8d615 Merge 30b0ce0fdfe1e751d13ff5540cdb336ee94b1b37 into fbebbe4fff84e903976f6cb5b5647ab915bfa235
> ```

fbebbe4fff84e903976f6cb5b5647ab915bfa235 is the `main` ref _before_ #258 merged

PR #250, opened 2025-06-05
[JExtract] Santize trivia for declartion signature string

rintaro:
Introduce `triviaSanitizedDescription` that prints tokens with trivia condensed into a single whitespace, or removed after opening or before closing parentheses.


ktoso (approved):
Nice, thank you!

PR #252, opened 2025-06-05
[JExtract] Lazy Cdecl lowering and Java translation 

rintaro:
(Based on #251)

Make `Swift2JavaTranslator.analyze()` analyze only Swift signatures.
Postpone Cdecl lowering and Java translation for cleaner separation of responsibilities.


ktoso (approved):
Nice, thank you!

PR #253, opened 2025-06-06
[WIP][JExtract] Import any C compatible closure

rintaro:
Generalized closure parameter import, part 1.

Start with C-compatible closures which don't need any conversions.


Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -124,16 +128,155 @@ extension Swift2JavaTranslator {
...
+    _ bindingDescriptorName: String
+  ) {
+    let cdeclDescritor = "\(bindingDescriptorName).$\(functionType.name)"
+    if functionType.isTrivial {
+      // If the usser-facing function interface is C-compatible, just extend the
+      // lowered function pointer parameter interface.
+      printer.print(
+        """
+        public interface \(functionType.name) extends \(cdeclDescritor).Function {
+          default MemorySegment toUpcallStub(Arena arena) {
```
rintaro: Maybe it's better to make a static func outside the interface, so it won't leak `toUpcallStub` method?
```java
private static func $toUpcallStub(\(functionType.name) fi, Arena arena) { ... }
```

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -124,16 +128,155 @@ extension Swift2JavaTranslator {
...
+  ///     interface Function {
+  ///       <return-type> apply(<parameters>);
+  ///     }
+  ///     static final MethodDescriptor DESC = FunctionDescriptor.of(...);
+  ///     static final MethodHandle HANDLE = SwiftKit.upcallHandle(Function.class, "apply", DESC);
+  ///     static MemorySegment toUpcallStub(Function fi, Arena arena) {
+  ///       return Linker.nativeLinker().upcallStub(HANDLE.bindTo(fi), DESC, arena);
+  ///     }
+  ///   }
+  ///   ```
```
ktoso: That's pretty good structure, I like it 👍

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -124,16 +128,155 @@ extension Swift2JavaTranslator {
...
+  func printFunctionPointerParameterDescriptorClass(
+    _ printer: inout CodePrinter,
+    _ name: String,
+    _ cType: CType
+  ) {
+    guard case .pointer(.function(let cResultType, let cParameterTypes, variadic: false)) = cType else {
+      preconditionFailure("must be a C function pointer type")
```
ktoso: Include the name and cType in the crash message?

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -124,16 +128,155 @@ extension Swift2JavaTranslator {
...
+       * {snippet lang=c :
+       * \(cType)
+       * }
+       */
+      private static class \(name)
+      """
+    ) { printer in
+      printer.print(
+        """
+        public interface Function {
```
ktoso:
```suggestion
        @FunctionalInterface
        public interface Function {
```

This isn't "required" in order for an interface to be a functional interface, but it's a good pattern.

What this does is that if we accidentally added one more non-default method to the interface then the java compiler would then error about it. That's important because we want callers to use the () -> {} lambda syntax rather than spell out the explicit types

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
ktoso: Agreed, that'll be nicer. In order keep the interface without default impls so users aren't confused what those funcs are about.

Most importantly we want only one non-default method, but it'll be nice to have those not have any defaulted ones either

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
```diff
@@ -124,16 +128,155 @@ extension Swift2JavaTranslator {
...
+        public interface \(functionType.name) extends \(cdeclDescritor).Function {
+          default MemorySegment toUpcallStub(Arena arena) {
+            return \(bindingDescriptorName).$\(functionType.name).toUpcallStub(this, arena);
+          }
+        }
+        """
+      )
+    } else {
+      // Otherwise, the lambda must be wrapped with the lowered function instance.
+      assertionFailure("should be unreachable at this point")
```
ktoso: If always unreachable, should we make this fatal error and remove the code below it? It's not like we'll be testing if that code is correct after all right?

Sources/JExtractSwift/Swift2JavaTranslator+JavaTranslation.swift
```diff
@@ -82,19 +79,46 @@ struct TranslatedResult {
...
 /// Since this holds the lowered signature, and the original `SwiftFunctionSignature`
 /// in it, this contains all the API information (except the name) to generate the
 /// cdecl thunk, Java binding, and the Java wrapper function.
-struct TranslatedFunctionSignature {
-  var loweredSignature: LoweredFunctionSignature
+struct TranslatedFunctionDecl {
+  /// Java function name.
+  let name: String
 
+  /// Function interfaces (i.e. lambda types) required for the Java method.
```
ktoso:
```suggestion
  /// Function interfaces (i.e. functional interface types) required for the Java method.
```

Nitpick but there's no real lambda types in java, just functional interface types which happen to get implemented with a lambda expression

Sources/JExtractSwift/Swift2JavaTranslator+JavaTranslation.swift
```diff
@@ -82,19 +79,46 @@ struct TranslatedResult {
...
 
+/// Represent a Swift closure type in the user facing Java API.
+///
+/// Closures are translated to named functional interfaces in Java.
+struct TranslatedFunctionType {
+  var name: String
+  var parameters: [TranslatedParameter]
+  var result: TranslatedResult
+
+  var isTrivial: Bool {
```
ktoso: Is trivial the right word? or are those c-convention compatible or something like that?

Sources/JExtractSwift/Swift2JavaTranslator+JavaTranslation.swift
```diff
@@ -121,12 +145,92 @@ extension TranslatedFunctionSignature {
...
+
+    return TranslatedFunctionDecl(
+      name: javaName,
+      functionTypes: funcTypes,
+      translatedSignature: translatedSignature,
+      loweredSignature: loweredSignature
+    )
+  }
+
+  /// Translate Swift closure type to Java function interface.
```
ktoso:
```suggestion
  /// Translate Swift closure type to Java functional interface.
```

minor wording nitpick

ktoso (approved):

ktoso:
Sorry about the conflicts 🙏 No more big "move everything" incoming now anymore I think after the big command refactor.

Sources/JExtractSwift/Swift2JavaTranslator+JavaBindingsPrinting.swift
rintaro: I'd prefer to keep the code below for now. It illustrates how things would work for non-C-compat closures, and it's also something I plan to implement _pretty_ soon.

rintaro:
(holding off merging until #249 is merged)

ktoso:
https://github.com/swiftlang/swift-java/pull/249 is merged so this needs a rebase and can be merged at will! 👍

PR #255, opened 2025-06-09
[CI] Try out validating samples in individual jobs

ktoso:
This way we get more parallelism and easier to spot signal which sample breaks when it does.

Since the samples are our primary test drivers given all the source generation and fetching, giving them individual jobs I think is reasonable.


ktoso:
This gives us quicker CI turnaround and better result reporting.

PR #256, opened 2025-06-09
[JavaRuntime] Add Android shims

colemancda:
Adds C shims written by @lhoward for Android support.


Sources/JavaRuntime/AndroidSupport.cpp
```diff
@@ -0,0 +1,98 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#ifdef __ANDROID__
```
ktoso, line 15: Comment from other thread @lhoward:

> This code needs to be removed and replaced with targeting a NDK version which exports the JNI.
> 
> It should be fairly trivial to do with a current Swift Android SDK, I just haven't had time to do it.

How about we do these steps to use the current android sdk right away?

lhoward: I'll try and take a look tomorrow.

ktoso: Thank you!

lhoward: In short should just need to link against `libnativehelper.so` and set the NDK version to >=31, which can be done by passing `--swift-sdk aarch64-unknown-linux-android31` to the build command line. Building now to test.

lhoward: I don't have time to test in the emulator right now but see d6db74a3991aa6aa9caf4343c52f751138a031ec.

colemancda: I will close this PR then. Thanks

colemancda: @lhoward Wouldnt the `JavaObjectHolder.swift` changes still be needed though?

lhoward: > @lhoward Wouldnt the `JavaObjectHolder.swift` changes still be needed though?

Yes, that is in a separate commit, see my open PRs.

lhoward: Ah I see what I did, I squashed removing `AndroidSupport.cpp` into the sample commit. Let me fix that now.

lhoward: OK, so my PR has three commits:

* ba56de0a03c84ef38a8ef0feb5a4efc200a92c6f which contains Android-specific changes to the top-level package
* 4d4d60e0fabe5d75b3817e64f03fb961f2e70b1e is the `JavaObjectHolder.swift` to which you refer. This is broken out separately as I believe it really needs to be applied on all platforms.
* 79d7d0d71298964ce04a9f56e501f29020efc6e9 is a sample Android application

PR #267, opened 2025-06-11
feature: Code enhancement

KeoFoxy:
- Omitted `internal` access control. No need in explicit declaration because it is default
- Following SE-0345 refactored if let constructions
- Fixed some typos

### DIFF
--- a/Sources/_Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/_Subprocess/Platforms/Subprocess+Darwin.swift
@@ -224,7 +224,7 @@ extension Subprocess {
         public var launchRequirementData: Data? = nil
         /// An ordered list of steps in order to tear down the child
         /// process in case the parent task is cancelled before
-        /// the child proces terminates.
+        /// the child process terminates.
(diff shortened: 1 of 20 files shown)


Sources/_Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -32,9 +32,9 @@ import _CShims
 
 // Darwin specific implementation
 extension Subprocess.Configuration {
```
ktoso: Please don't modify anything in `_Subprocess`, this is a verbatim source copy from the subprocess project https://github.com/swiftlang/swift-subprocess while we wait for an official release, doing changes here will make syncing with upstream harder.

KeoFoxy: Thank you for your review! I discarded all the changes in Subprocess folder

ktoso: Thanks!

ktoso:
Thanks :)

PR #268, opened 2025-06-12
Extract `configure` into its own subcommand (first of such extractions)

ktoso:
We're going towards a shape where `swift-java` has sub commands for the independent tasks it can perform: `swift-java [configure|resolve|jextract|wrap-java???]` and they have some **common options** but also some options that only make sense for some of the sub commands.

This is the first steps towards this world of a more cleaned-up swift-java command line tool.


Sources/SwiftJavaTool/CommonOptions.swift
```diff
@@ -0,0 +1,59 @@
...
+import JavaKitConfigurationShared
+import JavaKitShared
+
+protocol HasCommonOptions {
+  var commonOptions: SwiftJava.CommonOptions { get set }
+}
+
+protocol HasCommonJVMOptions {
+  var commonJVMOptions: SwiftJava.CommonJVMOptions { get set }
+}
```
rintaro, line 34: I don't think these protocols are actually used at this point. Any plan to utilize these?

Maybe like this?
```swift
extension HasCommonOptions {
  var outputDirectory: String? { commonOptions.outputDirectory }
  var inputSwift: String? { commonOptions.inputSwift }
  var logLevel: Logger.Level { commonOptions.logLevel }
}
```

Sources/JExtractSwiftLib/Swift2Java.swift
```diff
@@ -26,12 +26,12 @@ public struct SwiftToJava {
   }
 
   public func run() throws {
-    guard let swiftModule = config.swiftModule else {
+    guard let moduleName = config.swiftModule else {
```
rintaro: You're changing "module name" to "swift module", but this is opposite? 🤔

Sources/SwiftJavaTool/SwiftJava+FetchDependencies.swift
rintaro, line 1: Rename to `SwiftJava+ResolveCommand.swift`?

Sources/SwiftJavaTool/SwiftJava+EmitConfiguration.swift
rintaro, line 1: I think it'd be easier to navigate if the file name matches the actual type name `ConfigureCommand`
`SwiftJava+ConfigureCommand.swift`, or maybe `Commands/ConfigureCommand.swift`

ktoso: Yeha I'm ok with matching the command type names, we cna do the nesting maybe

Sources/JExtractSwiftLib/Swift2Java.swift
ktoso: Good catch, I probably resolved a conflict the wrong way

Sources/SwiftJavaTool/CommonOptions.swift
ktoso: The primary reason was trying to rip out common functions like `write...` into extension `where Self: HasCommonOptions` but I didn't do that yet.

Your suggestion is also nice. I'll see how much use we get out of them and if none we could remove it

PR #269, opened 2025-06-12
[jextract] Generate code for free functions with primitive types in JNI mode

madsodgaard:
This PR adds support for a new mode for `jextract` that would generate JNI code instead.

We will start of with generating Java bindings and Swift thunks for free functions and primitive types.


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
```diff
@@ -0,0 +1,152 @@
...
+  private func printModuleClass(_ printer: inout CodePrinter, body: (inout CodePrinter) -> Void) {
+    printer.printBraceBlock("public final class \(swiftModuleName)") { printer in
+      body(&printer)
+    }
+  }
+
+  private func printFunctionBinding(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
+    let returnType = decl.functionSignature.result.type.javaType
+    let params = decl.functionSignature.parameters.map {
+      "\($0.type.javaType) \($0.parameterName!)"
```
madsodgaard: Should probably safeguard the same way FFM does, by just naming them `_0`, `_1` etc.. however, I am not sure when `parameterName` actually can be `nil`? 🤔

rintaro: `func foo(_: Int)` case _should_ be `nil`. Not so common, but for example:
```swift
protocol SomeProtocol {
  init()
}
func foo<T: SomeProtocol>(_: T.Type) -> T {
  return T.init()
}
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
```diff
@@ -0,0 +1,152 @@
...
+        return javaType
+      }
+
+      fatalError("unsupported nominal type: \(nominalType)")
+
+    case .tuple([]):
+      return .void
+
+    case .metatype, .optional, .tuple, .function:
+      fatalError("unsupported")
```
ktoso:
```suggestion
      fatalError("unsupported type: \(self)")
```

Just so we have some more output in the crash if it'll happen somewhere out in the wild

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
ktoso: yeah we should handle the nil, it's not very common but we should handle it.

We can maybe centralize it somehow to provide some one place where we do `effectiveJavaParameters()`? The impl of naming those could be just:

```
[...].enumerated().map { idx, param in
  "\(param.type.javaType) \(param.paremeterName ?? "arg\(idx)")"
}
```

At the same time maybe it's a good time to then fix them to become maybe `arg0` which is a common Java way to spell those "when names are unknown" 🤔

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
```diff
@@ -0,0 +1,152 @@
...
+      body(&printer)
+    }
+  }
+
+  private func printFunctionBinding(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
+    let returnType = decl.functionSignature.result.type.javaType
+    let params = decl.functionSignature.parameters.map {
+      "\($0.type.javaType) \($0.parameterName!)"
+    }
+
```
ktoso, line 200: TODO: doing the comment with the Swift signature here :)  (the bit with `{@snippet lang=swift :`)

ktoso:
Good first step -- consider adding a test as well so we can keep the generated signatures tested using unit tests.

I'd be happy to just merge these small steps one by one but let's add a small test here maybe? 👍

madsodgaard:
The last thing I need to handle in this PR, is the unsigned integers which do not conform to `JavaValue`. I guess we should just not support generating code for them, for now?

ktoso:
Yeah totally ok to do that separately 👍 Please file an issue about it so we don't forget :)

Tests/JExtractSwiftTests/JNI/JNIModuleTests.swift
```diff
@@ -0,0 +1,96 @@
+import JExtractSwiftLib
```
ktoso, line 15: missing license header, watch out for those when creating new files

ktoso (approved):
LGTM :-) Great first step and thanks for adding the tests!

PR #270, opened 2025-06-13
WIP: Ideas about different allocators

ktoso:
Just playing around with getting Unsafe when we can't get the Arena for allocation etc.


ktoso:
For now we don't need these, closing

PR #273, opened 2025-06-15
[jextract] add support for strings in JNI

madsodgaard:
Adds simple support for strings in JNI generation mode.


ktoso (approved):
Very nice 👍 I'm wondering if worth adding an "runtime test" by adding this to a JNI example app, we can do this later though

Issue #261, opened 2025-06-10
Command line restructuring: jextract command


ktoso:
We should make jextract a subcommand of swift-java rather than infer from options

Issue #260, opened 2025-06-10
Command line restructuring: resolve command


ktoso:
The way to resolve dependencies should be `swift-java resolve` a subcommand, rather than infer the behavior from passed parameters/options.

PR #276, opened 2025-06-17
Resolve & jextract subcommands and further dis-entangle commands

ktoso:
This is a big refactor; but completes our moves to subcommands and dis-entangles all the commands from eachother.

This also:

- bumps swift to 6.1.2
- bumps Java to the supported non-LTS 24 release -- this is the last non-LTS before 25 which will be released in September

❗ When developing, **please make sure that your JAVA_HOME uses e.g. JDK21**, but you have 24 installed. This way gradle will build using 21 for running "itself" but will use the JDK 24 for the actual code.

resolves https://github.com/swiftlang/swift-java/issues/261
resolves https://github.com/swiftlang/swift-java/issues/260


ktoso:
In order to land this big refactor, squashed this all into one commit; and will be doing smaller followups from here onwards.

Any review comments please do leave if you have any and we'll follow up in separate PRs!

PR #279, opened 2025-06-17
[JExtract] Bridge UnsafeRawBufferPointer

rintaro:
Lower `Unsafe(Mutable)RawBufferPointer` to a pair of pointer and count. Implement parameter and result conversion both in Swift thunks and Java bindings.

`JavaConversionStep` is now nested structure just like `ConversionStep`.


ktoso (approved):
Nice, lgtm!

PR #282, opened 2025-06-23
[JExtract] Bridge closures with UnsafeRawBufferPointer parameter

rintaro:
First step to bridging closures with conversions.

E.g.
```swift
public func withBuffer(body: (UnsafeRawBufferPointer) -> Void)
```
is lowered to:
```c
void c_withBuffer(void (*body)(const void *, ptrdiff_t));
```
then translated to:
```java
public static class withBuffer {
    @FunctionalInterface
    public interface body {
      void apply(java.lang.foreign.MemorySegment _0);
    }
}
public static void withBuffer(withBuffer.body body);
```

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/CDeclLowering/FFMSwift2JavaGenerator+FunctionLowering.swift
+++ b/Sources/JExtractSwiftLib/FFM/CDeclLowering/FFMSwift2JavaGenerator+FunctionLowering.swift
@@ -332,14 +332,15 @@ struct CdeclLowering {
     var parameters: [SwiftParameter] = []
     var parameterConversions: [ConversionStep] = []
 
-    for parameter in fn.parameters {
-      if let _ = try? CType(cdeclType: parameter.type) {
-        parameters.append(SwiftParameter(convention: .byValue, type: parameter.type))
-        parameterConversions.append(.placeholder)
-      } else {
-        // Non-trivial types are not yet supported.
-        throw LoweringError.unhandledType(.function(fn))
-      }
+    for (i, parameter) in fn.parameters.enumerated() {
+      let parameterName = parameter.parameterName ?? "_\(i)"
+      let loweredParam = try lowerClosureParameter(
+        parameter.type,
+        convention: parameter.convention,
+        parameterName: parameterName
+      )
+      parameters.append(contentsOf: loweredParam.cdeclParameters)
+      parameterConversions.append(loweredParam.conversion)
     }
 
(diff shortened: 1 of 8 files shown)


Sources/JExtractSwiftLib/FFM/CDeclLowering/FFMSwift2JavaGenerator+FunctionLowering.swift
```diff
@@ -352,15 +353,74 @@ struct CdeclLowering {
...
+    let isCompatibleWithC = parameterConversions.allSatisfy(\.isPlaceholder) && resultConversion.isPlaceholder
 
     return (
       type: .function(SwiftFunctionType(convention: .c, parameters: parameters, resultType: resultType)),
-      conversion: .placeholder
+      conversion: isCompatibleWithC ? .placeholder : .closureLowering(parameters: parameterConversions, result: resultConversion)
     )
   }
 
+  func lowerClosureParameter(
```
rintaro, line 364: It's unfortunate that we now have _another_ lower/translate function for closure parameters (in addition to regular parameters, and results), and probably we'd add more for closure _result_ conversions. But this works for now, and we can consolidate them after we implement more types, and find common logics among them.

Sources/JExtractSwiftLib/FFM/ConversionStep.swift
```diff
@@ -165,6 +174,48 @@ enum ConversionStep: Equatable {
...
+      var closure = ExprSyntax(
+        """
+        { (\(raw: params.joined(separator: ", "))) in
+          }
+        """
+      ).cast(ClosureExprSyntax.self)
+
+      closure.statements = CodeBlockItemListSyntax {
+        body.map {
+          $0.with(\.leadingTrivia, [.newlines(1), .spaces(4)])
```
ktoso, line 215: Oh I didn't know about .newlines, that's nice

Sources/JExtractSwiftLib/FFM/CDeclLowering/FFMSwift2JavaGenerator+FunctionLowering.swift
ktoso: Yeah, but still managable... agreed on converging later if we find ways to do so 👍

ktoso:
Happy to merge this now and I'll finish up landing the commands and test infra fixes today on top of this 👍 Should have not impact really on that ongoing work.

PR #284, opened 2025-06-25
CI: Correct validation & bring back pointer "max" bounds workaround

ktoso:
FYI @rintaro bringing back the workaround for the ADDRESS bounds.

I'll look into the real solution here though, to properly size things when we use a pointer.

### DIFF
--- a/.github/scripts/validate_sample.sh
+++ b/.github/scripts/validate_sample.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -e
+set -x
+
 # shellcheck disable=SC2034
 declare -r GREEN='\033[0;32m'
 declare -r BOLD='\033[1m'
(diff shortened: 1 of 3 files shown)


.github/scripts/validate_sample.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -e
+set -x
```
ktoso, line 4: duh! we missed those

.github/scripts/validate_sample.sh
```diff
@@ -14,14 +17,11 @@ echo "========================================================================"
...
-  echo -e "${BOLD}Gradle${RESET} build..."
-  ./gradlew build || ./gradlew build --info # re-run to get better failure output
+  echo -e "Run ${BOLD}${CI_VALIDATE_SCRIPT}${RESET} script..."
+  ./${CI_VALIDATE_SCRIPT}
 else
-  echo -e "${BOLD}SwiftPM${RESET} build..."
-  swift build || exit
+  echo -e "${BOLD}Missing ${CI_VALIDATE_SCRIPT} file!${RESET}"
+  exit
 fi
```
ktoso, line 25: Yeah let's just have each file be explicit about how it validates instead...

Issue #229, opened 2025-05-10
Swift 6.1: JavaKitSampleApp fails with redefinition of module 'JavaRuntime'


ykws:
I encountered a module redefinition error when building `JavaKitSampleApp` using `swift build`.
It looks similar to the following issue.

- #183 

### Swift
```
% swift --version
swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0
```

```
% xcrun --toolchain swift swift --version
Apple Swift version 6.2-dev (LLVM 815013bbc318474, Swift 1459ecafa998782)
Target: arm64-apple-macosx15.0
Build config: +assertions
```

### Java
```
% java --version
java 22.0.2 2024-07-16
Java(TM) SE Runtime Environment (build 22.0.2+9-70)
Java HotSpot(TM) 64-Bit Server VM (build 22.0.2+9-70, mixed mode, sharing)
```

### Steps
```
% cd Samples/JavaKitSampleApp
% swift build
```

or

```
% xcrun --toolchain swift swift build
```

### Error
```
error: emit-module command failed with exit code 1 (use -v to see invocation)
/Users/ykws/Developer/GitHub/swiftlang/swift-java/Samples/JavaKitSampleApp/.build/arm64-apple-macosx/debug/JavaRuntime-tool.build/module.modulemap:1:8: error: redefinition of module 'JavaRuntime'
1 | module JavaRuntime {
  |        `- error: redefinition of module 'JavaRuntime'
2 |     umbrella header "/Users/ykws/Developer/GitHub/swiftlang/swift-java/Sources/JavaRuntime/include/JavaRuntime.h"
3 |     export *

/Users/ykws/Developer/GitHub/swiftlang/swift-java/Samples/JavaKitSampleApp/.build/arm64-apple-macosx/debug/JavaRuntime.build/module.modulemap:1:8: note: previously defined here
1 | module JavaRuntime {
  |        `- note: previously defined here
2 |     umbrella header "/Users/ykws/Developer/GitHub/swiftlang/swift-java/Sources/JavaRuntime/include/JavaRuntime.h"
3 |     export *

<unknown>:0: error: could not build Objective-C module 'SwiftShims'
[...]

Issue #183, opened 2024-11-19
Objective-C Errors when building


pnewell:
I am getting a bunch of objective-c errors when trying to run  `make javakit-run` (see below)

I am one an M3 MacBook Air.

This is java --version:
```
openjdk 23.0.1 2024-10-15
OpenJDK Runtime Environment (build 23.0.1+11-39)
OpenJDK 64-Bit Server VM (build 23.0.1+11-39, mixed mode, sharing)
```
Java was installed with brew (openjdk) and JAVA_HOME is pointed at `/Users/REDACTED/Library/Java/JavaVirtualMachines/openjdk-23.0.1/Contents/Home`

This is swift --version:
```
Apple Swift version 6.1-dev (LLVM c7635515f932833, Swift dbf388c0a58df0e)
Target: arm64-apple-macosx15.0
```

Here are the errors:
```
Writing Java class 'com.example.swift.HelloSubclass' translation to '/Users/REDACTED/Downloads/swift-java-main/Samples/JavaKitSampleApp/.build/plugins/outputs/javakitsampleapp/JavaKitExample/destination/Java2SwiftPlugin/generated/HelloSubclass.swift'... done.
Writing Java class 'com.example.swift.JavaKitSampleMain' translation to '/Users/REDACTED/Downloads/swift-java-main/Samples/JavaKitSampleApp/.build/plugins/outputs/javakitsampleapp/JavaKitExample/destination/Java2SwiftPlugin/generated/JavaKitSampleMain.swift'... done.
Writing Java class 'com.example.swift.HelloSwift' translation to '/Users/REDACTED/Downloads/swift-java-main/Samples/JavaKitSampleApp/.build/plugins/outputs/javakitsampleapp/JavaKitExample/destination/Java2SwiftPlugin/generated/HelloSwift.swift'... done.
/Users/REDACTED/Downloads/swift-java-main/Samples/JavaKitSampleApp/.build/arm64-apple-macosx/debug/JavaRuntime-tool.build/module.modulemap:1:8: error: redefinition of module 'JavaRuntime'
1 | module JavaRuntime {
  |        `- error: redefinition of module 'JavaRuntime'
2 |     umbrella header "/Users/REDACTED/Downloads/swift-java-main/Sources/JavaRuntime/include/JavaRuntime.h"
3 |     export *

[...]

PR #285, opened 2025-06-25
[JavaRuntime] Add module.modulemap

rintaro:
Fix a build error in 6.1 toolchain. Apparently some targets have both `JavaRuntime-tool.build` and `JavaRuntime.build` as the include paths. The auto generated `module.modulemap` in those directories cause the error `error: redefinition of module 'JavaRuntime'`. To avoid that, add `module.modulemap` in the source directory so it's not generated.

Fixes: https://github.com/swiftlang/swift-java/issues/229


ktoso:
Nice! 

This probably solves https://github.com/swiftlang/swift-java/issues/183 as well

Resolves: #183

PR #290, opened 2025-06-25
[JExtract] Stop using SyntaxVisitor for Swift2JavaVisitor

rintaro:
Visit manually instead of using SyntaxVisitor:

* Better visitation control than `SyntaxVisitorContinueKind`
* Robust type context control by passing it as a visit parameter.

### DIFF
--- a/Sources/JExtractSwiftLib/ImportedDecls.swift
+++ b/Sources/JExtractSwiftLib/ImportedDecls.swift
@@ -37,8 +37,8 @@ package class ImportedNominalType: ImportedDecl {
     self.swiftNominal = swiftNominal
   }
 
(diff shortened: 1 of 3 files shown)


Sources/JExtractSwiftLib/ImportedDecls.swift
```diff
@@ -37,8 +37,8 @@ package class ImportedNominalType: ImportedDecl {
     self.swiftNominal = swiftNominal
   }
 
-  var javaClassName: String {
-    swiftNominal.name
```
rintaro, line 41: This wasn't used.

ktoso:
Looks nicer, thank you :)

PR #292, opened 2025-06-30
ci: cleanup samples testing using matrix

ktoso:


ktoso:
This CI setup fixes a lot of duplication we had; we have a flaky test (or rather, perhaps a real issue in one of the samples), merging to keep iterating on CI setup

PR #293, opened 2025-06-30
ci: make workflow to verify samples, list all subdirs

ktoso:
New workflow to list and validate all samples


ktoso:
I'll revisit this later fresh

PR #294, opened 2025-06-30
ci: simplistic way to get macOS running

ktoso:


ktoso:
This can be further improved in the future, but first step is done -- we have macOS CI now :) With some duplication in the workflow definitions but we can keep working on that

PR #295, opened 2025-06-30
Add JExtract JNI sample app

madsodgaard:
Adds a new sample app that actually does runtime "tests" for the new JNI mode for jextract.

### DIFF
--- a/Samples/JExtractJNISampleApp/build.gradle
+++ b/Samples/JExtractJNISampleApp/build.gradle
@@ -159,7 +159,7 @@ tasks.named('test', Test) {
 }
 
 application {
-    mainClass = "com.example.swift.HelloJava2Swift"
+    mainClass = "com.example.swift.HelloJava2SwiftJNI"
 
     applicationDefaultJvmArgs = [
         "--enable-native-access=ALL-UNNAMED",


Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2Swift.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.example.swift;
+
+// Import swift-extract generated sources
+
+// Import javakit/swiftkit support libraries
+
+import org.swift.swiftkit.SwiftKit;
+
+public class HelloJava2Swift {
```
ktoso: let's call it
```suggestion
public class HelloJava2SwiftJNI {
```

maybe?

Samples/JExtractJNISampleApp/Sources/MySwiftLibrary/swift-java.config
```diff
@@ -0,0 +1,4 @@
+{
+  "javaPackage": "com.example.swift",
+  "mode": "jni"
```
ktoso, line 3: 😄

ktoso:
Sadly can't just rename a public class in Java without also renaming the filename :) 

Good that the testing works and failed properly though ;)

ktoso:
Nice, let's include this and I'll make sure it works in the new macOS CI as well tomorrow

PR #298, opened 2025-07-02
[JExtract] Generate JNI code for classes (static methods and initializers)

madsodgaard:
This PR adds support for generating Java wrappers for Swift classes. It generates static methods and initializers.

Currently, this PR is only focused on actual allocation and initialization. Memory management is not yet addressed. We first need to provide a non-FFM version of `SwiftKit` and its tools for memory management, @ktoso said he would look into this.

This PR differs from the way FFM does it by delegating memory allocations to Swift instead of Java. This means we do not have to rely on any "unsafe" Java APIs for allocating memory without FFM. It also simplifies the code needed to allocate for structs, for example. We can easily change this approach later, if we decide to.

The code generates a java method `init`, which calls down into a native `allocatingInit` which does the actual allocation and returns the pointer to Java, for it to store. The plan is later to migrate over to something similiar to `SwiftInstance` to be used for memory cleanup.

### DIFF
--- a/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
+++ b/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
@@ -1,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import JavaTypes
+
+extension JavaType {
+  var jniTypeSignature: String {
+    switch self {
+    case .boolean: "Z"
+    case .byte: "B"
+    case .char: "C"
+    case .short: "S"
+    case .int: "I"
+    case .long: "J"
+    case .float: "F"
+    case .double: "D"
+    case .class(let package, let name):
+      if let package {
+        "L\(package.replacingOccurrences(of: ".", with: "/"))/\(name);"
+      } else {
+        "L\(name);"
+      }
+    case .array(let javaType): "[\(javaType.jniTypeSignature)"
+    case .void: fatalError("There is no type signature for 'void'")
+    }
+  }
+}
--- a/Sources/JavaTypes/JavaType+JNI.swift
+++ b/Sources/JavaTypes/JavaType+JNI.swift
@@ -12,12 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if canImport(FoundationEssentials)
-import FoundationEssentials
-#else
-import Foundation
-#endif
-
 extension JavaType {
   /// Map this Java type to the appropriate JNI type name.
   package var jniTypeName: String {
@@ -61,27 +55,6 @@ extension JavaType {
     case .double: "d"
     case .class, .array: "l"
     case .void: fatalError("There is no field name for 'void'")
-    }
-  }
-
-  package var jniTypeSignature: String {
-    switch self {
-    case .boolean: "Z"
-    case .byte: "B"
-    case .char: "C"
-    case .short: "S"
-    case .int: "I"
-    case .long: "J"
-    case .float: "F"
-    case .double: "D"
-    case .class(let package, let name):
-      if let package {
-        "L\(package.replacingOccurrences(of: ".", with: "/"))/\(name);"
-      } else {
-        "L\(name);"
-      }
-    case .array(let javaType): "[\(javaType.jniTypeSignature)"
-    case .void: fatalError("There is no type signature for 'void'")
     }
   }
 }


ktoso:
Nice, and here's the issue to track the swiftkit split-up https://github.com/swiftlang/swift-java/issues/299

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -0,0 +1,180 @@
...
+      let filename = "\(ty.swiftNominal.name).java"
+      logger.info("Printing contents: \(filename)")
+      printImportedNominal(&printer, ty)
+
+      if let outputFile = try printer.writeContents(
+        outputDirectory: javaOutputDirectory,
+        javaPackagePath: javaPackagePath,
+        filename: filename
+      ) {
+        print("[swift-java] Generated: \(ty.swiftNominal.name.bold).java (at \(outputFile))")
```
ktoso: this can use the `logger`
```suggestion
        logger.info("[swift-java] Generated: \(ty.swiftNominal.name.bold).java (at \(outputFile))")
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -0,0 +1,180 @@
...
+      printer.print(
+        """
+        long selfPointer = \(type.qualifiedName).allocatingInit(\(initArguments.joined(separator: ", ")));
+        return new \(type.qualifiedName)(selfPointer);
+        """
+      )
+    }
+
+    let parameters = translatedDecl.translatedFunctionSignature.parameters.map(\.asParameter)
+    printer.print("private static native long allocatingInit(\(parameters.joined(separator: ", ")));")
```
ktoso, line 169: ok name :)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -0,0 +1,133 @@
...
+    let name: String
+
+    /// The name of the parent scope this function is declared in (or nil if global)
+    let parentName: String?
+
+    /// Function signature
+    let translatedFunctionSignature: TranslatedFunctionSignature
+  }
+
+  struct JavaParameter {
```
ktoso: we have this type already, can we reuse the one from `FFMSwift2JavaGenerator+JavaTranslation.swift`? You can just move it out from the FFM file into some common place

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -0,0 +1,133 @@
...
+    let name: String
+    let type: JavaType
+
+    var asParameter: String {
+      "\(type) \(name)"
+    }
+
+    var asArgument: String {
```
ktoso: hm a bit unsure about the "asArgument" argument name would come from wherever we created the value we pass to this instead 🤔

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -0,0 +1,270 @@
...
+      let downcallArguments = renderDowncallArguments(
+        swiftFunctionSignature: decl.functionSignature,
+        translatedFunctionSignature: translatedDecl.translatedFunctionSignature
+      )
+      // TODO: Throwing initializers
+      printer.print(
+        """
+        let selfPointer = UnsafeMutablePointer<\(typeName)>.allocate(capacity: 1)
+        selfPointer.initialize(to: \(typeName)(\(downcallArguments)))
+        return Int64(Int(bitPattern: selfPointer)).getJNIValue(in: environment)
```
ktoso, line 122: Yeah the int64 here is okey I think, and the generated signatures all just work on `long` which makes sense... If running on 32bit I think we'd likely do the same anyway, and just check at runtime values make sense maybe. Having that said, I don't think we really have a 32bit requirement right

ktoso:
Very nice so far! :-)

Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2SwiftJNI.java
```diff
@@ -40,6 +40,11 @@ static void examples() {
         long i = MySwiftLibrary.globalMakeInt();
         SwiftKit.trace("globalMakeInt() = " + i);
 
+        MySwiftClass.method();
+
+        MySwiftClass myClass = MySwiftClass.init(10, 5);
+        MySwiftClass myClass2 = MySwiftClass.init();
```
ktoso, line 46: I was saying in our sync that either tests or the main app is ok but I'm actually wondering now if it won't be better to move more into tests...

How about we do a test target and add all these "can create a class" etc as tests. We'd keep the executable as well but make it something more "look, it works!" rather than have to cover all the various cases...

We need to do the same refactoring in the other samples I guess.

And we'd update the ci-validate script to do both `swift run` and `swift test`

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
madsodgaard: I have a todo about thinking about 32-bit later! but I guess some Android devices can run 32-bit, so we probably have to think about it.

Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2SwiftJNI.java
madsodgaard: agreed. I'll add tests for this sample, for now.

PR #300, opened 2025-07-03
Separate SwiftKit into SwiftKitCore and SwiftKitFFM

madsodgaard:
The goal of this PR is to separate SwiftKit into two modules:
* `SwiftKitCore`: A module that supports non-FFM memory management and uses JNI for any downcalls.
* `SwiftKitFFM`: A module that uses FFM for allocation of memory and downcalls.


BuildLogic/src/main/kotlin/build-logic.java-common-conventions.gradle.kts
```diff
@@ -33,9 +33,9 @@ repositories {
 
 testing {
     suites {
-        val test by getting(JvmTestSuite::class) {
-            useJUnitJupiter("5.10.3")
-        }
+//        val test by getting(JvmTestSuite::class) {
+//            useJUnitJupiter("5.10.3")
+//        }
```
ktoso: was this necessary?

Samples/SwiftAndJavaJarSampleLib/src/jmh/java/org/swift/swiftkit/JavaToSwiftBenchmark.java
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-package org.swift.swiftkit;
+package org.swift.swiftkitffm;
```
ktoso:
```suggestion
package org.swift.swiftkit.ffm;
```

I'd do a sub package like this; and everything we do in this module should be contained in this package then

SwiftKitFFM/build.gradle
```diff
@@ -16,7 +16,7 @@ plugins {
     id("build-logic.java-application-conventions")
 }
 
-group = "org.swift.swiftkit"
+group = "org.swift.swiftkitffm"
```
ktoso:
```suggestion
group = "org.swift.swiftkit.ffm"
```

let's do subpackage, that's "typical"

SwiftKitFFM/src/main/java/org/swift/swiftkitffm/AllocatingSwiftArena.java
```diff
@@ -0,0 +1,21 @@
+package org.swift.swiftkitffm;
```
ktoso: Just a heads up to remember the license headers

SwiftKitFFM/src/main/java/org/swift/swiftkitffm/AllocatingSwiftArena.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.swift.swiftkitcore.ClosableSwiftArena;
+import org.swift.swiftkitcore.ConfinedSwiftMemorySession;
+import org.swift.swiftkitcore.SwiftArena;
+
+import java.lang.foreign.MemorySegment;
+import java.util.concurrent.ThreadFactory;
+
+public interface AllocatingSwiftArena extends SwiftArena {
```
ktoso: Very nice, I think this makes sense; and the factories on the interface as well

SwiftKitCore/src/main/java/org/swift/swiftkitcore/JNISwiftInstance.java
```diff
@@ -0,0 +1,28 @@
...
+     *
+     * @param pointer a pointer to the memory containing the value
+     * @param arena   the arena this object belongs to. When the arena goes out of scope, this value is destroyed.
+     */
+    protected JNISwiftInstance(long pointer, SwiftArena arena) {
+        super(pointer, arena);
+    }
+
+    @Override
+    public SwiftInstanceCleanup makeCleanupAction() {
```
ktoso: Java style would be `create` or `get`
```suggestion
    public SwiftInstanceCleanup createCleanupAction() {
```

SwiftKitCore/src/main/java/org/swift/swiftkitcore/AutoSwiftMemorySession.java
```diff
@@ -0,0 +1,84 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package org.swift.swiftkitcore;
```
ktoso: Package wise I think we should do 

```suggestion
package org.swift.swiftkit.core;
```

or we could consider just `package org.swift.swiftkit;` for the core package... but let's go with very explicit for now.

The only thing to watch out for that all classses in a module stay in the same package, this is to make the java module system happy

SwiftKitCore/src/main/java/org/swift/swiftkitcore/JNISwiftInstance.java
```diff
@@ -0,0 +1,28 @@
+package org.swift.swiftkitcore;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class JNISwiftInstance extends SwiftInstance {
```
ktoso: That works out nicely I think 👍

SwiftKitCore/src/main/java/org/swift/swiftkitcore/JNISwiftInstanceCleanup.java
```diff
@@ -0,0 +1,28 @@
...
+        this.markAsDestroyed = markAsDestroyed;
+    }
+
+    @Override
+    public void run() {
+        markAsDestroyed.run();
+
+        try {
+            // Use reflection to look for the static destroy method on the wrapper.
+            Method method = clazz.getDeclaredMethod("destroy", long.class);
```
ktoso: We can pass a Function rather than `Class` to the constructor and store that; We shouldn't need to resort to reflection here AFAICS.

SwiftKitCore/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+plugins {
+    id("build-logic.java-application-conventions")
+}
+
+group = "org.swift.swiftkitcore"
```
ktoso: So the group for them all is `org.swift.swiftkit`

and the artif

```suggestion
group = "org.swift.swiftkit"
```

but the artifact id will come from the module name, which will be `SwiftKitCore` and `SwiftKitFFM` -> `swiftkit-core` + `swiftkit-ffm`

Setting those is done in `publishing { publications { maven(MavenPublication)` if I read this correctly... Either way -- the group ID remains the same for all those and you don't need to worry about the artifact IDs yet

SwiftKitFFM/src/main/java/org/swift/swiftkitffm/FFMSwiftInstanceCleanup.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Override
+    public void run() {
+        markAsDestroyed.run();
+
+        // Allow null pointers just for AutoArena tests.
+        if (selfType != null && selfPointer != null) {
+            System.out.println("[debug] Destroy swift value [" + selfType.getSwiftName() + "]: " + selfPointer);
+            SwiftValueWitnessTable.destroy(selfType, selfPointer);
+        }
```
ktoso: looks good

SwiftKitFFM/src/main/java/org/swift/swiftkitffm/SwiftFFM.java
```diff
@@ -12,28 +12,23 @@
...
 import java.util.Optional;
 import java.util.stream.Collectors;
 
-import static org.swift.swiftkit.util.StringUtils.stripPrefix;
-import static org.swift.swiftkit.util.StringUtils.stripSuffix;
+import static org.swift.swiftkitffm.util.StringUtils.stripPrefix;
+import static org.swift.swiftkitffm.util.StringUtils.stripSuffix;
 
-public class SwiftKit {
+public class SwiftFFM {
```
ktoso: I think we can call this type `SwiftRuntime` now as it is the runtime entrypoints we call using ffm... No need to call it FFM in the type -- it is in the ffm module (or rather `.foreign` module)

SwiftKitCore/src/main/java/org/swift/swiftkitcore/SwiftLibraries.java
```diff
@@ -0,0 +1,79 @@
...
+import java.nio.file.StandardCopyOption;
+
+public class SwiftLibraries {
+
+    public static final String STDLIB_DYLIB_NAME = "swiftCore";
+    public static final String SWIFTKITSWIFT_DYLIB_NAME = "SwiftKitSwift";
+    public static final boolean TRACE_DOWNCALLS = Boolean.getBoolean("jextract.trace.downcalls");
+
+    private static final String STDLIB_MACOS_DYLIB_PATH = "/usr/lib/swift/libswiftCore.dylib";
+
```
ktoso: Very nice, that's a nice centralization of this logic.

Let's make the class `final`

ktoso:
> But the vision I have is that SwiftInstance now provides a destroy method that the JNI code can then override to call into the VWT from Swift and deallocate the pointer, while FFM can still use the existing VWT calling.

I think that's the right call, but let's avoid getting the method via reflection and instead just give the cleanup something to run, rather that reflection trick and it'll be fine 👍

.github/workflows/pull_request.yml
```diff
@@ -31,10 +31,14 @@ jobs:
...
-      - name: Gradle :SwiftKit:check
-        run: ./gradlew :SwiftKit:check --info
+      - name: Gradle :SwiftKitCore:build
+        run: ./gradlew :SwiftKitCore:build -x test
+      - name: Gradle :SwiftKitCore:check
+        run: ./gradlew :SwiftKitCore:check --info
+      - name: Gradle :SwiftKitFFM:build
+        run: ./gradlew :SwiftKitFFM:build -x test
+      - name: Gradle :SwiftKitFFM:check
+        run: ./gradlew :SwiftKitFFM:check --info
```
ktoso, line 41: yeah that's nice actually to keep them separate rather than one big `:check`

Samples/SwiftAndJavaJarSampleLib/ci-validate.sh
```diff
@@ -27,9 +27,10 @@ fi
 
 # check if we can compile a plain Example file that uses the generated Java bindings that should be in the generated jar
 # The classpath MUST end with a * if it contains jar files, and must not if it directly contains class files.
-SWIFTKIT_CLASSPATH="$(pwd)/../../SwiftKit/build/libs/*"
+SWIFTKIT_CORE_CLASSPATH="$(pwd)/../../SwiftKitCore/build/libs/*"
+SWIFTKIT_FFM_CLASSPATH="$(pwd)/../../SwiftKitFFM/build/libs/*"
```
ktoso, line 31: Yup, sorry you had to go through figuring that out, but now you know hah :)

Samples/SwiftKitSampleApp/build.gradle
```diff
@@ -148,7 +147,8 @@ tasks.clean {
 }
 
 dependencies {
-    implementation(project(':SwiftKit'))
+    implementation(project(':SwiftKitCore'))
+    implementation(project(':SwiftKitFFM'))
```
ktoso, line 151: Nitpick, no need to fix up right now -- since FFM depends on Core we could just list the FFM dependency here, it'd resolve properly I think even with project dependencies

SwiftKitCore/src/main/java/org/swift/swiftkit/core/JNISwiftInstanceCleanup.java
```diff
@@ -0,0 +1,31 @@
...
+    private final Runnable markAsDestroyed;
+
+    public JNISwiftInstanceCleanup(Runnable destroyFunction, Runnable markAsDestroyed) {
+        this.destroyFunction = destroyFunction;
+        this.markAsDestroyed = markAsDestroyed;
+    }
+
+    @Override
+    public void run() {
+        markAsDestroyed.run();
```
ktoso, line 28: We can do this in a follow up but we should only run the destroy if the mark permits us to do so; we can fix that later though

PR #301, opened 2025-07-07
[JExtract] Import `Foundation.Data` if used

rintaro:
* Introduce `KnownSwiftModules` to represent `Swift` and `Foundation` modules.
* `SwiftSymbolTable.setup()` is now a factory method to create `SwiftSymbolTable`. It scans the source file for module dependency. If `import Foundation` is found, add `Foundation` from the `KnownSwiftModules`  (Just `Data` for now)
* In `Swift2JavaTranslator` import `Data` APIs if any of the imported APIs use `Data`
* `Data` has limited hardcoded APIs `init(bytes:count:)`, `getter:count`, and `withUnsafeBytes(_:)`.
* Add `import Foundation` in Swift thunk if any of the the original source files import it


Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -100,6 +100,13 @@ struct JExtractSwiftBuildToolPlugin: SwiftJavaPluginProtocol, BuildToolPlugin {
       outputSwiftDirectory.appending(path: "\(sourceModule.name)Module+SwiftJava.swift")
     ]
 
+    // If the module uses 'Data' type, the thunk file is emitted as if 'Data' is declared
+    // in that module. Declare the thunk file as the output.
+    // FIXME: Make this conditional.
+    outputSwiftFiles += [
+      outputSwiftDirectory.appending(path: "Data+SwiftJava.swift")
+    ]
```
ktoso, line 108: That FIXME will be hard to resolve... we'd have to grep files manually for any foundation imports etc I guess before we run any of the commands :(

But yeah, good to keep that fixme

Samples/SwiftKitSampleApp/src/main/java/com/example/swift/HelloJava2Swift.java
```diff
@@ -78,6 +82,20 @@ static void examples() {
...
+            var origBytes = arena.allocateFrom("foobar");
+            var origDat = Data.init(origBytes, origBytes.byteSize(), arena);
+            SwiftRuntime.trace("origDat.count = " + origDat.getCount());
+            
+            var retDat = MySwiftLibrary.globalReceiveReturnData(origDat, arena);
+            retDat.withUnsafeBytes((retBytes) -> {
+                var str = retBytes.getString(0);
+                SwiftRuntime.trace("retStr=" + str);
+            });
+        }
```
ktoso, line 96: Very nice! 

btw I'm thinking we should start trying to move all these examples into tests, so we'll get a better report "what failed" when CI fails.

But this we can do later separately

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -65,6 +63,9 @@ package class FFMSwift2JavaGenerator: Swift2JavaGenerator {
         return String(filePathPart.replacing(".swift", with: "+SwiftJava.swift"))
       })
       self.expectedOutputSwiftFiles.insert("\(translator.swiftModuleName)Module+SwiftJava.swift")
+
+      // FIXME: Can we avoid this?
+      self.expectedOutputSwiftFiles.insert("Data+SwiftJava.swift")
```
ktoso, line 68: Yeah, this is unavoidable in today's SwiftPM.

Can we call the file `Foundation_Data+SwiftJava.swift` maybe? just to lessen the chance of clashing with some user defined Data type?

FYI ongoing sadness with build plugins @bnbarham heh

Sources/JExtractSwiftLib/Swift2JavaTranslator.swift
```diff
@@ -108,11 +99,69 @@ extension Swift2JavaTranslator {
       log.trace("Analyzing \(input.filePath)")
       visitor.visit(sourceFile: input.syntax)
     }
+
+    // If any API uses 'Foundation.Data', import 'Data' as if it's declared in
+    // this module.
+    if let dataDecl = self.symbolTable[.data] {
+      if self.isUsing(dataDecl) {
+        visitor.visit(nominalDecl: dataDecl.syntax!.asNominal!, in: nil)
+      }
```
ktoso, line 108: 👍 Good enough for our current needs.

Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownModules.swift
```diff
@@ -0,0 +1,94 @@
...
+  public struct UnsafeMutableBufferPointer<Element> {}
+  
+  // FIXME: Support 'typealias Void = ()'
+  public struct Void {}
+  
+  public struct String {
+    public init(cString: UnsafePointer<Int8>)
+    public func withCString(_ body: (UnsafePointer<Int8>) -> Void)
+  }
+  """
```
ktoso, line 86: Makes sense but I'm a bit worried about those "inline" pseudo-interface files... While we can't use the real files... maybe we could make this info at least real files and not just strings hardcoded like this, put them in `SwiftTypes/KnownModules/Foundation.swiftinterface` and then use them using the `Bundle` APIs? Seems that could work?

Sources/JExtractSwiftLib/SwiftTypes/SwiftParsedModuleSymbolTableBuilder.swift
```diff
@@ -0,0 +1,175 @@
...
+  /// Add a nominal type declaration and all of the nested types within it to the symbol
+  /// table.
+  mutating func handle(
+    nominalTypeDecl node: NominalTypeDeclSyntaxNode,
+    parent: SwiftNominalTypeDeclaration?
+  ) {
+    // If we have already recorded a nominal type with the name in this module,
+    // it's an invalid redeclaration.
+    if let _ = symbolTable.lookupType(node.name.text, parent: parent) {
+      // TODO: Diagnose?
```
ktoso: might be good to at least log on debug here

ktoso (approved):
Very cool, lgtm -- some minor things inline

Sources/JExtractSwiftLib/Swift2JavaTranslator.swift
```diff
@@ -108,11 +99,69 @@ extension Swift2JavaTranslator {
...
+      return true
+    }
+    for importedType in self.importedTypes.values {
+      if importedType.initializers.contains(where: check) {
+        return true
+      }
+      if importedType.methods.contains(where: check) {
+        return true
+      }
+      if importedGlobalVariables.contains(where: check) {
```
rintaro:
```suggestion
      if importedType.variables.contains(where: check) {
```

Sources/JExtractSwiftLib/SwiftTypes/SwiftParsedModuleSymbolTableBuilder.swift
rintaro: Oh yeah, I forgot to hook up the `Logger` passed in to `SwiftSymbolTable.setup()`

ktoso: That’s be nice to do probably, thank you

Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownModules.swift
rintaro: I will try making them resources in followups

PR #302, opened 2025-07-10
[JExtract] Generate JNI code for memory management with `SwiftKitCore`

madsodgaard:
Adds SwiftKitCore support for the generated code for classes. This means that we will actually destroy the instances once the arena decides to so.

### DIFF
--- a/Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2SwiftJNI.java
+++ b/Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2SwiftJNI.java
@@ -19,6 +19,7 @@ package com.example.swift;
 // Import javakit/swiftkit support libraries
 
 import org.swift.swiftkit.core.SwiftLibraries;
+import org.swift.swiftkit.core.ConfinedSwiftMemorySession;
 
 public class HelloJava2SwiftJNI {
 
@@ -40,8 +41,10 @@ public class HelloJava2SwiftJNI {
 
         MySwiftClass.method();
 
-        MySwiftClass myClass = MySwiftClass.init(10, 5);
-        MySwiftClass myClass2 = MySwiftClass.init();
+        try (var arena = new ConfinedSwiftMemorySession(Thread.currentThread())) {
+            MySwiftClass myClass = MySwiftClass.init(10, 5, arena);
+            MySwiftClass myClass2 = MySwiftClass.init(arena);
+        }
 
         System.out.println("DONE.");
     }
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -12,6 +12,19 @@
 //
 //===----------------------------------------------------------------------===//
 
+
+// MARK: Defaults
+
+extension JNISwift2JavaGenerator {
+  /// Default set Java imports for every generated file
+  static let defaultJavaImports: Array<String> = [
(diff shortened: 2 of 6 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -266,3 +291,29 @@ extension JNISwift2JavaGenerator {
     .joined(separator: ", ")
   }
 }
+
+extension String {
+  /// Returns a version of the string correctly escaped for a JNI
+  var nativeJNIEscaped: String {
```
ktoso: Hmm, `escapedJNIIdentifier` maybe? It's not any random string that is getting escaped this way

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -182,14 +201,39 @@ extension JNISwift2JavaGenerator {
...
+        """
+        long $selfPointer = this.pointer(); 
+        return new Runnable() {
+          @Override
+          public void run() {
+            \(type.swiftNominal.name).$destroy($selfPointer);
+          }
+        };
+        """
+      )
```
ktoso, line 220: Very nice, and nice how simple it is to do the pair of JNI thunk and the native method here :-)

Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2SwiftJNI.java
```diff
@@ -40,8 +41,10 @@ static void examples() {
 
         MySwiftClass.method();
 
-        MySwiftClass myClass = MySwiftClass.init(10, 5);
-        MySwiftClass myClass2 = MySwiftClass.init();
+        try (var arena = new ConfinedSwiftMemorySession(Thread.currentThread())) {
```
ktoso, line 44: Nitpick that this may want to be `SwiftArena.ofConfined()`

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -12,6 +12,19 @@
...
+
+// MARK: Defaults
+
+extension JNISwift2JavaGenerator {
+  /// Default set Java imports for every generated file
+  static let defaultJavaImports: Array<String> = [
+    "org.swift.swiftkit.core.*",
+    "org.swift.swiftkit.core.util.*",
+  ]
+}
```
ktoso, line 24: 👍

ktoso (approved):

Samples/JExtractJNISampleApp/src/main/java/com/example/swift/HelloJava2SwiftJNI.java
madsodgaard: agreed, but the problem is that Java 7 does not support static methods on interfaces.

ktoso: Ugh... heh I wonder if we need to really support so long back, we'll see as we chat with android workgroup.

Otherwise the Java pattern is to do a type like `SwiftArenas` that's a class and put the static methods on that

ktoso: At least we can make a `()` constructor that gets the current thread as well then, we can do this in a follow up

ktoso (approved):

PR #305, opened 2025-07-12
[JExtract] Add support for JNI member methods

madsodgaard:
Adds support for generating code for member methods in JNI mode. Also fixes a bug with throwing functions, that also affected the JavaKit macros.

```swift
public func doSomething(x: Int64) {}
```

is generated as:

### Java
```java
 public void doSomething(long x) {
  long selfPointer = this.pointer();
  MyClass.$doSomething(x, selfPointer);
}

private static native void $doSomething(long x, long selfPointer);
```

### Swift
```swift
@_cdecl("Java_com_example_swift_MyClass__00024doSomething__JJ")
func Java_com_example_swift_MyClass__00024doSomething__JJ(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, x: jlong, selfPointer: jlong) {
  let self$ = UnsafeMutablePointer<MyClass>(bitPattern: Int(Int64(fromJNI: selfPointer, in: environment!)))!
  self$.pointee.doSomething(x: Int64(fromJNI: x, in: environment!))
}
```

### DIFF
--- a/Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
+++ b/Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
@@ -35,4 +35,30 @@ public class MySwiftClassTest {
             assertNotNull(c);
         }
     }
+
+    @Test
+    void sum() {
+        try (var arena = new ConfinedSwiftMemorySession(Thread.currentThread())) {
+            MySwiftClass c = MySwiftClass.init(20, 10, arena);
+            assertEquals(30, c.sum());
+        }
+    }
+
+    @Test
+    void xMultiplied() {
+        try (var arena = new ConfinedSwiftMemorySession(Thread.currentThread())) {
+            MySwiftClass c = MySwiftClass.init(20, 10, arena);
+            assertEquals(200, c.xMultiplied(10));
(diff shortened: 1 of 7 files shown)


Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
```diff
@@ -35,4 +35,30 @@ void init_withParameters() {
             assertNotNull(c);
         }
     }
+
+    @Test
+    void sum() {
+        try (var arena = new ConfinedSwiftMemorySession(Thread.currentThread())) {
```
ktoso, line 41: Can you please add a constructor that takes no arguments and uses the current thread? There's no good reason to keep propagating this shape of end user passing in the thread like we do here

ktoso (approved):
Very good! just the nitpick about the passing in the thread. Let's fix that separately maybe :-)

Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
madsodgaard: yes!

PR #306, opened 2025-07-12
[JExtract/JNI] Add support for variables

madsodgaard:
Adds support for both free and member variables. Also supports computed properties, throwing getters and getters/setters.

Boolean getters are now also generated as `isX` both for FFM and JNI to match Java conventions.

### DIFF
--- a/Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
+++ b/Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
@@ -111,5 +111,11 @@ public class MySwiftClassTest {
         }
     }
 
-
+    @Test
+    void isWarm() {
+        try (var arena = new ConfinedSwiftMemorySession()) {
+            MySwiftClass c = MySwiftClass.init(20, 10, arena);
+            assertFalse(c.isWarm());
+        }
+    }
 }


Sources/JExtractSwiftLib/Convenience/String+Extensions.swift
```diff
@@ -22,4 +22,14 @@ extension String {
 
     return "\(f.uppercased())\(String(dropFirst()))"
   }
-}
\ No newline at end of file
+
+  /// Returns whether the string is of the format `isX`
+  var hasJavaBooleanNamingConvention: Bool {
```
ktoso, line 27: Yup, nice

Sources/JExtractSwiftLib/Convenience/String+Extensions.swift
```diff
@@ -22,4 +22,14 @@ extension String {
...
+    guard self.hasPrefix("is"), self.count > 2 else {
```
ktoso, line 28: Maybe we need to check that the third character is uppercase as well. Otherwise we may be seeing a “iss()” which isn’t a setter but just a funny named method

Sources/JExtractSwiftLib/ImportedDecls.swift
```diff
@@ -154,3 +154,27 @@ extension ImportedFunc: Hashable {
...
+    }
+  }
+
+  var javaSetterName: String {
+    let firstParameterIsBoolean = self.functionSignature.parameters.first?.type.asNominalTypeDeclaration?.knownTypeKind == .bool
+
+    if !firstParameterIsBoolean || self.name.hasJavaBooleanNamingConvention {
+      return "set\(self.name.toCamelCase)"
+    } else {
+      return "setIs\(self.name.toCamelCase)"
```
ktoso: The JavaBean naming convention is still just “setThing” in this case https://docstore.mik.ua/orelly/java-ent/jnut/ch06_02.htm

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -192,19 +202,32 @@ extension JNISwift2JavaGenerator {
...
+        translatedFunctionSignature: translatedDecl.translatedFunctionSignature
+      )
+      result = "\(tryClause)\(calleeName).\(decl.name)(\(downcallParameters))"
+
+    case .getter:
+      result = "\(tryClause)\(calleeName).\(decl.name)"
+
+    case .setter:
+      assert(decl.functionSignature.parameters.count == 1)
+      let newValueParameter = decl.functionSignature.parameters[0]
```
ktoso: Generally avoid indexing with integers, maybe rephrase this in a `guard let … = ….parameters.first else { … crash here }`

SwiftKitCore/src/main/java/org/swift/swiftkit/core/ConfinedSwiftMemorySession.java
```diff
@@ -28,6 +28,10 @@ public class ConfinedSwiftMemorySession implements ClosableSwiftArena {
 
     final ConfinedResourceList resources;
 
+    public ConfinedSwiftMemorySession() {
+        this(Thread.currentThread());
+    }
```
ktoso, line 33: Cool thanks

Tests/JExtractSwiftTests/JNI/JNIVariablesTests.swift
```diff
@@ -0,0 +1,386 @@
...
+      return 0
+    }
+    public var computedThrowing: Int64 {
+      get throws { return 0 }
+    }
+    public var getterAndSetter: Int64 {
+      get { return 0 }
+      set { }
+    }
+      public var someBoolean: Bool
```
ktoso, line 35: I’m on a phone but it seems this formatting is a bit off?

ktoso: I wonder if we should add an “get async” here and make sure we properly DONT import it, rather than try to import and blow up in compile or generated code?

ktoso (approved):
Very nice, had some comments on this one. Please have a look

Tests/JExtractSwiftTests/JNI/JNIVariablesTests.swift
madsodgaard: so something like `assertNoOutput` that just checks whether the Java code does not include a `getAsyncVariable` name?

ktoso: Maybe that’s the easiest way to test yeah hm…

ktoso: We decided to just rely on the sample compiling to detect if we generated some "bad code", so let's just have some async funcs in our sample input swift files and that'll be enough :)

Samples/JExtractJNISampleApp/src/test/java/com/example/swift/MySwiftClassTest.java
```diff
@@ -22,43 +22,94 @@
...
+    void mutableDividedByTwo() {
+        try (var arena = new ConfinedSwiftMemorySession()) {
+            MySwiftClass c = MySwiftClass.init(20, 10, arena);
+            assertEquals(0, c.getMutableDividedByTwo());
+            c.setMutable(20);
+            assertEquals(10, c.getMutableDividedByTwo());
+            c.setMutableDividedByTwo(5);
+            assertEquals(10, c.getMutable());
+        }
+    }
```
ktoso, line 112: We're missing the boolean property in a test here, can we add it?

madsodgaard: added

PR #310, opened 2025-07-13
[WIP] Android support 

colemancda:
WIP branch with changes that enable use of JavaKit on Android.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,5 @@
-// swift-tools-version: 6.0
+// swift-tools-version: 6.1
 // The swift-tools-version declares the minimum version of Swift required to build this package.
-
 import CompilerPluginSupport
 import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -226,7 +226,7 @@ let package = Package(
       exclude: ["swift-java.config"],
       swiftSettings: [
         .swiftLanguageMode(.v5),
-        .unsafeFlags(["-I\(javaIncludePath)", "-I\(javaPlatformIncludePath)"])
+        .unsafeFlags(["-I\(javaIncludePath)", "-I\(javaPlatformIncludePath)"], .when(platforms: [.macOS, .linux, .windows]))
       ],
       linkerSettings: [
         .unsafeFlags(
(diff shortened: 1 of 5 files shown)


ktoso:
Do we intend this PR to supersede https://github.com/swiftlang/swift-java/pull/144 @colemancda @lhoward ?

lhoward:
I haven’t been following this one. Is it a superset of mine? If so, makes sense. If it’s an intersecting set then, I would propose reviewing mine and then rebasing this one after it is merged (presuming it is approved).

ktoso:
I'd like for us to catch up, it's a bit tricky for me to accept PRs without Android CI and without clear picture which ones we should be moving forward with and what is only PoC etc. Both are Draft so I understand them not really intended as being merged, but I also don't want to waste either of your efforts.

Will both of you in the android/java call we're scheduling with the android workgroup?

lhoward:
I’ll be. Sorry I only get to check in on this every month or so after an initial spurt of activity!

ktoso:
No problem at all :-) Just trying to figure out what everyone's plans here are 👍

lhoward:
It does that appear @colemancda has duplicated much of the work in the original PR, which I suppose is my fault for not following up on trying to get it merged earlier.

Comments on specific commits are below (I've avoided using commit hashes are they're liable to change).

* [JavaRuntime] Add Android shims: dlopen()'ing `libnativehelper.so` was in my original patch set, but was removed IIRC at @ktoso's suggestion. NDK versions 31 and above export the necessary symbols. I don't have a particularly strong position on this (if it works, it works), but arguably we shouldn't be using symbols that we can't link at build time.
* JavaKit: always use ambient javaEnvironment on Android: I approve this fix, as it's cherry-picked from my PR.
* [JavaKit] Add JNIEnvPointer for Android: this was included in my commit JavaKit: support for Android NDK JVM; no objection to it being a distinct commit, but suggest a more verbose commit message.
* [JavaKit] Fix JNINativeInterface_ for Android: as above.
* [JavaKit] Don't include JVM path for Android: as above.

The commit:

* Samples: add sample Android app

from my PR may be useful to include.

colemancda:
I'm moving the shim to AndroidKit since it seems the official Swift Android release will target newer SDKs.

ktoso:
We added Android CI so we're making steps here, @madsodgaard is working on various pieces of the android support.

Going to close this :)

Issue #307, opened 2025-07-12
[JExtract/JNI] Unable to map `UInt8` to Java


colemancda:
Source code at https://github.com/PureSwift/Bluetooth/pull/197

```
coleman@Alseys-MacBook-Pro Bluetooth % SWIFT_JAVA=1 swift build --target Bluetooth
[1/1] Planning build
Building for debugging...
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/DeclarationUUID.swift
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/UnitIdentifiers.swift
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/ServiceUUID.swift
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/DescriptorUUID.swift
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/CharacteristicUUID.swift
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/MemberUUID.swift
Generated /Users/coleman/Developer/Bluetooth/.build/plugins/outputs/bluetooth/Bluetooth/destination/GenerateBluetoothDefinitions/CompanyIdentifiers.swift
JExtractSwiftLib/JNISwift2JavaGenerator+JavaTranslation.swift:71: Fatal error: unsupported known type: uint8
[163/165] Generate Java wrappers for Swift types
```

colemancda:
@madsodgaard Can types that don't have a native Java counterpart just be omitted from the generated code? I'm more than happy to use `jextract` just as a way to pass around `Opaque` Swift types, I can write additional Java / JNI later if I really need that value.

madsodgaard:
@colemancda Yes, I'll make sure that we just omit any unsupported types for now, instead of crashing.

ktoso:
That's a good first step, to exclude them. 

[...]

PR #312, opened 2025-07-13
WIP: Prepare Unsigned integer types for SwiftKit

ktoso:
Using these we'll be able to use them to more safely express unsigned types, rather than just pretend it's all the same and fine when we import them into **signed** java types.

We'll also be able to offer extract mode that can just ignore this and extract as raw int/long etc, but these APIs may be safer


SwiftKitCore/src/main/java/org/swift/swiftkit/core/primitives/UnsignedByte.java
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    /**
+     * Accept a signed Java @{code int} value, and interpret it as-if it was an unsigned value.
+     * In other words, do not interpret the negative bit as "negative", but as part of the unsigned integers value.
+     *
+     * @param bits bit value to store in this unsigned integer
+     * @return unsigned integer representation of the passed in value
+     */
+    public static UnsignedByte representedByBitsOf(byte bits) {
```
ktoso, line 46: TODO: benchmark if it's worth it

ktoso:
WIP towards resolving https://github.com/swiftlang/swift-java/issues/307

ktoso:
This is WIP and I need to confirm how to attribute Guava so we can reuse some of the nicer conversion logic UnsignedLong -> floating points etc. We'll do a NOTICE.txt but need to check how to attribute in specific files.

PR #313, opened 2025-07-13
[JExtract/JNI] Omit unsupported types instead of crashing

madsodgaard:
Omits any unsupported types such as `UInt8` instead of crashing the generation.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -176,7 +176,7 @@ extension JNISwift2JavaGenerator {
   }
 
   private func printFunctionBinding(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
-    guard let translatedDecl = translatedDecl(for: decl) else {
+    guard let _ = translatedDecl(for: decl) else {
       // Failed to translate. Skip.
       return
     }
@@ -201,7 +201,7 @@ extension JNISwift2JavaGenerator {
   /// and passes it down to another native function along with the arguments
   /// to call the Swift implementation.
   private func printMemberMethodBindings(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
-    let translatedDecl = translatedDecl(for: decl)!
+    let translatedDecl = translatedDecl(for: decl)! // We will only call this method if we can translate the decl.
 
     printDeclDocumentation(&printer, decl)
     printer.printBraceBlock("public \(renderFunctionSignature(decl))") { printer in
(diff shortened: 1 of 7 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -275,7 +283,7 @@ extension JNISwift2JavaGenerator {
   /// `func method(x: Int, y: Int) -> Int` becomes
   /// `long method(long x, long y)`
   private func renderFunctionSignature(_ decl: ImportedFunc) -> String {
-    let translatedDecl = translatedDecl(for: decl)
+    let translatedDecl = translatedDecl(for: decl)!
```
ktoso: Are we sure about the !, can we either make a guard with a better message or document why the ! Should not happen

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,6 +176,11 @@ extension JNISwift2JavaGenerator {
   }
 
   private func printFunctionBinding(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
+    guard let translatedDecl = translatedDecl(for: decl) else {
+      // Failed to translate. Skip.
```
ktoso, line 180: Can we log on debug here what we missed to emit due to that

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -220,7 +225,10 @@ extension JNISwift2JavaGenerator {
   }
 
   private func printInitializerBindings(_ printer: inout CodePrinter, _ decl: ImportedFunc, type: ImportedNominalType) {
-    let translatedDecl = translatedDecl(for: decl)
+    guard let translatedDecl = translatedDecl(for: decl) else {
```
ktoso, line 228: Same here

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
ktoso: Aha we logged in the translation already here.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -17,13 +17,19 @@ import JavaTypes
...
     }
 
-    let translation = JavaTranslation(swiftModuleName: self.swiftModuleName)
-    let translated = translation.translate(decl)
+    let translated: TranslatedFunctionDecl?
+    do {
+      let translation = JavaTranslation(swiftModuleName: swiftModuleName)
+      translated = try translation.translate(decl)
+    } catch {
+      self.logger.info("Failed to translate: '\(decl.swiftDecl.qualifiedNameForDebug)'; \(error)")
```
ktoso: Lower to debug please, as this may be quite noisy

ktoso:
Minor log level nitpick.

Does the FFM mode already do the same? Can we add the same “test” (just a decl we ignore) in FFM?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
madsodgaard: yea, all these methods are called by the `printFunctionBinding` which checks if the decl can be translated. I just took a similar approach to FFM.

madsodgaard:
@ktoso Yes FFM should also ignore any unsupported types, but I also added a unsupported type to the FFM sample app.

PR #321, opened 2025-07-15
[SwiftKit] Move 'selfPointer' to 'JNISwiftInstance'

rintaro:
`FFMSwiftInstance` doesn't need it.


ktoso:
Hmm, let me change this slightly... we can keep the pointer in the base class and implement it as `return $memorySegment().address();` in FFM

SwiftKitCore/src/main/java/org/swift/swiftkit/core/JNISwiftInstance.java
```diff
@@ -17,14 +17,25 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public abstract class JNISwiftInstance extends SwiftInstance {
+    /// Pointer to the "self".
```
ktoso, line 20: nit: `/**` in java :)

ktoso:
This causes some initialization problems, I've solved in #322 and added a bunch of logging and cleanup since noticing this was pretty tricky without downcall traces

Issue #314, opened 2025-07-14
Investigate build instability LLBuildProgressTracker crashes


ktoso:
On CI we frequently get crashes like this:

```
Thread 4 crashed:

0                         0x00007f1559da359a swift_slowAlloc + 74 in [libswiftCore.so](http://libswiftcore.so/)
1 [ra]                    0x00007f1559da3877 swift_allocObject + 38 in [libswiftCore.so](http://libswiftcore.so/)
2 [ra]                    0x00007f1559b85571 _allocateStringStorage(codeUnitCapacity:) + 160 in [libswiftCore.so](http://libswiftcore.so/)
3 [ra]                    0x00007f1559d03452 _StringGuts.grow(_:) + 273 in [libswiftCore.so](http://libswiftcore.so/)
4 [ra]                    0x00007f1559bf7e88 specialized static String._fromUTF8Repairing(_:) + 999 in [libswiftCore.so](http://libswiftcore.so/)
5 [ra] [inlined] [system] 0x000056315a96e7ef specialized String.init<A, B>(decoding:as:) in swift-package at //<compiler-generated>
6 [ra] [inlined]          0x000056315a96e7ef stringFromData(_:) in swift-package at /home/build-user/llbuild/products/llbuildSwift/Internals.swift:49:12
7 [ra]                    0x000056315a96e7ef Command.name.getter + 78 in swift-package at /home/build-user/llbuild/products/llbuildSwift/BuildSystemBindings.swift:720:20
8 [ra]                    0x00005631593a61a2 closure #1 in LLBuildProgressTracker.commandProcessFinished(_:process:result:) + 33 in swift-package at /home/build-user/swiftpm/Sources/Build/LLBuildProgressTracker.swift:347:65
9 [ra] [thunk]            0x00005631592f6989 thunk for @escaping @callee_guaranteed @Sendable () -> () + 24 in swift-package at //<compiler-generated>
```

Rebuilding usually gets rid of the issue... We need to figure out what's going on here. I've already reached out to swiftpm team also for any ideas.

rdar://155752697

rconnell9:
https://github.com/swiftlang/swift-package-manager/pull/8998

ktoso:
Thank you @rconnell9 ! Hope we can get it into a build soon :-)

ktoso:
Confirmed this fixes the problem -- https://github.com/swiftlang/swift-java/pull/357

[...]

PR #322, opened 2025-07-16
[SwiftKit] Move 'selfPointer' to 'JNISwiftInstance' & downcall trace logs

ktoso:
Seems I could not push "to" the PR here https://github.com/swiftlang/swift-java/pull/321

Minor refactor, we can easily keep the pointer in the base type

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -206,13 +206,15 @@ extension JNISwift2JavaGenerator {
     printDeclDocumentation(&printer, decl)
     printer.printBraceBlock("public \(renderFunctionSignature(decl))") { printer in
       var arguments = translatedDecl.translatedFunctionSignature.parameters.map(\.name)
-      arguments.append("selfPointer")
+
+      let selfVarName = "self$"
+      arguments.append(selfVarName)
 
       let returnKeyword = translatedDecl.translatedFunctionSignature.resultType.isVoid ? "" : "return "
 
       printer.print(
         """
-        long self$ = this.$memoryAddress();
+        long \(selfVarName) = this.$memoryAddress();
         \(returnKeyword)\(translatedDecl.parentName).$\(translatedDecl.name)(\(arguments.joined(separator: ", ")));
         """
       )
@@ -267,18 +269,17 @@ extension JNISwift2JavaGenerator {
     printer.printBraceBlock("protected Runnable \(funcName)()") { printer in
       printer.print(
         """
+        long self$ = this.$memoryAddress();
         if (CallTraces.TRACE_DOWNCALLS) {
           CallTraces.traceDowncall("\(type.swiftNominal.name).\(funcName)",
               "this", this,
-              "this.$memoryAddress()", this.$memoryAddress());
+              "self", self$);
         }
-        long self$ = this.$memoryAddress();
-        assert(self$ > 0);
         return new Runnable() {
           @Override
           public void run() {
             if (CallTraces.TRACE_DOWNCALLS) {
-              CallTraces.traceDowncall("\(type.swiftNominal.name).$destroy(" + self$ + ")");
+              CallTraces.traceDowncall("\(type.swiftNominal.name).$destroy", "self", self$);
             }
             \(type.swiftNominal.name).$destroy(self$);
           }
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -235,8 +235,8 @@ extension JNISwift2JavaGenerator {
       let initArguments = translatedDecl.translatedFunctionSignature.parameters.map(\.name)
       printer.print(
         """
-        long selfPointer = \(type.qualifiedName).allocatingInit(\(initArguments.joined(separator: ", ")));
-        return new \(type.qualifiedName)(selfPointer, swiftArena$);
+        long self$ = \(type.qualifiedName).allocatingInit(\(initArguments.joined(separator: ", ")));
+        return new \(type.qualifiedName)(self$, swiftArena$);
         """
       )
     }
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -212,7 +212,7 @@ extension JNISwift2JavaGenerator {
 
       printer.print(
         """
-        long selfPointer = this.pointer();
+        long self$ = this.$memoryAddress();
         \(returnKeyword)\(translatedDecl.parentName).$\(translatedDecl.name)(\(arguments.joined(separator: ", ")));
         """
       )
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -262,15 +262,25 @@ extension JNISwift2JavaGenerator {
   private func printDestroyFunction(_ printer: inout CodePrinter, _ type: ImportedNominalType) {
     printer.print("private static native void $destroy(long selfPointer);")
 
+    let funcName = "$createDestroyFunction"
     printer.print("@Override")
-    printer.printBraceBlock("protected Runnable $createDestroyFunction()") { printer in
+    printer.printBraceBlock("protected Runnable \(funcName)()") { printer in
       printer.print(
         """
-        long $selfPointer = this.pointer(); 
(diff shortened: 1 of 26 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -206,13 +206,15 @@ extension JNISwift2JavaGenerator {
...
+
+      let selfVarName = "self$"
+      arguments.append(selfVarName)
 
       let returnKeyword = translatedDecl.translatedFunctionSignature.resultType.isVoid ? "" : "return "
 
       printer.print(
         """
-        long selfPointer = this.pointer();
+        long \(selfVarName) = this.$memoryAddress();
```
ktoso, line 217: These are basically unsafe, so I figured we shoudl hide them using our $ pattern.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -206,13 +206,15 @@ extension JNISwift2JavaGenerator {
     printDeclDocumentation(&printer, decl)
     printer.printBraceBlock("public \(renderFunctionSignature(decl))") { printer in
       var arguments = translatedDecl.translatedFunctionSignature.parameters.map(\.name)
-      arguments.append("selfPointer")
+
+      let selfVarName = "self$"
```
ktoso, line 210: The reason we're doing `$` in names in bodies is so they don't clash with any parameter name that someone can come up with, so I moved to self$ here -- FYI @madsodgaard

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -262,15 +264,24 @@ extension JNISwift2JavaGenerator {
...
+    printer.printBraceBlock("protected Runnable \(funcName)()") { printer in
       printer.print(
         """
-        long $selfPointer = this.pointer(); 
+        long self$ = this.$memoryAddress();
+        if (CallTraces.TRACE_DOWNCALLS) {
+          CallTraces.traceDowncall("\(type.swiftNominal.name).\(funcName)",
+              "this", this,
+              "self", self$);
+        }
```
ktoso, line 277: a bit of downtracing helped diagnose issues

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -320,28 +318,54 @@ extension JNISwift2JavaGenerator {
...
+        fatalError("Missing JNIEnv in downcall to \\(#function)")
+      }
+      assert(\(selfPointerParam.name) != 0, "\(selfPointerParam.name) memory address was null")
+      let selfBits$ = Int(Int64(fromJNI: \(selfPointerParam.name), in: env$))
+      assert(selfBits$ != 0, "$self memory address was null: \(selfPointerParam.name) = \\(\(selfPointerParam.name))" )
+      guard let \(newSelfParamName) = UnsafeMutablePointer<\(swiftParentName)>(bitPattern: selfBits$) else {
+        fatalError("Missing self pointer in call to \\(#function)!")
+      }
+      """
+    )
```
ktoso, line 363: This code was repeated 2 times, so now it's in 1 place.

The reason to split it up into more lines and not just the single line with multiple `!` is so we have better messages when things fail; It took me a while to diagnose an issue without this so I'd suggest to keep this in.

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftInstance.java
```diff
@@ -17,49 +17,38 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public abstract class SwiftInstance {
-    /// Pointer to the "self".
-    private final long selfPointer;
+
+    // TODO: make this a flagset integer and/or use a field updater
+    /** Used to track additional state of the underlying object, e.g. if it was explicitly destroyed. */
+    private final AtomicBoolean $state$destroyed = new AtomicBoolean(false);
```
ktoso, line 23: fields should be before methods; we should install a formatting tool tbh

SwiftKitCore/src/main/java/org/swift/swiftkit/core/JNISwiftInstance.java
```diff
@@ -42,7 +55,7 @@ protected JNISwiftInstance(long pointer, SwiftArena arena) {
     protected abstract Runnable $createDestroyFunction();
 
     @Override
-    public SwiftInstanceCleanup createCleanupAction() {
+    public SwiftInstanceCleanup $createCleanup() {
```
ktoso, line 58: Also not a method end users should be invoking explicitly, so changed it while at it.

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftInstance.java
```diff
@@ -17,49 +17,38 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public abstract class SwiftInstance {
-    /// Pointer to the "self".
-    private final long selfPointer;
```
ktoso, line 21: "The change" the `SwiftInstance` does not need storage, we can just have FFM or JNI one implement the abstract method instead

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftInstance.java
```diff
@@ -17,49 +17,38 @@
...
 
-    /**
-     * The designated constructor of any imported Swift types.
-     *
-     * @param pointer a pointer to the memory containing the value
-     * @param arena the arena this object belongs to. When the arena goes out of scope, this value is destroyed.
-     */
-    protected SwiftInstance(long pointer, SwiftArena arena) {
-        this.selfPointer = pointer;
-        arena.register(this);
```
ktoso, line 60: Necessary to remove this since otherwise we'd race the register with a non-initialized memory address value in the class calling `super()`

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftObjects.java
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package org.swift.swiftkit.core;
+
+/**
+ * Utility functions, similar to @{link java.util.Objects}
+ */
+public class SwiftObjects {
```
ktoso, line 20: Following `java.util.Objects` but with more helpers for us ;)

SwiftKitFFM/src/main/java/org/swift/swiftkit/ffm/FFMSwiftInstance.java
```diff
@@ -22,31 +22,39 @@
...
      * The pointer to the instance in memory. I.e. the {@code self} of the Swift object or value.
      */
     public final MemorySegment $memorySegment() {
         return this.memorySegment;
     }
 
+    @Override
+    public long $memoryAddress() {
+        return $memorySegment().address();
+    }
```
ktoso, line 48: :-)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
madsodgaard: good idea!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -320,28 +318,54 @@ extension JNISwift2JavaGenerator {
...
+    printer.print(
+      """
+      guard let env$ = environment else {
+        fatalError("Missing JNIEnv in downcall to \\(#function)")
+      }
+      assert(\(selfPointerParam.name) != 0, "\(selfPointerParam.name) memory address was null")
+      let selfBits$ = Int(Int64(fromJNI: \(selfPointerParam.name), in: env$))
+      assert(selfBits$ != 0, "$self memory address was null: \(selfPointerParam.name) = \\(\(selfPointerParam.name))" )
+      guard let \(newSelfParamName) = UnsafeMutablePointer<\(swiftParentName)>(bitPattern: selfBits$) else {
+        fatalError("Missing self pointer in call to \\(#function)!")
```
madsodgaard: I think we can move the above assert message to this fatalError instead. The `UnsafeMutablePointer` [initializer](https://developer.apple.com/documentation/swift/unsafemutablepointer/init(bitpattern:)-84bz2?changes=_8) will fail if the bitPattern is zero.

madsodgaard (commented):
Very nice 👍

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
ktoso: Good point, done

ktoso:
Continued spm instability on linux :( 

https://github.com/swiftlang/swift-java/issues/314

PR #324, opened 2025-07-17
Build: simple local maven publishing

ktoso:


ktoso:
We can see if we can make it less repetetive later on

PR #325, opened 2025-07-18
[JExtract/JNI] Fix boolean naming for variables already called `isX`

madsodgaard:
Fixes a bug where variables named `isX` was imported as `IsX` instead.

### DIFF
--- a/Tests/JExtractSwiftTests/JNI/JNIVariablesTests.swift
+++ b/Tests/JExtractSwiftTests/JNI/JNIVariablesTests.swift
@@ -411,7 +411,7 @@ struct JNIVariablesTests {
   }
 
   @Test
-  func boolean_swiftThunks() throws {
+  func someBoolean_swiftThunks() throws {
     try assertOutput(
       input: membersSource,
       .jni,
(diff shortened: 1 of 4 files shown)


Tests/JExtractSwiftTests/JNI/JNIVariablesTests.swift
```diff
@@ -450,4 +450,58 @@ struct JNIVariablesTests {
...
+      detectChunkByInitialLines: 8,
+      expectedChunks: [
+      """
+      /**
+      * Downcall to Swift:
+      * {@snippet lang=swift :
+      * public let isBoolean: Bool
+      * }
+      */
+      public boolean isBoolean() {
```
ktoso, line 469: I wonder what we should do here... this isn't "correct" java beans getter naming technically speaking. Because the rule is _strictly_ `is<property name>()` which here _should be_ `isIsBoolean()` ([beans spec](https://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf?AuthParam=1753149426_47bfd76c8c160605a0beb22c4efa4386)).

So the problem is that we broke the symmetry between is<Property> and `set<property>` because the generated setter here would be `setIsBoolean()` which is breaking the naming convention 🤔 Some tools rely on the naming convention so I think we may want to follow it to the dot...

Specifically: the variable is "weirdly named" and thus we get not nice setter/getter: `isIsBoolean` (today we don't) and `setIsBoolean` (today we do).

Alternatively, we would have to fix the setter to become `setBoolean` (today we don't), so that it matches the `isBoolean` (today we do this).

I'd be okey with either but we need to match the convention I think, WDYT @madsodgaard ?

ktoso: I'd also recommend making this a `var` so we can test the setter's shape for this "weird case"

madsodgaard: We agreed to do `isBoolean` and `setBoolean`

PR #326, opened 2025-07-20
[JExtract/JNI] Add support for classes/structs as parameters and return values

madsodgaard:
Adds support for passing in classes/structs as parameters to functions and returning them as well. All functions now generate both a public "user-facing" Java method and a native method that is called from that. Any classes are passed as pointers and returned as pointers.

I also refactored quite a bit of the internal code to make it match the FFM style more, this should make it easier to understand and contribute to both modes.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -181,9 +181,9 @@ extension JNISwift2JavaGenerator {
       return
     }
 
-    var modifiers = "public"
+    var modifiers = ["public"]
     if decl.isStatic || decl.isInitializer || !decl.hasParent {
-      modifiers.append(" static")
+      modifiers.append("static")
     }
 
     let translatedSignature = translatedDecl.translatedFunctionSignature
@@ -194,9 +194,12 @@ extension JNISwift2JavaGenerator {
     }
     let throwsClause = decl.isThrowing ? " throws Exception" : ""
 
+    let modifiersStr = modifiers.joined(separator: " ")
+    let parametersStr = parameters.joined(separator: ", ")
+
     printDeclDocumentation(&printer, decl)
     printer.printBraceBlock(
-      "\(modifiers) \(resultType) \(translatedDecl.name)(\(parameters.joined(separator: ", ")))\(throwsClause)"
+      "\(modifiersStr) \(resultType) \(translatedDecl.name)(\(parametersStr))\(throwsClause)"
     ) { printer in
       printDowncall(&printer, decl)
     }
@@ -208,14 +211,13 @@ extension JNISwift2JavaGenerator {
     let translatedDecl = translatedDecl(for: decl)! // Will always call with valid decl
     let nativeSignature = translatedDecl.nativeFunctionSignature
     let resultType = nativeSignature.result.javaType
-    let nativeName = "$\(translatedDecl.name)"
     var parameters = nativeSignature.parameters
     if let selfParameter = nativeSignature.selfParameter {
       parameters.append(selfParameter)
     }
     let renderedParameters = parameters.map { "\($0.javaParameter.type) \($0.javaParameter.name)"}.joined(separator: ", ")
 
-    printer.print("private static native \(resultType) \(nativeName)(\(renderedParameters));")
+    printer.print("private static native \(resultType) \(translatedDecl.nativeFunctionName)(\(renderedParameters));")
   }
 
   private func printDowncall(
@@ -241,7 +243,7 @@ extension JNISwift2JavaGenerator {
     //=== Part 3: Downcall.
     // TODO: If we always generate a native method and a "public" method, we can actually choose our own thunk names
     // using the registry?
-    let downcall = "\(translatedDecl.parentName).$\(translatedDecl.name)(\(arguments.joined(separator: ", ")))"
+    let downcall = "\(translatedDecl.parentName).\(translatedDecl.nativeFunctionName)(\(arguments.joined(separator: ", ")))"
 
     //=== Part 4: Convert the return value.
     if translatedFunctionSignature.resultType.javaType.isVoid {
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -117,7 +117,7 @@ extension JNISwift2JavaGenerator {
       printer.println()
 
       for initializer in decl.initializers {
-        printInitializerBindings(&printer, initializer, type: decl)
+        printFunctionBinding(&printer, initializer)
         printer.println()
       }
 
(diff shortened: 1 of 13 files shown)


madsodgaard:
@rintaro As I refactored the code to match the way FFM does things, with `ConversionStep`s and so on, it would be nice if you could take a look at this PR as well 😄

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,75 +176,80 @@ extension JNISwift2JavaGenerator {
   }
 
   private func printFunctionBinding(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
-    guard let _ = translatedDecl(for: decl) else {
+    guard let translatedDecl = translatedDecl(for: decl) else {
       // Failed to translate. Skip.
       return
     }
 
+    var modifiers = "public"
```
ktoso: Might be better as a list

```suggestion
    var modifiers = ["public"]
```

so we don't have to do the weird `" "` in the append below

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,75 +176,80 @@ extension JNISwift2JavaGenerator {
...
     if decl.isStatic || decl.isInitializer || !decl.hasParent {
-      printStaticFunctionBinding(&printer, decl)
-    } else {
-      printMemberMethodBindings(&printer, decl)
+      modifiers.append(" static")
```
```suggestion
      modifiers.append("static")
```

once we make it an array

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,75 +176,80 @@ extension JNISwift2JavaGenerator {
...
+    let throwsClause = decl.isThrowing ? " throws Exception" : ""
 
     printDeclDocumentation(&printer, decl)
-    printer.printBraceBlock("public \(renderFunctionSignature(decl))") { printer in
-      var arguments = translatedDecl.translatedFunctionSignature.parameters.map(\.name)
-
-      let selfVarName = "self$"
-      arguments.append(selfVarName)
+    printer.printBraceBlock(
+      "\(modifiers) \(resultType) \(translatedDecl.name)(\(parameters.joined(separator: ", ")))\(throwsClause)"
```
ktoso: minor cleanup, and once we made modifiers an array 

```suggestion
      let modifiersStr = modifiers.joined(separator: " ")
      let parametersStr = parameters.joined(separator: ", ")
      "\(modifiersStr) \(resultType) \(translatedDecl.name)(\(parametersStr))\(throwsClause)"
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,75 +176,80 @@ extension JNISwift2JavaGenerator {
...
-        """
-        long \(selfVarName) = this.$memoryAddress();
-        \(returnKeyword)\(translatedDecl.parentName).$\(translatedDecl.name)(\(arguments.joined(separator: ", ")));
-        """
-      )
+  private func printNativeFunction(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
+    let translatedDecl = translatedDecl(for: decl)! // Will always call with valid decl
+    let nativeSignature = translatedDecl.nativeFunctionSignature
+    let resultType = nativeSignature.result.javaType
+    let nativeName = "$\(translatedDecl.name)"
```
ktoso: Since we have `translatedDecl.nativeFunctionSignature` would we also want to put `translatedDecl.nativeFunctionBaseName` maybe that does the `$` prefix

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,75 +176,80 @@ extension JNISwift2JavaGenerator {
...
+    if let selfParameter = nativeSignature.selfParameter {
+      parameters.append(selfParameter)
     }
+    let renderedParameters = parameters.map { "\($0.javaParameter.type) \($0.javaParameter.name)"}.joined(separator: ", ")
 
-    let returnType = translatedDecl.translatedFunctionSignature.resultType
-    var parameters = translatedDecl.translatedFunctionSignature.parameters.map(\.asParameter)
-    parameters.append("long selfPointer")
-    printer.print("private static native \(returnType) $\(translatedDecl.name)(\(parameters.joined(separator: ", ")));")
+    printer.print("private static native \(resultType) \(nativeName)(\(renderedParameters));")
```
ktoso: nice cleanup

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -176,75 +176,80 @@ extension JNISwift2JavaGenerator {
...
-        """
-        long self$ = \(type.qualifiedName).allocatingInit(\(initArguments.joined(separator: ", ")));
-        return new \(type.qualifiedName)(self$, swiftArena$);
-        """
-      )
+    // 'self' parameter.
+    if let selfParameter = translatedFunctionSignature.selfParameter {
+      let lowered = selfParameter.conversion.render(&printer, "this")
+      arguments.append(lowered)
     }
```
ktoso, line 241: nice cleanup!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -125,13 +164,100 @@ extension JNISwift2JavaGenerator {
...
+    let conversion: JavaNativeConversionStep
+  }
+
+  /// Represent a Swift API result translated to Java.
+  struct TranslatedResult {
+    let javaType: JavaType
+
+    /// Represents how to convert the Java native result into a user-facing result.
+    let conversion: JavaNativeConversionStep
+  }
```
ktoso, line 221: Hmmm, I wonder if we should strive to eventually share those types rather than dupe them 🤔 WDYT @rintaro

Sources/JExtractSwiftLib/JNI/JNIType.swift
```diff
@@ -0,0 +1,79 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import JavaTypes
+
+enum JNIType {
```
ktoso, line 20: Worth putting some docs on this one

```suggestion
/// Represents types that are able to be passed over a JNI boundary.
///
/// - SeeAlso: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/types.html
enum JNIType {
```

Tests/JExtractSwiftTests/Asserts/TextAssertions.swift
```diff
@@ -115,7 +115,7 @@ func assertOutput(
       print("==== ---------------------------------------------------------------")
       
       #expect(output.contains(expectedChunk), sourceLocation: sourceLocation)
-      fatalError("Failed: \(filePath):\(line)")
+//      fatalError("Failed: \(filePath):\(line)")
```
ktoso: Ah that's my fault for leaving it in during last PR -- we can remove this
```suggestion
```

Tests/JExtractSwiftTests/JNI/JNIClassTests.swift
```diff
@@ -177,19 +183,21 @@ struct JNIClassTests {
...
-        @_cdecl("Java_com_example_swift_MyClass_allocatingInit__JJ")
-        func Java_com_example_swift_MyClass_allocatingInit__JJ(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, x: jlong, y: jlong) -> jlong {
-          let self$ = UnsafeMutablePointer<MyClass>.allocate(capacity: 1)
-          self$.initialize(to: MyClass(x: Int64(fromJNI: x, in: environment!), y: Int64(fromJNI: y, in: environment!)))
-          return Int64(Int(bitPattern: self$)).getJNIValue(in: environment)
+        @_cdecl("Java_com_example_swift_MyClass__00024init__")
+        func Java_com_example_swift_MyClass__00024init__(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass) -> jlong {
+          let _result$ = UnsafeMutablePointer<MyClass>.allocate(capacity: 1)
+          _result$.initialize(to: MyClass.init())
+          let _resultBits$ = Int64(Int(bitPattern: _result$))
```
ktoso: hmm, how do we end up with both `_` and `$`, I guess the `$` would be our preferred "so we don't conflict with user identifiers". We may have had this issue before already though, so don't worry about it in this PR

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -0,0 +1,207 @@
...
+            fatalError("\(inner) memory address was null in call to \\(#function)!")
+          }
+          """
+        )
+        return "\(inner)$"
+
+      case .allocateSwiftValue(let name, let swiftType):
+        printer.print(
+          """
+          let \(name)$ = UnsafeMutablePointer<\(swiftType)>.allocate(capacity: 1)
```
ktoso: I would suggest extracting the `\(name)$` into a variable, it's a bit unclean to have to suffix the $ in every spot

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
rintaro: Hmm, I'm all for aligning the architecture of the two implementations, but I'm not sure that sharing the translation and code generation code would work well. They're two different translations, after all.
To me, this PR strikes a good balance of shared/separate logic (i.e. sharing `JavaParameter/Result` and `SwiftParameter/Result`, while keeping `Translated*` and `ConversionStep` separate).

ktoso: Cool, let's keep this as is then :-) Yeah over-tying the two might cause more trouble than it's worth.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -110,7 +110,7 @@ extension JNISwift2JavaGenerator {
       // so we pass the pointer.
       return NativeResult(
         javaType: .long,
-        conversion: .getJNIValue(.allocateSwiftValue(name: "_result", swiftType: swiftResult.type))
+        conversion: .getJNIValue(.allocateSwiftValue(name: "result", swiftType: swiftResult.type))
```
ktoso, line 113: thanks :)

PR #327, opened 2025-07-21
[JExtract/JNI] Add support for primitive non-escaping closures.

madsodgaard:
Adds support for importing non-escaping closures that only handle primitive types supported by `JavaKit`.

Just like FFM, each closure is imported as an `@FunctionalInterface` that gets passed down to JNI. Then we call the apply method using JNI.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -335,7 +335,7 @@ extension JNISwift2JavaGenerator {
             """
         )
 
-        let upcall = "environment!.interface.\(nativeResult.javaType.jniCallMethodAName)(environment, \(placeholder), methodID$, arguments$)"
+        let upcall = "environment!.interface.\(nativeResult.javaType.jniType.callMethodAName)(environment, \(placeholder), methodID$, arguments$)"
         let result = nativeResult.conversion.render(&printer, upcall)
 
         if nativeResult.javaType.isVoid {
@@ -349,24 +349,6 @@ extension JNISwift2JavaGenerator {
 
         return printer.finalize()
       }
-    }
-  }
-}
-
-extension JavaType {
-  var jniCallMethodAName: String {
-    switch self {
-    case .boolean: "CallBooleanMethodA"
-    case .byte: "CallByteMethodA"
-    case .char: "CallCharMethodA"
-    case .short: "CallShortMethodA"
-    case .int: "CallIntMethodA"
-    case .long: "CallLongMethodA"
-    case .float: "CallFloatMethodA"
-    case .double: "CallDoubleMethodA"
-    case .void: "CallVoidMethodA"
-    case .class: "CallObjectMethodA"
-    case .array: "CallObjectMethodA"
     }
   }
 }
--- a/Sources/JExtractSwiftLib/JNI/JNIType.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNIType.swift
@@ -40,6 +40,22 @@ enum JNIType {
   case jfloatArray
   case jdoubleArray
   case jobjectArray
+
+  var callMethodAName: String {
+    switch self {
+    case .jboolean: "CallBooleanMethodA"
+    case .jbyte: "CallByteMethodA"
+    case .jchar: "CallCharMethodA"
+    case .jshort: "CallShortMethodA"
+    case .jint: "CallIntMethodA"
+    case .jlong: "CallLongMethodA"
+    case .jfloat: "CallFloatMethodA"
+    case .jdouble: "CallDoubleMethodA"
+    case .void: "CallVoidMethodA"
+    case .jobject, .jstring, .jclass, .jthrowable: "CallObjectMethodA"
+    case .jbooleanArray, .jbyteArray, .jcharArray, .jshortArray, .jintArray, .jlongArray, .jfloatArray, .jdoubleArray, .jobjectArray: "CallObjectMethodA"
+    }
+  }
 }
 
 extension JavaType {


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -175,12 +175,67 @@ extension JNISwift2JavaGenerator {
...
+
+    printJavaBindingWrapperHelperClass(&printer, decl)
+
+    printJavaBindingWrapperMethod(&printer, decl)
+  }
+
+  /// Print the helper type container for a user-facing Java API.
+  ///
+  /// * User-facing functional interfaces.
+  private func printJavaBindingWrapperHelperClass(
```
ktoso, line 197: Hmm, this needs to be plural

```diff
-  private func printJavaBindingWrapperHelperClass(
+  private func printJavaBindingWrapperHelperClasses(
```

ktoso: I noticed the same mis-naming in FFM, can you change it in both please?

ktoso:
Overall looks good! I was looking at the last commit in here :)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
madsodgaard: mhm, it actually only generates a single class for each imported decl, so I guess "class" is correct here?

ktoso: Ah, I misread the code a bit -- we have one class with more funcs, ok :)

Samples/JExtractJNISampleApp/Sources/MySwiftLibrary/Closures.swift
```diff
@@ -0,0 +1,31 @@
...
+}
+
+public func closureWithInt(input: Int64, closure: (Int64) -> Int64) -> Int64 {
+  return closure(input)
+}
+
+public func closureMultipleArguments(
+  input1: Int64,
+  input2: Int64,
+  closure: (Int64, Int64) -> Int64
```
ktoso, line 26: Shall we make the params different types, say, a Double, just to make sure we're not doing something silly with types (i don't think we do, but it's good practice to make test case types a bit diverse)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
ktoso: Meh, coming back to this -- there are multiple classes here, because the `printJavaBindingWrapperFunctionTypeHelper` are classes nested in this one. 

Perhaps the `printJavaBindingWrapperFunctionTypeHelper` should include Class in the name so we can easily find "stuff that makes classes", same applies in FFM

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -175,12 +175,67 @@ extension JNISwift2JavaGenerator {
...
+        public interface \(functionType.name) {
+          \(functionType.result.javaType) apply(\(apiParams.joined(separator: ", ")));
+        }
+        """
+    )
+  }
+
+  private func printJavaBindingWrapperMethod(_ printer: inout CodePrinter, _ decl: ImportedFunc) {
+    guard let translatedDecl = translatedDecl(for: decl) else {
+      fatalError("Decl was not translated, \(decl)")
```
ktoso, line 236: Let's avoid compiler speak in user visible messages: "Declaration" might be better here

```suggestion
      fatalError("Type '\(decl.name)' was not translated: \(decl)")
```

Samples/JExtractJNISampleApp/Sources/MySwiftLibrary/Closures.swift
madsodgaard: yup 👍

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
madsodgaard: makes sense!

Tests/JExtractSwiftTests/JNI/JNIClosureTests.swift
```diff
@@ -0,0 +1,129 @@
...
+      expectedChunks: [
+        """
+        @_cdecl("Java_com_example_swift_SwiftModule__00024closureWithArgumentsAndReturn__Lcom_example_swift_SwiftModule_00024closureWithArgumentsAndReturn_00024closure_2")
+        func Java_com_example_swift_SwiftModule__00024closureWithArgumentsAndReturn__Lcom_example_swift_SwiftModule_00024closureWithArgumentsAndReturn_00024closure_2(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, closure: jobject) {
+          SwiftModule.closureWithArgumentsAndReturn(closure: { _0, _1 in
+            let class$ = environment!.interface.GetObjectClass(environment, closure)
+            let methodID$ = environment!.interface.GetMethodID(environment, class$, "apply", "(JZ)J")!
+            let arguments$: [jvalue] = [_0.getJValue(in: environment!), _1.getJValue(in: environment!)]
+            return Int64(fromJNI: environment!.interface.CallLongMethodA(environment, closure, methodID$, arguments$), in: environment!)
+          }
```
ktoso, line 122: I'd kinda want to do one `guard` for the environment like we did elsewhere but this is minor, let's revisit holistically someday 👍

ktoso (approved):
LGTM!

PR #329, opened 2025-07-23
Build: don't crash when building in SPI and JAVA_HOME is missing

ktoso:


ktoso:
This should unblock https://github.com/SwiftPackageIndex/PackageList/issues/10521 since we'll be able to run `swift package dump-package` maybe

PR #340, opened 2025-07-31
[JExtract/JNI] Support optionals in JNI mode

madsodgaard:
Adds support for optional return types and parameters in JNI mode of JExtract.

The overall idea for the conversion is that we use a "discriminator", which is a `byte :`0` or `1` indicating whether the parameter or return value is present.

For parameters we pass in the discriminator as an additional parameter.

### Returning Integral types
Returning any type that can fit inside a bigger primitive integer type, we widen that type and store the discriminator at the end. 
So for example, returning `Optional<Int8>`, the underlying JNI would return a `short`, which is `byte || byte`. The above means that this method is used for `(U)Int8, (U)Int16, (U)Int32`.

### Other types
For any other types we do an indirect return to return the value of the discriminator.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -378,7 +378,7 @@ extension JNISwift2JavaGenerator {
             guard let genericArgs = nominalType.genericArguments, genericArgs.count == 1 else {
               throw JavaTranslationError.unsupportedSwiftType(swiftResult.type)
             }
-            return try translateOptionalResult(wrappedType: swiftResult.type)
+            return try translateOptionalResult(wrappedType: genericArgs[0])
 
           default:
             guard let javaType = JNIJavaTypeTranslator.translate(knownType: knownType, config: self.config), javaType.implementsJavaValue else {
--- a/Tests/JExtractSwiftTests/JNI/JNIOptionalTests.swift
+++ b/Tests/JExtractSwiftTests/JNI/JNIOptionalTests.swift
@@ -1,0 +1,253 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import JExtractSwiftLib
+import Testing
+
+@Suite
+struct JNIOptionalTests {
+  let source =
+    """
+    class MyClass { }
+    
+    public func optionalSugar(_ arg: Int64?) -> Int32?
+    public func optionalExplicit(_ arg: Optional<String>) -> Optional<String>
+    public func optionalClass(_ arg: MyClass?) -> MyClass?
+    public func optionalJavaKitClass(_ arg: JavaLong?)
+    """
+
+  let classLookupTable = [
+    "JavaLong": "java.lang.Long",
+  ]
+
+  @Test
+  func optionalSugar_javaBindings() throws {
+    try assertOutput(
+      input: source,
+      .jni,
+      .java,
+      javaClassLookupTable: classLookupTable,
+      expectedChunks: [
+      """
+      /**
+       * Downcall to Swift:
+       * {@snippet lang=swift :
+       * public func optionalSugar(_ arg: Int64?) -> Int32?
+       * }
+       */
+       public static OptionalInt optionalSugar(OptionalLong arg) {
+         long combined$ = SwiftModule.$optionalSugar((byte) (arg.isPresent() ? 1 : 0), arg.orElse(0L));
+         byte discriminator$ = (byte) (combined$ & 0xFF);
+         int value$ = (int) (combined$ >> 32);
+         return discriminator$ == 1 ? OptionalInt.of(value$) : OptionalInt.empty();
+       }
+      """,
+      """
+      private static native long $optionalSugar(byte arg_discriminator, long arg_value);
+      """
+      ]
+    )
+  }
+
+  @Test
+  func optionalSugar_swiftThunks() throws {
+    try assertOutput(
+      input: source,
+      .jni,
+      .swift,
+      detectChunkByInitialLines: 1,
+      javaClassLookupTable: classLookupTable,
+      expectedChunks: [
+        """
+        @_cdecl("Java_com_example_swift_SwiftModule__00024optionalSugar__BJ")
+        func Java_com_example_swift_SwiftModule__00024optionalSugar__BJ(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, arg_discriminator: jbyte, arg_value: jlong) -> jlong {
+          let value$ = SwiftModule.optionalSugar(arg_discriminator == 1 ? Int64(fromJNI: arg_value, in: environment!) : nil).map {
+            Int64($0) << 32 | Int64(1)
+          } ?? 0
+          return value$.getJNIValue(in: environment!)
+        }
+        """
+      ]
+    )
+  }
+
+  @Test
+  func optionalExplicit_javaBindings() throws {
+    try assertOutput(
+      input: source,
+      .jni,
+      .java,
+      javaClassLookupTable: classLookupTable,
+      expectedChunks: [
+      """
+      /**
+        * Downcall to Swift:
+        * {@snippet lang=swift :
+        * public func optionalExplicit(_ arg: Optional<String>) -> Optional<String>
+        * }
+        */
+        public static Optional<String> optionalExplicit(Optional<String> arg) {
+          byte[] result_discriminator$ = new byte[1];
+          java.lang.String result$ = SwiftModule.$optionalExplicit((byte) (arg.isPresent() ? 1 : 0), arg.orElse(null), result_discriminator$);
+          return (result_discriminator$[0] == 1) ? Optional.of(result$) : Optional.empty();
+        }
+      """,
+      """
+      private static native java.lang.String $optionalExplicit(byte arg_discriminator, java.lang.String arg_value, byte[] result_discriminator$);
+      """
+      ]
+    )
+  }
+
+  @Test
+  func optionalExplicit_swiftThunks() throws {
+    try assertOutput(
+      input: source,
+      .jni,
+      .swift,
+      detectChunkByInitialLines: 1,
+      javaClassLookupTable: classLookupTable,
+      expectedChunks: [
+        """
+        @_cdecl("Java_com_example_swift_SwiftModule__00024optionalExplicit__BLjava_lang_String_2_3B")
+        func Java_com_example_swift_SwiftModule__00024optionalExplicit__BLjava_lang_String_2_3B(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, arg_discriminator: jbyte, arg_value: jstring?, result_discriminator$: jbyteArray?) -> jstring? {
+          let result$: jstring?
+          if let innerResult$ = SwiftModule.optionalExplicit(arg_discriminator == 1 ? String(fromJNI: arg_value, in: environment!) : nil) {
+            result$ = innerResult$.getJNIValue(in: environment!) 
+            var flag$ = Int8(1)
+            environment.interface.SetByteArrayRegion(environment, result_discriminator$, 0, 1, &flag$)
+          } // render(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:624
+          else {
+            result$ = String.jniPlaceholderValue
+            var flag$ = Int8(0)
+            environment.interface.SetByteArrayRegion(environment, result_discriminator$, 0, 1, &flag$)
+          } // render(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:634
+          return result$
+        }
+        """
+      ]
+    )
+  }
+
+  @Test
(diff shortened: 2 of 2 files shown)


Samples/JExtractJNISampleApp/src/test/java/com/example/swift/OptionalsTest.java
```diff
@@ -0,0 +1,93 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift.org project authors
```
ktoso: Minor: I just realized, I guess we should take care to make these 2024-2025 in new files when we add them

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -394,13 +657,63 @@ extension JNISwift2JavaGenerator {
...
+          \(combinedType) combined$ = \(combined);
+          byte discriminator$ = (byte) (combined$ & 0xFF);
+          """
+        )
+
+        if valueType == .boolean {
+          printer.print("boolean value$ = ((byte) (combined$ >> 8)) != 0;")
+        } else {
+          printer.print("\(valueType) value$ = (\(valueType)) (combined$ >> \(valueSizeInBytes * 8));")
+        }
```
ktoso, line 672: very nice :)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -394,13 +657,63 @@ extension JNISwift2JavaGenerator {
...
+          """
+          \(combinedType) combined$ = \(combined);
+          byte discriminator$ = (byte) (combined$ & 0xFF);
+          """
+        )
+
+        if valueType == .boolean {
+          printer.print("boolean value$ = ((byte) (combined$ >> 8)) != 0;")
+        } else {
+          printer.print("\(valueType) value$ = (\(valueType)) (combined$ >> \(valueSizeInBytes * 8));")
```
ktoso, line 671: worried about name clashes when there would be multiple optional input parameters, do we need to disambiguate the value$ name?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -72,16 +72,30 @@ extension JNISwift2JavaGenerator {
...
+
+          default:
+            guard let javaType = JNIJavaTypeTranslator.translate(knownType: knownType, config: self.config),
+                  javaType.implementsJavaValue else {
+              throw JavaTranslationError.unsupportedSwiftType(swiftParameter.type)
+            }
+
+            return NativeParameter(
+              parameters: [
+                JavaParameter(name: parameterName, type: javaType)
```
ktoso, line 93: may not matter here, but watch out if we need to add `annotations:` here; this is very new from my `@Unsigned` work, just somrething to watch out for

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -124,19 +141,147 @@ extension JNISwift2JavaGenerator {
...
+      switch swiftType {
+      case .nominal(let nominalType):
+        if let knownType = nominalType.nominalTypeDecl.knownTypeKind {
+          guard let javaType = JNIJavaTypeTranslator.translate(knownType: knownType, config: self.config),
+                javaType.implementsJavaValue else {
+            throw JavaTranslationError.unsupportedSwiftType(swiftType)
+          }
+
+          // Check if we can fit the value and a discriminator byte in a primitive.
+          // so the return JNI value will be (value || discriminator)
```
ktoso: Suuuuper nitpicky but let's not spell it as || because it makes me think "or" but it is both :) 

```suggestion
          // so the return JNI value will be (value,  discriminator)
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -124,19 +141,147 @@ extension JNISwift2JavaGenerator {
...
+          if let nextIntergralTypeWithSpaceForByte = javaType.nextIntergralTypeWithSpaceForByte {
+            return NativeResult(
+              javaType: nextIntergralTypeWithSpaceForByte.java,
+              conversion: .getJNIValue(
+                .optionalRaisingWidenIntegerType(
```
ktoso, line 250: very nice 👍

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -124,19 +141,147 @@ extension JNISwift2JavaGenerator {
...
+                  .placeholder,
+                  valueType: javaType,
+                  combinedSwiftType: nextIntergralTypeWithSpaceForByte.swift,
```
ktoso: swift for variable name is a bit weird, maybe `swiftType`?

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
```diff
@@ -36,4 +36,14 @@ extension JavaType {
     case .void: fatalError("There is no type signature for 'void'")
     }
   }
+
+  /// Returns the next integral type with space for self and an additional byte.
+  var nextIntergralTypeWithSpaceForByte: (java: JavaType, swift: String, valueBytes: Int)? {
+    switch self {
+    case .boolean, .byte: (.short, "Int16", 1)
```
ktoso: Can we use some constants for the known swift types here?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -282,13 +377,104 @@ extension JNISwift2JavaGenerator {
...
+          case .char: ("Optional<Character>", "Optional")
+          case .short: ("Optional<Short>", "Optional")
+          case .int: ("OptionalInt", "OptionalInt")
+          case .long: ("OptionalLong", "OptionalLong")
+          case .float: ("Optional<Float>", "Optional")
+          case .double: ("OptionalDouble", "OptionalDouble")
+          case .javaLangString: ("Optional<String>", "Optional")
+          default:
+            throw JavaTranslationError.unsupportedSwiftType(swiftType)
+          }
```
ktoso, line 416: This switch seems like it may want to live somewhere else as a function?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -282,13 +377,104 @@ extension JNISwift2JavaGenerator {
...
+
+          let (returnType, optionalClass) = switch javaType {
+          case .boolean: ("Optional<Boolean>", "Optional")
+          case .byte: ("Optional<Byte>", "Optional")
+          case .char: ("Optional<Character>", "Optional")
+          case .short: ("Optional<Short>", "Optional")
+          case .int: ("OptionalInt", "OptionalInt")
+          case .long: ("OptionalLong", "OptionalLong")
+          case .float: ("Optional<Float>", "Optional")
+          case .double: ("OptionalDouble", "OptionalDouble")
```
ktoso: It always _slightly_ annoyed me how the JDK randomly decided "yeah, float's not worth it" making all these asymmetric APIs... 😅

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -383,6 +637,15 @@ extension JNISwift2JavaGenerator {
       case .placeholder:
         return placeholder
 
+      case .constant(let value):
+        return value
+
+      case .explodedName(let component):
```
ktoso: hmmm... not sure if exploded is the right name 🤔 is it "composite" or "combined"? exploded to me means: `(placeholder, component)` into two distinct values

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -124,19 +141,147 @@ extension JNISwift2JavaGenerator {
...
+      case .metatype, .tuple, .existential, .opaque, .genericParameter:
         throw JavaTranslationError.unsupportedSwiftType(swiftParameter.type)
       }
     }
 
+    func translateOptionalParameter(
+      wrappedType swiftType: SwiftType,
+      parameterName: String
+    ) throws -> NativeParameter {
+      let descriptorParameter = JavaParameter(name: "\(parameterName)_descriptor", type: .byte)
```
ktoso: I think this should be 
```suggestion
      let descriptorParameter = JavaParameter(name: "\(parameterName)__discriminator$", type: .byte)
```

in order to match result_discriminator$

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -253,6 +274,70 @@ extension JNISwift2JavaGenerator {
...
+          let (translatedClass, orElseValue) = switch javaType {
+          case .boolean: ("Optional<Boolean>", "false")
+          case .byte: ("Optional<Byte>", "(byte) 0")
+          case .char: ("Optional<Character>", "(char) 0")
+          case .short: ("Optional<Short>", "(short) 0")
+          case .int: ("OptionalInt", "0")
+          case .long: ("OptionalLong", "0L")
+          case .float: ("Optional<Float>", "0f")
+          case .double: ("OptionalDouble", "0.0")
+          case .javaLangString: ("Optional<String>", #""""#)
```
ktoso:
```suggestion
          case .javaLangString: ("Optional<String>", "null")
```

can we use null for _all_ other values that are not primitives? We'd remove the default's throw here as well and just default everything to null

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
ktoso: A good testcase to add would be accepting many optionals

madsodgaard: This conversion is only for results, so should not clash.

I have a added a test for multiple optional parameters as well.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
madsodgaard: Yup! should not matter here, these are the native functions.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
madsodgaard: Stole it from FFM :D - but in the end I did not need it, so just removed it:)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
madsodgaard: yes!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
madsodgaard: yup, good idea

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
ktoso: Gotcha thanks!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
ktoso: Ha, I see :) Gotcha

Samples/JExtractJNISampleApp/Sources/MySwiftLibrary/Optionals.swift
```diff
@@ -52,6 +54,14 @@ public func optionalClass(input: Optional<MySwiftClass>) -> MySwiftClass? {
   return input
 }
 
+public func optionalJavaKitLong(input: Optional<JavaLong>) -> Int64? {
+  if let input {
+    return input.longValue()
+  } else {
+    return nil
+  }
+}
```
ktoso, line 63: very nice that this works :)

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -63,7 +63,8 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | `Foundation.Data`, `any Foundation.DataProtocol`                                     | ✅        | ❌   |
 | Tuples: `(Int, String)`, `(A, B, C)`                                                 | ❌        | ❌   |
 | Protocols: `protocol`, existential parameters `any Collection`                       | ❌        | ❌   |
-| Optional types: `Int?`, `AnyObject?`                                                 | ❌        | ❌   |
+| Optional parameters: `func f(i: Int?, class: MyClass?)`                              | ✅        | ✅   |
+| Optional return types: `func f() -> Int?`, `func g() -> MyClass?`                    | ❌        | ✅   |
```
ktoso, line 67: 🥳

PR #344, opened 2025-08-01
Mark Sendable types as @ThreadSafe

ktoso:
We do this already in javakit, even through we don't do much about threading yet... so might as well mark things as @threadsafe even if it is just informational for now


ktoso:
github got confused; closing in favor of https://github.com/swiftlang/swift-java/pull/341

PR #345, opened 2025-08-02
Update Package.swift

finestructure:
This check needs to happen at an earlier stage than SPI's build system and we've just added a new variable to indicate that things are happening in the context of SPI's processing in general via SPI_PROCESSING.

FYI @ktoso


ktoso:
Awesome, thank you!

PR #347, opened 2025-08-04
Update Package.swift

finestructure:
Improve SPI JAVA_HOME check: don't suppress proper error messaging when the package manifest is used in the context of the SPI's build system.

(This won't make a difference once Java is actually installed on our builders but gives a proper error message in the logs until that's the case.)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,8 +32,8 @@ func findJavaHome() -> String {
   }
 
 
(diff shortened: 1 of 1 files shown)


ktoso (approved):

ktoso:
Thank you!

PR #351, opened 2025-08-05
[JExtract] Fix importing extensions

rintaro:
The condition was incorrect in `Swift2JavaVisitor.visit(extensionDecl:in:)`. All valid extensions were simply ignored.


ktoso (approved):
Nice catch, thank you :)

PR #353, opened 2025-08-07
WIP: [JExtract] Add `auto` arena to SwiftKitCore and add memory management options

### LINKED ISSUE #318: Decide what to do about AutoArena (GC backed) for Android
We need `Cleaner` which is only a recent API, so decide what we want to do here.

madsodgaard:
Since `java.lang.ref.Cleaner` is not available in Android (before API 34), we introduce our own `Cleaner` which is very similar to the original JDK implementation, but a few simplications for our use case. This allows us to introduce the `Auto` arena to the JNI mode through `SwiftArena.ofAuto()`.

In an ideal world we would somehow differentiate between users needing to support Android (use our own) and people just running on regular JVM (then we use the original `Cleaner`) - perhaps that is something we could look into in the future, by providing an Android "compatibility" library of `SwiftKitCore`.

On top of that we introduce a new "memory management mode" _(maybe we need a different name?)_. This mode can be specified using `--memory-management-mode` and has the following options:

- `force-explicit`: Generates only wrappers that require explicit passing of `SwifttArena`
- `allow-automatic`: Generates wrappers that use a default global auto arena, but also explicit APIs
- `force-automatic`: Generates only wrappers that use the global auto arena, completely removing the responsibility of the user to pass any `SwiftArena`s around.

This option is only supported in JNI for now.

Resolves #318

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -16,6 +16,8 @@ import JavaTypes
 
 // MARK: Defaults
 
+private let globalArenaName = "GLOBAL_ARENA"
+
 extension JNISwift2JavaGenerator {
   /// Default set Java imports for every generated file
   static let defaultJavaImports: Array<String> = [
@@ -72,6 +74,11 @@ extension JNISwift2JavaGenerator {
     printImports(&printer)
 
     printModuleClass(&printer) { printer in
+      if config.effectiveMemoryManagementMode.requiresGlobalArena {
+        printer.print("static final SwiftArena \(globalArenaName) = SwiftArena.ofAuto();")
+        printer.println()
+      }
+
       printer.print(
         """
         static final String LIB_NAME = "\(swiftModuleName)";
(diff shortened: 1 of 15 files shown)


Sources/JavaKitConfigurationShared/GenerationMode.swift
```diff
@@ -76,3 +76,29 @@ extension JExtractMinimumAccessLevelMode {
     .public
   }
 }
+
+
+/// Configures how memory should be managed by the user
+public enum JExtractMemoryManagementMode: String, Codable {
+  /// Force users to provide an explicit `SwiftArena` to all calls that require them.
+  case forceExplicit
```
ktoso: Maybe just `explicit`
```suggestion
  case explicit
```

Sources/JavaKitConfigurationShared/GenerationMode.swift
```diff
@@ -76,3 +76,29 @@ extension JExtractMinimumAccessLevelMode {
...
+
+  /// Provide both explicit `SwiftArena` support
+  /// and a default global automatic `SwiftArena` that will deallocate memory when the GC decides to.
+  case allowAutomatic
```
ktoso: maybe this way

```suggestion
  case allowGlobalAutomatic
```

because auto is always allowed if you pass it, so this is about the global one

Sources/JavaKitConfigurationShared/GenerationMode.swift
```diff
@@ -76,3 +76,29 @@ extension JExtractMinimumAccessLevelMode {
...
+
+  /// Force all memory management to a default global automatic `SwiftArena`
+  /// that will deallocate memory when the GC decides to.
+  case forceAutomatic
```
ktoso: I'd be leaning towards not having this option -- you can do allow global and just not pass the arena anytime. It's not quite worth preventing people from doing explicvit management which is usually the "right thing" IMHO

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -16,6 +16,8 @@ import JavaTypes
 
 // MARK: Defaults
 
+private let globalArenaName = "GLOBAL_ARENA"
```
```suggestion
private let globalArenaName = "GLOBAL_SWIFT_JAVA_ARENA"
```

?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -72,6 +74,11 @@ extension JNISwift2JavaGenerator {
     printImports(&printer)
 
     printModuleClass(&printer) { printer in
+      if config.effectiveMemoryManagementMode.requiresGlobalArena {
+        printer.print("static final SwiftArena \(globalArenaName) = SwiftArena.ofAuto();")
```
ktoso: If we're doing that it should be living in SwiftKitCore and not in the generated sources as this way you'd have one per module, but we'd want one global one I think

Sources/SwiftJavaTool/Commands/JExtractCommand.swift
```diff
@@ -117,6 +121,10 @@ extension SwiftJava.JExtractCommand {
       case .wrapGuava:
         () // OK
       }
+    } else if self.mode == .ffm {
+      guard self.memoryManagementMode == .forceExplicit else {
```
ktoso: Please make an issue so we add support for it in ffm

SwiftKitCore/src/main/java/org/swift/swiftkit/core/ref/Cleaner.java
```diff
@@ -0,0 +1,67 @@
...
+package org.swift.swiftkit.core.ref;
+
+import java.lang.ref.ReferenceQueue;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.ThreadFactory;
+
+public class Cleaner implements Runnable {
```
ktoso: I'd call it SwiftCleaner just so it's easier to notice when we use which one

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftArena.java
```diff
@@ -27,6 +29,15 @@ public interface SwiftArena  {
...
     void register(SwiftInstance instance);
+
+    static ClosableSwiftArena ofConfined() {
+        return new ConfinedSwiftMemorySession(Thread.currentThread());
+    }
+
+    static SwiftArena ofAuto() {
+        ThreadFactory cleanerThreadFactory = r -> new Thread(r, "AutoSwiftArenaCleanerThread");
+        return new AutoSwiftMemorySession(cleanerThreadFactory);
+    }
```
ktoso, line 40: 🙂

ktoso:
Pretty good, I think that'd be okey -- some comments inline.

I remain unclear about which android versions don't / do have this, so that may be worth double checking before we commit our own cleaner like things but we could do this if we need to.

Sources/JavaKitConfigurationShared/GenerationMode.swift
madsodgaard: fine by me

madsodgaard:
@ktoso It was first added in API level 33: https://developer.android.com/reference/java/lang/ref/Cleaner

ktoso:
I see, and we're trying to support 28/29+ I guess... Let's take this then

ktoso (approved):

PR #364, opened 2025-08-15
Allow config if to auto load native libraries with -Dswift-java.auto-load-libraries=false

ktoso:
This way we can disable the automatic loading of libraries by passing `-Dswift-java.auto-load-libraries=false` to the jvm.

We may also consider a runtime API to do this?

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
@@ -194,8 +194,8 @@ extension FFMSwift2JavaGenerator {
         @SuppressWarnings("unused")
         private static final boolean INITIALIZED_LIBS = initializeLibs();
         static boolean initializeLibs() {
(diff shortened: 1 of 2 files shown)


Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -343,10 +343,11 @@ extension FFMSwift2JavaGenerator {
         """
         static final SymbolLookup SYMBOL_LOOKUP = getSymbolLookup();
         private static SymbolLookup getSymbolLookup() {
-            // Ensure Swift and our Lib are loaded during static initialization of the class.
-            SwiftLibraries.loadLibrary("swiftCore");
-            SwiftLibraries.loadLibrary("SwiftKitSwift");
-            SwiftLibraries.loadLibrary(LIB_NAME);
```
ktoso, line 349: The SwiftLibraries load was "overdoing it" a bit.

PR #368, opened 2025-08-18
Docs/wrap command

koloyyee:
**Title**
Adding documentation on functions in WrapCommand.swift

**Summary**
Providing a short definition about the function and parameters on generateWrappers(...) and effectiveSwiftModuleURL

**Type of Change**
Adding documentation

p.s.: I feel like I am messing up the formatting :/


ktoso:
Hmm sorry but this change seems to just be adding two comments while reformating a bunch more... I don't think this is good as is.

We do need to enable the swift-format validation -- previously it was hanging but maybe 6.2 will be working well again, I'll follow up on it https://github.com/swiftlang/swift-java/issues/371


-- 


If you want to document things perhaps focus on JavaKit APIs or even better some text in `SwiftJavaDocumentation` instead? These APIs documented in this PR don't really matter to end users.

koloyyee:
Thank you for taking the time to review this. 
I'm sorry for the trouble with the formatting. 
I will go ahead and close this PR and will dedicate my attention to SwiftJavaDocumentation and JavaKit.

PR #374, opened 2025-08-26
Add repositories configuration option in swift-java.config

bo2themax:
I was trying out a Gradle package and found out that the Maven Central repository was not enough for this package, so I added a configuration option for this. It can be configured in the `swift-java.config` file in the following format based on https://docs.gradle.org/current/userguide/supported_repository_types.html#sec:maven-repo.:

> **I’m not an experienced Java/Kotlin developer, but I believe this would meet the needs of most existing packages. If future issues arise, it could be extended to support Ivy and Flat directory repositories.**

```json
{
  "repositories": [
    { "type": "maven", "url": "https://repo.mycompany.com/maven2" },
    {
      "type": "maven",
      "url": "https://repo2.mycompany.com/maven2",
      "artifactUrls": [
        "https://repo.mycompany.com/jars",
        "https://repo.mycompany.com/jars2"
      ]
    },
    { "type": "maven", "url": "https://secure.repo.com/maven2" },
    { "type": "mavenLocal", "includeGroups": ["com.example.myproject"] },
    { "type": "maven", "url": "build/repo" }, // Relative to build folder of the temporary project, better to use absolute path here, no need to add `file:` prefix
    { "type": "mavenCentral" },
    { "type": "mavenLocal" },
    { "type": "google" }
  ]
}
```

> Note: Authentication for private repositories is not currently handled. If you need to access packages from a private repository that requires credentials, you can use Maven to download the required artifacts and then reference them via your local Maven repository in your configuration.

Alongside this PR, I’ve also added:

- A `JavaJson` target to `JavaDependencySampleApp`.
- A dedicated test under the `SwiftJavaTests` target.
- Drafted documentation in `SwiftJavaCommandLineTool.md`. If you spot any incorrect wording, please let me know in the comments or add a review in the code.

---

Test results:

image

Sample App outputs:

image


ktoso:
Looks OK but this isn't tested at all, `Samples/JavaDependencySampleApp/Sources/JavaCommonsCSV/swift-java-with-custom-repositories.config` Perhaps make a new target next to `JavaCommonsCSV` that uses this as its `swift-java.config`. 

I'd be worried about having this accepted without the sample validating it.

Sources/SwiftJavaConfigurationShared/Configuration.swift
```diff
@@ -133,6 +135,41 @@ public struct JavaDependencyDescriptor: Hashable, Codable {
...
+  public var url: String?
+  public var artifactUrls: [String]?
+
+  public init(type: RepositoryType, url: String? = nil, artifactUrls: [String]? = nil) {
+    self.type = type
+    self.url = url
+    self.artifactUrls = artifactUrls
+  }
+
+  public var descriptionGradleStyle: String? {
```
ktoso: maybe call this `renderGradleRepository()`, that's how we call things that render pieces of strings we're emitting in code gen

Samples/JavaDependencySampleApp/Sources/JavaCommonsCSV/swift-java-with-custom-repositories.config
```diff
@@ -0,0 +1,29 @@
...
+  ],
+  "repositories": [
+    {
+      "type": "maven",
+      "url": "https://jitpack.io",
+      "artifactUrls": []
+    },
+    {
+      "type": "maven",
+      "url": "file:~/.m2/repository"
```
ktoso: Does this actually work? I'd expect `file://` prefix here

ktoso:
Would be good to document this in some docs as well, as otherwise it'll be hard to discover (we have md files in SwiftJavaDocumentation for docs, feel free to add one about resolving!)

bo2themax:
> Looks OK but this isn't tested at all, `Samples/JavaDependencySampleApp/Sources/JavaCommonsCSV/swift-java-with-custom-repositories.config` Perhaps make a new target next to `JavaCommonsCSV` that uses this as its `swift-java.config`. 
> 
> 
> 
> I'd be worried about having this accepted without the sample validating it.

Yeah make sense, I'll add the target and some related documentation later

Samples/JavaDependencySampleApp/Sources/JavaCommonsCSV/swift-java-with-custom-repositories.config
bo2themax: Yeah, it is. I tested it. I'm not the expert here, but that seems like the raw URI string in Kotlin: https://docs.gradle.org/current/userguide/supported_repository_types.html#:~:text=uri(layout.buildDirectory.dir(%22repo%22))

bo2themax: You’re right, my previous test passed because of the caches. Now the `JavaRepositoryDescriptor` has been changed to an enum, and a dedicated test has been added to resolve them with repositories. Please review them again when you have time.

Package.swift
```diff
@@ -454,6 +454,7 @@ let package = Package(
       name: "SwiftJavaTests",
       dependencies: [
         "SwiftJava", 
+        "SwiftJavaTool",
```
ktoso: Please let's make this a new test target that is `SwiftJavaToolTests`

Samples/JavaDependencySampleApp/Sources/JavaDependencySample/main.swift
```diff
@@ -52,4 +58,33 @@ for record in try CSVFormatClass.RFC4180.parse(reader)!.getRecords()! {
   }
 }
 
+print("Now testing Json library...")
+
+let json = Json(#"{"host": "localhost", "port": 80}"#)
```
ktoso: Can we do this in a test in the JavaDependencySample rather than in the main.swift here? In general we'd prefer using more tests and less "everything in main" nowadays

Samples/JavaDependencySampleApp/Package.swift
```diff
@@ -99,6 +100,25 @@ let package = Package(
       ]
     ),
 
+    .target(
+      name: "JavaJson",
```
ktoso: Can we call this "OrgAndrejsJson", so it follows the recently established pattern and is more clear that it's wrapping java?
```suggestion
      name: "OrgAndrejsJson",
```

Samples/JavaDependencySampleApp/Sources/JavaJson/swift-java.config
```diff
@@ -0,0 +1,15 @@
...
+    "org.andrejs.json.JsonFactory": "JsonFactory"
+  },
+  "dependencies": [
+    "org.andrejs:json:1.2"
+  ],
+  "repositories": [
+    {
+      "type": "maven",
+      "url": "https://jitpack.io"
+    }
```
ktoso, line 13: nice!

Sources/SwiftJavaDocumentation/Documentation.docc/SwiftJavaCommandLineTool.md
```diff
@@ -199,9 +199,89 @@ struct HelloSwiftMain: ParsableCommand {
...
+    "org.apache.commons.csv.CSVFormat" : "CSVFormat",
+    "org.apache.commons.csv.CSVParser" : "CSVParser",
+    "org.apache.commons.csv.CSVRecord" : "CSVRecord"
+  }
+}
+```
+ 
+When you run `swift-java wrap-java` (or build your project with the plugin), Swift source files are generated for each mapped class. For instance, the above config will result in `CSVFormat.swift`, `CSVParser.swift`, `CSVRecord.swift`, `FilenameUtils.swift` and `IOCase.swift` files, each containing a Swift class that wraps the corresponding Java class and exposes its constructors, methods, and fields for use in Swift.
+
+This mapping allows you to use Java APIs directly from Swift, with type-safe wrappers and automatic bridging of method calls and data types.
```
ktoso, line 284: Very nice docs, thank you!

Sources/SwiftJavaDocumentation/Documentation.docc/SwiftJavaCommandLineTool.md
```diff
@@ -199,9 +199,89 @@ struct HelloSwiftMain: ParsableCommand {
...
+    { "type": "mavenLocal", "includeGroups": ["com.example.myproject"] },
+    { "type": "maven", "url": "build/repo" }, // Relative to build folder of the temporary project, better to use absolute path here, no need to add `file:` prefix
+    { "type": "mavenCentral" },
+    { "type": "mavenLocal" },
+    { "type": "google" }
+  ]
+}
+```
+
+> Note: Authentication for private repositories is not currently handled directly by `swift-java`. If you need to access packages from a private repository that requires credentials, you can use Maven to download the required artifacts and then reference them via your local Maven repository in your configuration.
```
ktoso: Please create an issue for it on github :)

Tests/SwiftJavaTests/JavaRepositoryTests.swift
```diff
@@ -0,0 +1,313 @@
...
+      try await resolve(configuration: .andriodCoreInCentral)
+    }
+    try await #expect(processExitsWith: .failure, "androidLifecycleInRepo") {
+      try await resolve(configuration: .androidLifecycleInRepo)
+    }
+  }
+
+  @Test
+  func respositoryDecoding() throws {
+    let data = #"[{"type":"maven","url":"https://repo.mycompany.com/maven2"},{"type":"maven","url":"https://repo2.mycompany.com/maven2","artifactUrls":["https://repo.mycompany.com/jars","https://repo.mycompany.com/jars2"]},{"type":"maven","url":"https://secure.repo.com/maven2","credentials":{"username":"user123","password":"secret"}},{"type":"mavenLocal","includeGroups":["com.example.myproject"]},{"type":"maven","url":"build/repo"},{"type":"mavenCentral"},{"type":"mavenLocal"},{"type":"google"}]"#.data(using: .utf8)!
```
ktoso: Maybe make this a multiline string instead? with """

Tests/SwiftJavaTests/JavaRepositoryTests.swift
```diff
@@ -0,0 +1,313 @@
...
+      let repo = remoteRepo
+    else {
+      return
+    }
+    let splitPackage = splitPackage ?? false
+
+    let process = Process()
+    process.executableURL = URL(fileURLWithPath: "/usr/bin/env")
+    process.arguments = [
+      "mvn", "dependency:get",
```
ktoso: we don't have maven on CI nodes so this will fail in CI; is there a way to test this functionality using invoking `swift-java resolve` instead? We don't need to be testing external tools and the `resolve` should cause a download after all

bo2themax:
> Looks good overall, I'm a bit unsure about the added test; can we make it work with just `resolve` and without calling out to mvn?

I think so. I can only think of embedding the package inside the test resource, or maybe leveraging Gradle caches somehow.

bo2themax:
@ktoso  Beside the reviews, I also added a few more changes this time. It’s better to go through the whole diff again🙏🏻

- Removed `System` dependency from `OrgAndrejsJsonTests`, so it can be tested on Linux.
- Added a SimpleJavaProject to JavaRepositoryTests.
- Wrapped the `nonResolvableDependency` test in a compiler check for API availability for different versions of Swift.

These four jobs are now passed both on Linux(Ubuntu 25, Swift 6.1 & 6.2) and macOS(26, Swift 6.2), so they should be able to pass on ci runners as well.
image

> **Jobs for SwiftJavaExtract seem to fail on the main branch too, so it still fails with this pull request**

ktoso:
Some of the jobs sometimes crash during compilation and that's a known Swift 6.1 issue we're aiming to resolve soon, so don't worry if compilation crashes, we can ignore those failures.

bo2themax:
> Some of the jobs sometimes crash during compilation and that's a known Swift 6.1 issue we're aiming to resolve soon, so don't worry if compilation crashes, we can ignore those failures.

Got it, but I still need to update the licence header. I will request another review after.

bo2themax:
@ktoso Two commits added

- Add missing license headers to `Tests/SwiftJavaToolTests/SimpleJavaProject`
- Update Swift Build steps when testing different Swift versions to avoid these two errors **(Will increase CI times)**
    - `error: module compiled with Swift 6.1.2 cannot be imported by the Swift 6.2 compiler: /home/ubuntu/Downloads/swift-java/.build/aarch64-unknown-linux-gnu/debug/Modules-tool/SwiftJavaTool.swiftmodule`
    - `error: no such module 'SwiftJavaTool'`

.github/workflows/pull_request.yml
```diff
@@ -147,7 +147,7 @@ jobs:
       - name: Prepare CI Environment
         uses: ./.github/actions/prepare_env
       - name: Swift Build
-        run: "swift build --build-tests --disable-sandbox"
+        run: "swift build --disable-sandbox"
```
ktoso: this change doesn't do what the commit description says; this just _stops building tests_ which isn't what we want.

bo2themax: But the following test step, will build and run the tests, right?

ktoso: Yes, the reason to build here is such that building is in a separate CI step than the testing; so we can easily spot if a failure was "during building" or "during testing"

bo2themax:
> @ktoso Two commits added
> 
> * Add missing license headers to `Tests/SwiftJavaToolTests/SimpleJavaProject`
> * Update Swift Build steps when testing different Swift versions to avoid these two errors **(Will increase CI times)**
>   
>   * `error: module compiled with Swift 6.1.2 cannot be imported by the Swift 6.2 compiler: /home/ubuntu/Downloads/swift-java/.build/aarch64-unknown-linux-gnu/debug/Modules-tool/SwiftJavaTool.swiftmodule`
>   * `error: no such module 'SwiftJavaTool'`

@ktoso Strange behaviour on my Mac though, that’s why I committed the last change:
- `swift build --build-tests --disable-sandbox` failed with SwiftJavaTool not found error
- `swift build --disable-sandbox && swift test` succeeded
> Each with a clean build, opened a issue in [SPM](https://github.com/swiftlang/swift-package-manager/issues/9080)

https://github.com/user-attachments/assets/647d1fe4-e183-40c0-87a4-eb7771f31b02

Package.swift
```diff
@@ -462,6 +462,20 @@ let package = Package(
       ]
     ),
 
+    .testTarget(
+      name: "SwiftJavaToolTests",
+      dependencies: [
+        "SwiftJavaTool",
```
ktoso: Actually, I suspect this is causing the trouble here... let's not test the executable tool target but the Lib target; AFAIR depending on executables is rather problematic

bo2themax:
To test the resolve command with `SwiftJavaToolLib`, I moved the core logic of `ResolveCommand` into a new utility type called `JavaResolver` within `SwiftJavaToolLib`.

Now both a clean `swift build --build-tests --disable-sandbox` and xcodebuild will pass, but you still cannot run tests in Xcode, due to the sandbox.

ktoso:
re-running actions, sorry for the delay here

ktoso:
Would you mind following up to resolve the build issues here or should I try to take over the PR?

bo2themax:
> Would you mind following up to resolve the build issues here or should I try to take over the PR?

Surely not, if this is not urgent. I'm afraid I might only have time over the weekend now.

bo2themax:
Should be working now

image

ktoso:
Seems like a real test failure 

✘ Test nonResolvableDependency() recorded an issue at JavaResolverTests.swift:48:11: A system failure occurred
↳ androidCoreInCentral
↳ No child processes
✘ Test nonResolvableDependency() recorded an issue at JavaResolverTests.swift:48:11: Expectation failed: .failure → .exitCode(0)
↳ androidCoreInCentral
✘ Test nonResolvableDependency() failed after 26.867 seconds with 2 issues.

bo2themax:
Hmm, strange, I have to test this on some docker to investigate this. Swift on macOS seems ok

Sources/SwiftJavaToolLib/JavaResolver.swift
```diff
@@ -0,0 +1,331 @@
...
+import SwiftJavaConfigurationShared
+import SwiftJavaShared
+#if canImport(System)
+import System
+#else
+@preconcurrency import SystemPackage
+#endif
+
+/// Utility that downloads and resolves Java dependencies for your Swift project.
+package enum JavaResolver {
```
ktoso: Let's call this more precisely `JavaDependencyResolver`

ktoso:
Terribly sorry for the delay here -- let me try to push this over the finish line 🙏

ktoso:
Finished it up over at https://github.com/swiftlang/swift-java/pull/625 thank you again and sorry for the delay!

PR #383, opened 2025-08-30
Update README and SupportedFeatures

koloyyee:
**Description**

Updating a few places from JavaKit to SwiftJava


ktoso (approved):
Thanks!

PR #386, opened 2025-09-01
[PoC] package traits to disable/remove macros when not needed

ktoso:
This is playing with the idea of using package traits to disable any use of macros when they are not necessary.


ktoso:
Dropping this idea in favor of separate repo

PR #387, opened 2025-09-05
[SwiftJava] Fix Android build by making JNINativeInterface_ public

kateinoigakukun:
The typealias has to be `public` since it's referenced by `public var interface: JNINativeInterface_`


ktoso:
re-running tests but LGTM :)

PR #390, opened 2025-09-27
Fixing export JAVA_HOME command missing quote

gevartosky:


ktoso (approved):

ktoso:
We have some macOS ci issues now which we'll resolve separately. Thanks for the update :)

PR #394, opened 2025-10-07
Add note about installing JDK

tkrajacic:
This does duplicate some of the instructions (See `Preparing your environment`) but let's go from here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ $ brew install openjdk
 $ sudo ln -sfn /opt/homebrew/opt/openjdk/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk.jdk
 
 # or if you have a distribution as cask it will be installed into /Library/Java/JavaVirtualMachines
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,6 +19,23 @@ The primary purpose of this repository is to create an environment for collabora
...
 
+On macOS for example, you can install the JDK with [homebrew](https://brew.sh) using
+
+```bash
+$ brew install openjdk
+# and create a symlink into /Library/Java/JavaVirtualMachines
+$ sudo ln -sfn /opt/homebrew/opt/openjdk/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk.jdk
+
+# or if you have a distribution as cask it will be installed into /Library/Java/JavaVirtualMachines
+$ brew install --cask temurin
```
ktoso:
```suggestion
$ brew install --cask corretto
```

let's stick with corretto since that's what we're verifying with; either should work, but to keep things consistent

PR #395, opened 2025-10-07
Explain self-publishing swiftkit

ktoso:
Until we're able to publish libraries, this workaround is required to use jextract


ktoso:
Unrelated CI trouble we're working out in https://github.com/swiftlang/swift-java/pull/396

PR #397, opened 2025-10-08
wrap-java revamp and add generic methods, types, and supertypes support

ktoso:
This reworks wrap-java on some real usage feedback and use-cases.

We now handle generics better:
- both super classes may be generic
- methods may be generic, including generics from a class used on a method level etc.

This caused quite some fallout that needed to be addressed, and is still incomplete; e.g. https://github.com/swiftlang/swift-java/issues/436

### DIFF
--- a/Sources/SwiftJava/generated/JavaOptional.swift
+++ b/Sources/SwiftJava/generated/JavaOptional.swift
@@ -4,7 +4,7 @@ import CSwiftJavaJNI
 @JavaClass("java.util.Optional")
 open class JavaOptional<T: AnyJavaObject>: JavaObject {
   @JavaMethod
-  open func get() -> T!
+  open func get() -> JavaObject! // FIXME: Currently we do generate -> T https://github.com/swiftlang/swift-java/issues/439
 
   @JavaMethod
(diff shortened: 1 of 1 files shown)


Sources/JavaStdlib/JavaLangReflect/generated/Type.swift
```diff
@@ -6,4 +6,13 @@ import CSwiftJavaJNI
 public struct Type {
   @JavaMethod
   public func getTypeName() -> String
+
+  @JavaMethod
+  public func toString() -> String
```
ktoso, line 11: Need to properly fix this for all interfaces https://github.com/swiftlang/swift-java/issues/430

understanding what's going on without being able to get toString of the java types is quite a problem, i'll tackle it next. This is obviously going to be thrown away (the edit in the generated source)

ktoso:
Parking some of the enum fun for later https://github.com/swiftlang/swift-java/issues/436

ktoso:
This finally resolved the fallout from adding more proper generics support, follow ups to be done in smaller PRs, this big one has been going for far too long.

Going to squash all that and get back into small PRs mode; the fallout of the generics was kinda huge here.

PR #400, opened 2025-10-11
Fixed problems when build JavaKitSampleApp on Windows

rayman-zhao:
These changes are for minimal effect on other parts of the swift-java. With whole picture, there should be better way to make this project cross platforms.

BTW, to successfully build still need to remove dependency of Benchmark in swift-java/Package.swift.


ktoso:
This is an ok change but we're not currently giving any guarantees this won't regress.

If you'b be interested in helping out stand up a windows version of all the github actions that would be helpful towards officially supporting windows. Currently we're still bringing up the tool and library on other platforms and not focused on windows so far.

PR #401, opened 2025-10-12
Minimum changes for Android support

madsodgaard:
This has the bare minimum changes needed to make the project work on Android.

I have tested it with the upcoming official Android SDK and works 👍 

Credit should be given to #144

### DIFF
--- a/Sources/CSwiftJavaJNI/AndroidSupport.cpp
+++ b/Sources/CSwiftJavaJNI/AndroidSupport.cpp
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2024 Apple Inc. and the Swift.org project authors
+// Copyright (c) 2025 Apple Inc. and the Swift.org project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information


lhoward:
Notes:

* #157 is not addressed; this will cause issues in any complex Android application that passes objects between threads
* if you target the correct NDK version you should be able to link against `libnativehelper.so` rather than dynamically loading it (cc @finagolfin)

ktoso:
- https://github.com/swiftlang/swift-java/issues/157 is still unresolved so that's tracked by that

For 
> if you target the correct NDK version you should be able to link against libnativehelper.so rather than dynamically loading it (cc @finagolfin)

Would you mind filing a new issue please @lhoward ?

madsodgaard:
> * if you target the correct NDK version you should be able to link against `libnativehelper.so` rather than dynamically loading it (cc @finagolfin)

I think you mean Android SDK and not NDK, right? The libnativehelper functions were exported in API 31. And we can't do conditionally linking based on version via SwiftPM unfortunately.

lhoward:
I do probably mean that, Android development details have been swapped out of my brain :) Anyway, if you need to dynamically load it, so be it.

Issue #365, opened 2025-08-15
[JExtract] Support multiple types in the same file or filename not equal to type name


madsodgaard:
Current we generate a Swift thunk for each imported type `Type+SwiftJava.swift`. The problem is that we need to tell the SwiftPM build plugin the output file names, where we pass in `filename+SwiftJava.swift`.

That means that the following code does not work in `Types.swift`:
```swift
public class Class1 {}
public class Class2 {}
```

We should change this behaviour, by perhaps generating a Swift file per imported file instead.

ktoso:
Sadly this can't quite work as `"swift file name"` -> `"same name".java` because public java types must be contained in a file of the same name... so we have to search the swift files to find out all public types, and then use them as the base names...

because the plugin at that phase cannot use swift syntax, we'll have to do some nasty "grep for public class|enum|struct|actor" basically...

We'll have to do it though

ktoso:
Disregard the above, this is about the Swift thunk files; they totally can be filename based.

ktoso:
This is now solved by https://github.com/swiftlang/swift-java/pull/403

PR #403, opened 2025-10-16
jextract: generate one output swift file per input file

ktoso:
Rather than generating an output file per TYPE which we did before.

The per-type mode cannot work correctly because SwiftPM expects to know all the output Swift files. So we'd have to parse sources and determine up front what the outputs will be -- doable, but problematic -- instead, we now generate files based on input files, and map which type goes into which output file.

This makes it also easier to find where thunks are -- they are in the same named file as the original type or func was declared in.

This may have some edge case problems still.

Resolves https://github.com/swiftlang/swift-java/issues/365

### DIFF
--- a/Samples/SwiftJavaExtractFFMSampleApp/ci-validate.sh
+++ b/Samples/SwiftJavaExtractFFMSampleApp/ci-validate.sh
@@ -3,5 +3,7 @@
 set -x
 set -e
 
+swift build # as a workaround for building swift build from within gradle having issues on CI sometimes
+
 ./gradlew run
 ./gradlew test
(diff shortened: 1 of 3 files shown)


Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MultiplePublicTypes.swift
```diff
@@ -0,0 +1,26 @@
...
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// This file exists to exercise the swiftpm plugin generating separate output Java files 
+// for the public types; because Java public types must be in a file with the same name as the type.
+
+public struct PublicTypeOne {
```
ktoso, line 18: TODO: add this same case to the JNI example

ktoso:
Pending some cleanups and handling the JNI mode

Issue #402, opened 2025-10-13
swift-java jextract: `--mode` argument ignored


gmondada:
The `--mode jni` argument is ignored on `swift-java jextract`. Code is generated in `ffm` mode.

To reproduce the problem, just run the command:

```
cd /tmp
swift-java jextract --mode jni --swift-module Foo --input-swift foo_in --output-swift foo_out/swift --output-java foo_out/java
```

The output is:

```
[info][swift-java] Run JExtractCommand: swift-java jextract --mode jni --swift-module Foo --input-swift foo_in --output-swift foo_out/swift --output-java foo_out/java
[info][swift-java] Current work directory: /private/tmp
[debug][swift-java] Load config from module base directory: /private/tmp/Sources/Foo
[debug][swift-java] Running 'swift-java jextract' in mode: ffm
[debug][swift-java] Dependent configs: 0
[warning][JExtractSwiftLib/Swift2Java.swift:39](run()) Configured java package is '', consider specifying concrete package for generated sources. 
[info][JExtractSwiftLib/Swift2Java.swift:46](run()) Input swift = foo_in 
[info][JExtractSwiftLib/Swift2Java.swift:48](run()) Input paths = [foo_in] 
[info][JExtractSwiftLib/Swift2Java.swift:55](run()) Input path: foo_in 
[swift-java] Generated: FooModule+SwiftJava.swift (at foo_out/swift/FooModule+SwiftJava.swift -- file:///private/tmp/)
[swift-java] Generated Swift sources (module: 'Foo') in: foo_out/swift/
[swift-java] Generated: Foo.java (at foo_out/java/Foo.java -- file:///private/tmp/)
[swift-java] Generated Java sources (package: '') in: foo_out/java/
[swift-java] Imported Swift module 'Foo': done.
[debug][swift-java] Done: swift-java jextract --mode jni --swift-module Foo --input-swift foo_in --output-swift foo_out/swift --output-java foo_out/java
```

On the 4th line, you see that the selected mode is `ffm`.

I can work around the problem by creating a `swift-java.config` file in the `Sources/Foo` folder.

Environment:
* MacOS 15.6.1
* swift-6.2-RELEASE (OSS)
* swift-java: 92361949bf5809e48e6e22d053ef611cccfa09ee

PR #404, opened 2025-10-19
Fix passed JExtractGenerationMode not being used by JExtractCommand.

pelekon:
When calling JExtractCommand, the value passed for --mode is not passed to Configuration, therefore the default value FFM is used.

Resolves: #402


ktoso:
Hmm weird CI issue, I'll look tomorrow thank you!

pelekon:
@ktoso I think I broke those tests, not CI. Tests should be fixed.

pelekon:
Now I think its CI fault 😄

ktoso:
Nice, thank you :)

PR #405, opened 2025-10-19
Properly extract usage of Data type from FoundationEssentials.

pelekon:
I discovered that Swift2JavaTranslator does not detect the use of Data or DataProtocol when importing FoundationEssentials. I believe that on platforms other than Darwin, the use of such imports is common in order to reduce the size of the binary file due to ICU data, so I decided to fix this.

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+SwiftThunkPrinting.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+SwiftThunkPrinting.swift
@@ -128,7 +128,7 @@ extension FFMSwift2JavaGenerator {
     let mainSymbolSourceModules = Set(
       self.lookupContext.symbolTable.importedModules.values.filter { $0.alternativeModules?.isMainSourceOfSymbols ?? false }.map(\.moduleName)
     )
-
+    
     for module in self.lookupContext.symbolTable.importedModules.keys.sorted() {
       guard module != "Swift" else {
         continue
@@ -167,9 +167,13 @@ extension FFMSwift2JavaGenerator {
         }
       }
 
-      printer.print("#else")
-      printer.print("import \(module)")
-      printer.print("#endif")
+      if (importGroups.keys.isEmpty) {
+        printer.print("import \(module)")
+      } else {
+        printer.print("#else")
+        printer.print("import \(module)")
+        printer.print("#endif")
+      }
     }
     printer.println()
   }


Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -411,7 +411,7 @@ extension FFMSwift2JavaGenerator {
               conversion: .call(.placeholder, function: "SwiftRuntime.toCString", withArena: true)
             )
 
-          case .data:
+          case .data, .essentialsData:
```
ktoso: I like `essentialsData`, I wonder if we should right away rename `data` to `foundationData` as well, could you do that please?

Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownModules.swift
```diff
@@ -96,3 +105,9 @@ private let foundationSourceFile: SourceFileSyntax = """
     public func withUnsafeBytes(_ body: (UnsafeRawBufferPointer) -> Void)
   }
   """
+
+private var foundationSourceFile: SourceFileSyntax {
+  // On platforms other than Darwin, imports such as FoundationEssentials, FoundationNetworking, etc. are used, 
+  // so this file should be created by combining the files of the aforementioned modules.
+  foundationEssentialsSourceFile
+}
```
ktoso, line 120: good 👍

Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownTypeDecls.swift
```diff
@@ -15,6 +15,7 @@
 import SwiftSyntax
 
 enum SwiftKnownTypeDeclKind: String, Hashable {
+  // MARK: -Std
```
ktoso: weird syntax? And we don't really say "std" about the stdlib -- how about just `// Swift` and `// Foundation` below

Tests/JExtractSwiftTests/DataImportTests.swift
```diff
@@ -354,6 +420,26 @@ final class DataImportTests {
         """
       ]
     )
+
+    try assertOutput(
+      input: essentials_dataProtocol_interfaceFile, .ffm, .swift,
+      expectedChunks: [
+        """
+        import FoundationEssentials
```
ktoso: I think we should handle this differently when we render imports though:


```suggestion
#if canImport(FoundationEssentials)
import FoundationEssentials
#else
import Foundation
#endif
```

and also we should have a test which has this pattern of import.

Would you be able to add that please?

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
pelekon: Sure

Tests/JExtractSwiftTests/DataImportTests.swift
pelekon: I will work on that :)

pelekon: @ktoso Done. I decided to change a little bit code for analysis of imported modules for visited source file, I hope it meets required standards :)

Sources/JExtractSwiftLib/SwiftTypes/ImportedSwiftModule.swift
```diff
@@ -0,0 +1,32 @@
...
+  init(name: String, availableWithModuleName: String? = nil, alternativeModuleNames: Set<String> = [], isMainSourceOfSymbols: Bool = false) {
+    self.name = name
+    self.availableWithModuleName = availableWithModuleName
+    self.alternativeModuleNames = alternativeModuleNames
+    self.isMainSourceOfSymbols = isMainSourceOfSymbols
+  }
+
+  func hash(into hasher: inout Hasher) {
+    hasher.combine(name)
+  }
```
ktoso, line 35: Also do the equals based on the name I think then

pelekon: I knew I forgot about something :/ I will add it after work.

pelekon:
Tests fixed

ktoso:
Very nice, thank you!

PR #407, opened 2025-10-21
Change `SwiftKitCore` to Java 11 instead of 17

madsodgaard:
Changes `SwiftKitCore` to build bytecode that supports Java 11 instead of 17, as we promise for Android.

Also updates the toolchain version to 21 LTS instead of 17.


ktoso:
`      > Cannot find a Java installation on your machine (Linux 6.11.0-1018-azure amd64) matching: {languageVersion=21, vendor=any vendor, implementation=vendor-specific, nativeImageCapable=false}. Toolchain download repositories have not been configured.`

Something's off

PR #410, opened 2025-10-24
build: fix warning about kotlin not supporting jdk 25

ktoso:
it would fall back to 24, however also pollute logs with warnings about this: "Kotlin does not yet support 25 JDK target, falling back to Kotlin JVM_24 JVM target" so we can set this explicitly.

I looked at https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support and I think this may be good enough...?

### DIFF
--- a/BuildLogic/build.gradle.kts
+++ b/BuildLogic/build.gradle.kts
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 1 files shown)


ktoso:
Meh, it's not quite right since CI uses 17... I wonder if we really have to do that with 17 at all, and we could just set lower language targets but keep using the latest JDK hmm...

WDYT @madsodgaard

PR #412, opened 2025-10-25
Java module system/Jigsaw support for SwiftKitCore(JNI) and SwiftKitFFM

chengenzhao:
Hi:

Java 9 introduces Jigsaw/Java module system. 
& 25 introduces syntax of import module
```java
import module org.swift.swiftkit.swiftkitcore;
```
so it would be nice to have the Java modularize support.

potential risk:
Not clear the impact of Android port.

thx

### DIFF
--- a/SwiftKitCore/src/main/java/module-info.java
+++ b/SwiftKitCore/src/main/java/module-info.java
@@ -1,0 +1,8 @@
+module org.swift.swiftkit.swiftkitcore {
+  requires jdk.jfr;
+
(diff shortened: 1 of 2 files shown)


ktoso:
Hmmm ill have to think about this; if there would be no downside to e.g. Android we could consider modularizing right away I guess…

ktoso:
I appreciate the change but I'd like to delay this for now unless we can think of some very specific reason we should do it immediately

PR #413, opened 2025-10-28
[JExtract/JNI] Add support for `async` Swift methods

madsodgaard:
Adds support for extracting `async` methods from Swift.

We use an upcall from Swift to Java to complete a `CompletableFuture`

### Additional changes:
* Bump to Swift 6.2
* Rename `SwiftKitSwift` to `SwiftRuntimeFunctions`
* Add `SwiftJavaRuntimeSupport` for runtime files needed by JExtract

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -202,7 +202,7 @@ let package = Package(
 
   ],
   dependencies: [
-    .package(url: "https://github.com/swiftlang/swift-syntax", from: "601.0.1"),
+    .package(url: "https://github.com/swiftlang/swift-syntax", from: "602.0.0"),
     .package(url: "https://github.com/apple/swift-argument-parser", from: "1.5.0"),
     .package(url: "https://github.com/apple/swift-system", from: "1.4.0"),
 
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ It is possible to generate Swift bindings to Java libraries using SwiftJava by u
 
 Required language/runtime versions:
 - **JDK 17+**, any recent JDK installation should be sufficient, as only general reflection and JNI APIs are used by this integratio
-- **Swift 6.0.x**, because the library uses modern Swift macros
+- **Swift 6.2.x**, because the library uses modern Swift macros
 
 **swift-java jextract** 
 
--- a/Sources/JExtractSwiftLib/Convenience/SwiftSyntax+Extensions.swift
+++ b/Sources/JExtractSwiftLib/Convenience/SwiftSyntax+Extensions.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftDiagnostics
-import SwiftSyntax
+@_spi(ExperimentalLanguageFeatures) import SwiftSyntax
 
 extension WithModifiersSyntax {
   var accessControlModifiers: DeclModifierListSyntax {
@@ -218,6 +218,8 @@ extension DeclSyntaxProtocol {
       } else {
         "var"
       }
+    case .usingDecl(let node):
+      node.nameForDebug
     }
   }
 
--- a/Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
+++ b/Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
@@ -215,7 +215,7 @@ extension SwiftType {
     switch type.as(TypeSyntaxEnum.self) {
     case .arrayType, .classRestrictionType,
         .dictionaryType, .missingType, .namedOpaqueReturnType,
-        .packElementType, .packExpansionType, .suppressedType:
+        .packElementType, .packExpansionType, .suppressedType, .inlineArrayType:
       throw TypeTranslationError.unimplementedType(type)
 
     case .attributedType(let attributedType):
--- a/Sources/JExtractSwiftLib/SwiftTypes/SwiftTypeLookupContext.swift
+++ b/Sources/JExtractSwiftLib/SwiftTypes/SwiftTypeLookupContext.swift
@@ -42,24 +42,19 @@ class SwiftTypeLookupContext {
           return typeDeclaration(for: names)
         }
 
-      case .fromFileScope(_, let names):
-        if !names.isEmpty {
-          return typeDeclaration(for: names)
-        }
-
-      case .lookInMembers(let scopeNode):
+      case .lookForMembers(let scopeNode):
         if let nominalDecl = try typeDeclaration(for: scopeNode, sourceFilePath: "FIXME.swift") { // FIXME: no path here // implement some node -> file
           if let found = symbolTable.lookupNestedType(name.name, parent: nominalDecl as! SwiftNominalTypeDeclaration) {
             return found
           }
         }
 
-      case .lookInGenericParametersOfExtendedType(let extensionNode):
+      case .lookForGenericParameters(let extensionNode):
         // TODO: Implement
         _ = extensionNode
         break
 
-      case .mightIntroduceDollarIdentifiers:
+      case .lookForImplicitClosureParameters:
         // Dollar identifier can't be a type, ignore.
         break
       }
@@ -81,8 +76,9 @@ class SwiftTypeLookupContext {
         // TODO: Implement
         _ = implicitDecl
         break
-      case .dollarIdentifier:
-        break
+      case .equivalentNames(let equivalentNames):
+        // TODO: Implement
+        _ = equivalentNames
       }
     }
     return nil


Package.swift
```diff
@@ -169,9 +169,9 @@ let package = Package(
 
     // Support library written in Swift for SwiftKit "Java"
     .library(
-      name: "SwiftKitSwift",
+      name: "SwiftJavaRuntimeSupport",
```
ktoso, line 172: Yeah good name, thx

Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Async.swift
```diff
@@ -0,0 +1,55 @@
...
+        for await number in group {
+            collectedSum += number
+        }
+
+        return collectedSum
+    }
+
+    // This is the value returned by the `withTaskGroup` closure.
+    return totalSum
+}
```
ktoso, line 38: inconsistent formatting, please stick to 2 spaces?

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
```diff
@@ -111,4 +111,29 @@ extension JavaType {
       false
     }
   }
+
+  var wrapperClassIfNeeded: JavaType {
```
ktoso: I'd call this the "boxed type" and do optional

```suggestion
  /// Returns the boxed type, or self if the type is already a Java class.
  var boxedType: JavaType {
```

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
```diff
@@ -111,4 +111,29 @@ extension JavaType {
...
+    switch self {
+    case .boolean:
+      return .class(package: nil, name: "Boolean")
```
ktoso: We probably should have those be properly package-scoped:

```suggestion
      return .class(package: "java.lang", name: "Boolean")
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -279,7 +281,11 @@ extension JNISwift2JavaGenerator {
         genericRequirements: functionSignature.genericRequirements
       )
 
-      let resultType = try translate(swiftResult: functionSignature.result)
+      var resultType = try translate(swiftResult: functionSignature.result)
+
+      if functionSignature.effectSpecifiers.contains(.async) {
```
ktoso: Maybe let's add isAsync to SwiftFunctionSignature as well

```suggestion
      if functionSignature.isAsync {
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -912,6 +966,17 @@ extension JNISwift2JavaGenerator {
     /// Access a member of the value
     indirect case replacingPlaceholder(JavaNativeConversionStep, placeholder: String)
 
+    /// `return value`
+    indirect case `return`(JavaNativeConversionStep)
+
+    /// `() -> { return body; }`
+    indirect case lambda(body: JavaNativeConversionStep)
+
+    case null
```
ktoso: Comment here, even tho pretty self explanatory

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -59,10 +59,20 @@ extension JNISwift2JavaGenerator {
         nil
       }
 
-      return try NativeFunctionSignature(
+      var result = try translate(swiftResult: functionSignature.result)
+
+      if functionSignature.effectSpecifiers.contains(.async) {
```
```suggestion
      if functionSignature.isAsync {
```

Sources/JExtractSwiftLib/SwiftTypes/SwiftEffectSpecifier.swift
```diff
@@ -14,4 +14,5 @@
 
 enum SwiftEffectSpecifier: Equatable {
   case `throws`
+  case async
```
ktoso: huh, didn't need the ``?

Sources/SwiftJavaConfigurationShared/JExtract/JExtractAsyncMode.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Configures how Swift `async` functions should be extracted by jextract.
+public enum JExtractAsyncMode: String, Codable {
```
ktoso:
```suggestion
public enum JExtractAsyncFuncMode: String, Codable {
```

to make it a little more clear maybe

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -470,6 +476,52 @@ extension JNISwift2JavaGenerator {
...
+          .aggregate([
+            .print(result.conversion),
+            .null
+          ])
+        } else {
+          result.conversion
+        }
+
+        return TranslatedResult(
+          javaType: .class(package: "java.util.concurrent", name: "CompletableFuture<\(result.javaType.wrapperClassIfNeeded)>"),
```
ktoso:
```suggestion
          javaType: .completableFuture(result.javaType),
```

And 
```swift
extension JavaType { 
  static func completableFuture(_ T: JavaType) -> JavaType {
    .class(package: "java.util.concurrent", name: "CompletableFuture<\(T.boxedType)>")
  }
}
```

SwiftKitCore/src/main/java/org/swift/swiftkit/core/ConfinedSwiftMemorySession.java
```diff
@@ -16,28 +16,26 @@
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 
 public class ConfinedSwiftMemorySession implements ClosableSwiftArena {
```
ktoso, line 23: Let's open an issue to rename this type since it is not really following the existing pattern in the JDK where Confined was about confining to a specific thread.

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftAsync.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+public final class SwiftAsync {
+
+    public static final ExecutorService SWIFT_ASYNC_EXECUTOR = new ForkJoinPool(
+            Runtime.getRuntime().availableProcessors(),
+            ForkJoinPool.defaultForkJoinWorkerThreadFactory,
+            null,
+            true
+    );
```
ktoso: Yeah so with the blocking the current approach is doing, this becomes  tricky tradeoff. The FJP _will_ just keep adding threads (here's a fun [diagram](https://doc.akka.io/libraries/akka-http/current/handling-blocking-operations-in-akka-http-routes.html) i did ages ago and like to point out).

A ThreadPoolExecutor has fixed actually upholds the max thread count, so may be preferable for blocking work, at least on server -- idk how much folks on android care, but if we are trying to fix the number of blocked threads, the ThreadPoolExecutor would be better here...

_Having that said_, how about we explore the upcall+complete based implementation first? If it is acceptable we would not have to do any blocking which would be much preferable...

Tests/JExtractSwiftTests/JNI/JNIAsyncTests.swift
```diff
@@ -0,0 +1,293 @@
...
+              swiftResult$ = await SwiftModule.async(c: c$.pointee)
+              _semaphore$.signal()
+            } // render(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:873
+          } // render(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:872
+          else {
+            Task {
+              swiftResult$ = await SwiftModule.async(c: c$.pointee)
+              _semaphore$.signal()
+            } // render(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:878
+          } // render(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:877
```
ktoso: We should not include // stuff in the assertion chunks; I know this is probably just early draft for review though, but noting so we remove it later

Samples/SwiftJavaExtractJNISampleApp/src/jmh/java/com/example/swift/AsyncBenchmark.java
```diff
@@ -0,0 +1,40 @@
...
+
+@BenchmarkMode(Mode.AverageTime)
+@Warmup(iterations = 5, time = 200, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Fork(value = 3, jvmArgsAppend = { "--enable-native-access=ALL-UNNAMED" })
+public class AsyncBenchmark {
+
+    @Benchmark
+    public long asyncSum() {
```
ktoso: return a `Long` so we don't cause unboxing overhead here

Sources/JExtractSwiftLib/SwiftTypes/SwiftEffectSpecifier.swift
madsodgaard: hm nope, but I can just add them

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -818,71 +820,66 @@ extension JNISwift2JavaGenerator {
...
 
-        return ""
+        func printTask(printer: inout CodePrinter) {
+          printer.printBraceBlock("defer") { printer in
+            // Defer might on any thread, so we need to attach environment.
+            printer.print("let deferEnvironment = try! JavaVirtualMachine.shared().environment()")
+            printer.print("environment.interface.DeleteGlobalRef(deferEnvironment, globalFuture)")
+            for outParameter in outParameters {
+              printer.print("environment.interface.DeleteGlobalRef(deferEnvironment, \(outParameter.name))")
+            }
```
ktoso, line 846: Good that'll be OK, indeed we may have to reattach the env

SwiftKitCore/src/main/java/org/swift/swiftkit/core/ConfinedSwiftMemorySession.java
madsodgaard: will do

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -815,6 +804,82 @@ extension JNISwift2JavaGenerator {
...
+        )
+
+        func printDo(printer: inout CodePrinter) {
+          printer.print("let swiftResult$ = await \(placeholder)")
+          printer.print("environment = try JavaVirtualMachine.shared().environment()")
+          let inner = inner.render(&printer, "swiftResult$")
+          if swiftFunctionResultType.isVoid {
+            printer.print("environment.interface.CallBooleanMethodA(environment, globalFuture, _JNIMethodIDCache.CompletableFuture.complete, [jvalue(l: nil)])")
+          } else {
+            printer.printBraceBlock("withVaList([SwiftJavaRuntimeSupport._JNIBoxedConversions.box(\(inner), in: environment)])") { printer in
```
0xZaph, line 833: On Android, I'm getting this compilation error:
```
withVaList([SwiftJavaRuntimeSupport._JNIBoxedConversions.box(resultBits$.getJNIValue(in: environment), in: environment)]) {
    |                                                                  `- error: cannot convert value of type 'jobject' (aka 'UnsafeMutableRawPointer') to expected element type 'any CVarArg'
```

This appears to originate from this code, though it compiles fine on Darwin. I'm not certain if this is an issue with my local Android sdk or not so just wanted to flag it.

madsodgaard: @0xZaph thanks for reporting this! seems like the issue is that the Android JNI imports jobjects as `UnsafeRawMutablePointer`, but that does not conform to `CVarArg`. And the "normal" JNI, imports it as `OpaquePointer`, which does.

I'll move away from `withVaList`, which also should fix https://github.com/swiftlang/swift-android-examples/issues/15

PR #415, opened 2025-10-31
[jextract] Propagate build plugin config to command

madsodgaard:
Correctly propagates any configuration from `swift-java.config` to the command line tool, when using the build plugin.

Also fixes the docs to correctly specify the values of enum options, as they use camelCase and not kebab-case.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/build.gradle
+++ b/Samples/SwiftJavaExtractJNISampleApp/build.gradle
@@ -102,7 +102,6 @@ def jextract = tasks.register("jextract", Exec) {
 
     workingDir = layout.projectDirectory
     commandLine "swift"
-    // TODO: -v for debugging build issues...
     args("build") // since Swift targets which need to be jextract-ed have the jextract build plugin, we just need to build
     // If we wanted to execute a specific subcommand, we can like this:
     //     args("run",/*


ktoso (changes requested):
Thanks for the fix but we should approach this differently,

make all CLI options optional and only "override" the config values if they were set. This way we don't have to go to the plugin and pass strings anywhere, and it more models what we actually want:
```
func configure<T>(_ setting: inout T, overrideWith value: T?) {
  if let value {
    setting = value
  }
}
```
and
```
    config.outputJavaDirectory = outputJava
    config.outputSwiftDirectory = outputSwift
    configure(&config.unsignedNumbersMode, overrideWith: unsignedNumbers) <<<<
    config.minimumInputAccessLevelMode = minimumInputAccessLevel
    config.memoryManagementMode = memoryManagementMode
```

The problem is `@Flag` which does not support `Bool?` huh... so we need to figure that one out 🤔 Perhaps by manually parsing the command line arguments to see if it was passed or not?

Samples/SwiftJavaExtractJNISampleApp/build.gradle
```diff
@@ -103,7 +103,7 @@ def jextract = tasks.register("jextract", Exec) {
     workingDir = layout.projectDirectory
     commandLine "swift"
     // TODO: -v for debugging build issues...
```
ktoso: If removing the -v let's remove the TODO as well (this was for debugging commands on CI a while back)

```suggestion
```

ktoso (approved):
LGTM thanks for adjusting how we do this :)

ktoso:
We're having CI issues... I'll see what I can do about that, sadly 6.2 didn't get the fix yet we're waiting on

PR #417, opened 2025-11-04
Adjust Task.immediate use so we can bring back Swift 6.1 CI until 6.2 is fixed

ktoso:
We lost ability to CI with 6.1 since we introduced Task.immediate in the asynchronous func support. 

We need 6.1 support because 6.1.3 has a SwiftPM fix that prevents random build crashes.
6.2 has missed this fix and will only get it fixed in next releases: https://github.com/swiftlang/swift-package-manager/pull/9314 `Frequent crashes in LLBuildProgressTracker while building swift-java on linux and swift`.

This #if guards code using Task.immediate, adjusts the tests, and also brings back 6.1 CI. 

Failures with 6.2 are expected and will continue for the time being.


ktoso:
We'd have to platform guard the Task.immediate for the time being as a workaround perhaps... It's not great not having linux CI for all those samples so I'm looking at potential workarounds until 6.2 is fixed.

PR #421, opened 2025-11-04
Start replacing ad-hoc logging with swift-log in SwiftJavaTool

ktoso:
We had a lot of all-over-the-place logging which pretended to be loggers or just println + debug and info prefixes etc. This is getting out of hand and we should adopt a consistent logging mechanism.

The way to do this in Swift is swift-log, there's no reason anymore to reinvent the wheel about it. The dependency is small and likely to be present in most libs. Especially the source gen we need not be concnerned about using swift-log. Runtime does not have to use it.

This is just some initial steps, separated out from #397


ktoso:
I'll merge this as a "first step" and do follow ups later on

Issue #418, opened 2025-11-04
CI: Linux + 6.2 builds are broken with duplicated symbol issues when prebuilt swift-syntax is used


ktoso:
```

/usr/bin/ld.gold: /__w/swift-java/swift-java/.build/x86_64-unknown-linux-gnu/debug/SwiftDiagnostics-tool.build/FixIt.swift.o: previous definition here
/usr/bin/ld.gold: error: /__w/swift-java/swift-java/.build/prebuilts/swift-syntax/602.0.0/swift-6.2-RELEASE-MacroSupport-ubuntu_jammy_x86_64/lib/libMacroSupport.a(FixIt.swift.o): multiple definition of '$s16SwiftDiagnostics5FixItVN'
/usr/bin/ld.gold: /__w/swift-java/swift-java/.build/x86_64-unknown-linux-gnu/debug/SwiftDiagnostics-tool.build/FixIt.swift.o: previous definition here
/usr/bin/ld.gold: error: /__w/swift-java/swift-java/.build/prebuilts/swift-syntax/602.0.0/swift-6.2-RELEASE-MacroSupport-ubuntu_jammy_x86_64/lib/libMacroSupport.a(FixIt.swift.o): multiple definition of '$s16SwiftDiagnostics5FixItV6ChangeON'
/usr/bin/ld.gold: /__w/swift-java/swift-java/.build/x86_64-unknown-linux-gnu/debug/SwiftDiagnostics-tool.build/FixIt.swift.o: previous definition here
/usr/bin/ld.gold: error: /__w/swift-java/swift-java/.build/prebuilts/swift-syntax/602.0.0/swift-6.2-RELEASE-MacroSupport-ubuntu_jammy_x86_64/lib/libMacroSupport.a(IfConfigDecl+IfConfig.swift.o): multiple definition of '$s11SwiftSyntax012IfConfigDeclB0V0acD0E12activeClause2inAA0cdgB0VSg6clause_Say0A11Diagnostics10DiagnosticVG11diagnosticstx_tAD18BuildConfigurationRzlF'
/usr/bin/ld.gold: /__w/swift-java/swift-java/.build/x86_64-unknown-linux-gnu/debug/SwiftIfConfig-tool.build/IfConfigDecl+IfConfig.swift.o: previous definition here
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[1425/1426] Linking swift-javaPackageTests.xctest
Error: Process completed with exit code 1.
```

[...]

PR #422, opened 2025-11-05
Disable swift-syntax prebuilts (primarily on linux)

ktoso:
Because of problems Swift 6.2 exhibits with them on Linux.

We may bring this back once 6.2 is fixed, or turn it off more precisely specifically only on linux, even when called from gradle etc.

This is a workaround for https://github.com/swiftlang/swift-java/issues/418


ktoso:
finally no crashes anymore~!

PR #426, opened 2025-11-05
[jextract/jni] update docs about throwing and optional init and add tests

madsodgaard:
We already support throwing and optional initializers:-)


ktoso:
LGTM, sad about the growing gap between the modes hmmm, would be nice to do both sometimes when doing features i guess but they are so very different hmm

PR #428, opened 2025-11-06
[jextract] fix support for objects in async

madsodgaard:
Adds support for using objects as parameters and return types in `async` functions.


Tests/JExtractSwiftTests/JNI/JNIAsyncTests.swift
```diff
@@ -353,4 +353,65 @@ struct JNIAsyncTests {
...
+      .jni, .java,
+      detectChunkByInitialLines: 2,
+      expectedChunks: [
+        """
+        /**
+         * Downcall to Swift:
+         * {@snippet lang=swift :
+         * public func async(s: String) async -> String
+         * }
+         */
```
ktoso: I'd skip that from the expected chunk, it doesn't matter for this test and only makes adjusting them harder if we change the docs generation

PR #435, opened 2025-11-11
jextract: add support for arrays

madsodgaard:
Adds support the JNI mode to import primitive arrays, as well as arrays of classes/structs.

### DIFF
--- a/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
+++ b/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
@@ -58,7 +58,7 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Stored properties: `var`, `let` (with `willSet`, `didSet`)                           | ✅        | ✅   |
 | Computed properties: `var` (incl. `throws`)                                          | ✅ / TODO | ✅   |
 | Async functions `func async` and properties: `var { get async {} }`                  | ❌        | ✅   |
-| Arrays: `[UInt8]`, `[T]`                                                             | ❌        | ❌   |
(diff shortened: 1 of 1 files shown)


Sources/JExtractSwiftLib/CodePrinter.swift
```diff
@@ -86,12 +86,17 @@ public struct CodePrinter {
 
   public mutating func printBraceBlock(
     _ header: Any,
+    postHeaderBracket: Any? = nil,
```
ktoso: ~~let's call this `prologue`~~ maybe call this `parameters: [String]`? and suffix it with the `in`?

Tests/JExtractSwiftTests/JNI/JNIArrayTest.swift
```diff
@@ -0,0 +1,37 @@
...
+      input: "public func f(array: Array<UInt8>) -> Array<UInt8> {}",
+      .jni, .java,
+      detectChunkByInitialLines: 1,
+      expectedChunks: [
+        """
+        sdfad
+        """,
+        """
+        sdf
+        """,
```
ktoso, line 35: TODO?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -894,6 +1015,16 @@ extension JNISwift2JavaGenerator {
         }
 
         return ""
+
+      case .closure(let args, let body):
+        var printer = CodePrinter()
+        printer.printBraceBlock("", postHeaderBracket: "(\(args.joined(separator: ", "))) in") { printer in
```
ktoso: yeah let's make this accept the `args` raw, and call it `parameters:` so we don't have to join everywhere in callers

PR #437, opened 2025-11-12
jextract/jni: use deferEnvironment instead of environment interface

ktoso:
in defer that's destroying refs to objects.

Slightly cleaner, though where we get the .interface doesn't matter as much


ktoso:
wdyt @madsodgaard ?

madsodgaard (approved):
Yeah this is probably the most correct way

Issue #439, opened 2025-11-12
Correctly importing Optional::get()  as -> T causes MethodNotFound


ktoso:
We are now able to handle generics more properly in wrap-java and now we import Optional as:

```
@JavaClass("java.util.Optional")
open class JavaOptional<T: AnyJavaObject>: JavaObject {
  @JavaMethod
  open func get() -> T!

```

which then crashes

`SwiftJava/JavaOptional.swift:2: Fatal error: 'try!' expression unexpectedly raised an error: java.lang.NoSuchMethodError: Ljava/util/Optional;.get()Ljava/lang/String;`

so the String bit is coming unexpectedly here; even though it's the correct signature.

The problem is somewhere in the `dynamicJavaMethodCall` infra which can't handle the generic properly when forming the downcall signature

PR #440, opened 2025-11-12
JavaMethod's need to handle type erasure; Enable Optional::get() -> T

ktoso:
Now that we support generics, we need to do special handling for `-> T` upcalls... since the T would be replaced with a real type with Swift, which then messes up the function signature we're looking for in the JVM.

This works towards adjusting the `@JavaMethod` macro and then we have to adjust wrap-java to make use of that to "enable type erasure" when doing these calls.

This is some first steps, hardcoding the Optional::get() and i'll adjust the generator next


ktoso:
resolves https://github.com/swiftlang/swift-java/issues/439

PR #445, opened 2025-11-14
Allow // comments in swift-java.config "JSON with comments"

ktoso:
Grew tired of not being able to comment in config files, so now we support line comments with `//`


ktoso:
I'd love to support https://github.com/apple/pkl-swift sometime, but being careful about how we integrate that so we don't pull too many more deps

PR #458, opened 2025-11-20
feat: also pass mode and loglevel to jextract

pwallrich:
Currently the config values for mode and logLevel that are set inside the swift-java.config are ignored when the plugin is used.

Since jextract already uses those values they only need to be passed to jextract as command line arguments


madsodgaard:
Hm, this should actually already work. This is handled by
https://github.com/swiftlang/swift-java/blob/6a75aeb726e87da413b97682be2c966d8eaaabbe/Sources/SwiftJavaTool/Commands/JExtractCommand.swift#L88

such as:
```swift
configure(&config.mode, overrideWith: self.mode)
```

Are you seeing that it doesn't work?

pwallrich:
That's interesting. 

There seem to be some issues with my local project setup then since it didn't work without those settings.

I just tried one of the samples and it works. I need to investigate what the issue in my project is.

Issue #352, opened 2025-08-05
[JExtract] Support subscript APIs


rintaro:
Support subscript APIs. E.g.

```swift
public struct MySwiftStruct {
  public subscript(x: Int) -> Int { get set }
}
```

`subscript` accessors can be modeled in `ImportedFunc` as `.getter`/`.setter` with parameters, or new `SwiftAPIKind`.

PR #459, opened 2025-11-22
jextract (ffm, jni): Subscripts support

pelekon:
Resolves: #352


Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftStruct.swift
```diff
@@ -61,4 +61,9 @@ public struct MySwiftStruct {
   public func makeRandomIntMethod() -> Int {
     return Int.random(in: 1..<256)
   }
+
+  public subscript() -> Int {
+    get { return len }
+    set { len = newValue }
+  }
```
ktoso, line 85: What about a case with parameters?

Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftStruct.swift
```diff
@@ -61,4 +61,9 @@ public struct MySwiftStruct {
   public func makeRandomIntMethod() -> Int {
     return Int.random(in: 1..<256)
   }
+
+  public subscript() -> Int {
+    get { return len }
+    set { len = newValue }
```
ktoso: can it set some other value and not the `len` perhaps? Just so we don't step on values when we write bigger tests

Samples/SwiftJavaExtractFFMSampleApp/src/test/java/org/swift/swiftkitffm/MySwiftStructTest.java
```diff
@@ -33,4 +33,15 @@ void create_struct() {
             assertEquals(len, struct.getLength());
         }
     }
+
+    @Test
+    void testSubscript() {
+        try (var arena = AllocatingSwiftArena.ofConfined()) {
+            MySwiftStruct s = MySwiftStruct.init(1337, 42, arena);
+            long currentValue = s.getSubscript();
+            s.setSubscript(66);
```
ktoso, line 42: nice

Tests/JExtractSwiftTests/FFM/FFMSubscriptsTests.swift
```diff
@@ -0,0 +1,314 @@
...
+
+        import org.swift.swiftkit.core.*;
+        import org.swift.swiftkit.core.util.*;
+        import org.swift.swiftkit.ffm.*;
+        import org.swift.swiftkit.core.annotations.*;
+        import java.lang.foreign.*;
+        import java.lang.invoke.*;
+        import java.util.*;
+        import java.nio.charset.StandardCharsets;
+        """,])
```
ktoso: We can probably skip the imports from these tests; the more these tests assert text the more annoying it is to change things in the future;

Let's just test the things that this specific "feature" is

Tests/JExtractSwiftTests/FFM/FFMSubscriptsTests.swift
```diff
@@ -0,0 +1,314 @@
...
+        public static final GroupLayout $LAYOUT = (GroupLayout) SwiftValueWitnessTable.layoutOfSwiftType(TYPE_METADATA.$memorySegment());
+        public final GroupLayout $layout() {
+          return $LAYOUT;
+        }
+
+        private MyStruct(MemorySegment segment, AllocatingSwiftArena arena) {
+          super(segment, arena);
+        }
+        """
+      ])
```
ktoso: Please remove those thunks too, let's not assert anything that's not related to the actual subscripts -- like these chunks

Tests/JExtractSwiftTests/FFM/FFMSubscriptsTests.swift
```diff
@@ -0,0 +1,314 @@
...
+            throw new AssertionError("should not reach here", ex$);
+          }
+        }
+      """,
+      """
+      public void setSubscript(int index, int newValue) {
+        $ensureAlive();
+        swiftjava_SwiftModule_MyStruct_subscript$set.call(index, newValue, this.$memorySegment());
+      """,
+    ])
```
ktoso, line 164: that's the chunks which we do want to assert on, looking good, thanks!

Tests/JExtractSwiftTests/JNI/JNISubscriptsTests.swift
```diff
@@ -0,0 +1,353 @@
...
+          long self$ = this.$memoryAddress();
+          if (CallTraces.TRACE_DOWNCALLS) {
+            CallTraces.traceDowncall("MyStruct.$typeMetadataAddress",
+            "this", this,
+            "self", self$);
+          }
+          return MyStruct.$typeMetadataAddressDowncall();
+        }
+        """
+      ])
```
ktoso: same request about only the subscript text chunks as in the ffm one

ktoso:
Very nice, just some nitpicks about the tests

Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftStruct.swift
pelekon: I didn't notice I lost them in git stash :/ Fixed

PR #460, opened 2025-11-22
Support UInt handling in jextract JNI mode

pelekon:
I believe that both modes should behave similarly when handling Int/UInt types. This PR makes JNI handle these types in the same way as FFM.

### DIFF
--- a/Tests/JExtractSwiftTests/UnsignedNumberTests.swift
+++ b/Tests/JExtractSwiftTests/UnsignedNumberTests.swift
@@ -322,4 +322,58 @@ final class UnsignedNumberTests {
       expectedChunks: expectedChunks
     )
   }
+  
+  @Test(
+    "Import: take UInt return UInt (annotate)",
+    arguments: [
+      (
+        JExtractGenerationMode.ffm,
+        /* expected chunks */
+        [
+          """
+          /**
+           * {@snippet lang=c :
+           * size_t swiftjava_SwiftModule_unsignedLong_first_second(size_t first, size_t second)
+           * }
+           */
+          private static class swiftjava_SwiftModule_unsignedLong_first_second {
+            private static final FunctionDescriptor DESC = FunctionDescriptor.of(
+            /* -> */SwiftValueLayout.SWIFT_INT
+            /* first: */SwiftValueLayout.SWIFT_INT
+            /* second: */SwiftValueLayout.SWIFT_INT
+          );
+          """,
+          """
+          @Unsigned
+          public static long unsignedLong(@Unsigned long first, @Unsigned long second) {
+            return swiftjava_SwiftModule_unsignedLong_first_second.call(first, second);
+          }
+          """,
+        ]
+      ),
+      (
+        JExtractGenerationMode.jni,
+        /* expected chunks */
+        [
+          """
+          @Unsigned
+          public static long unsignedLong(@Unsigned long first, @Unsigned long second) throws SwiftIntegerOverflowException {
+            return SwiftModule.$unsignedLong(first, second);
+          }
+          private static native long $unsignedLong(long first, long second);
+          """,
+        ]
+      ),
+    ])
+  func echo_uint_annotate(mode: JExtractGenerationMode, expectedChunks: [String]) throws {
+    let config = Configuration()
+
+    try assertOutput(
+      input: "public func unsignedLong(first: UInt, second: UInt) -> UInt",
+      config: config,
+      mode, .java,
+      detectChunkByInitialLines: 2,
+      expectedChunks: expectedChunks
+    )
+  }
 }


madsodgaard:
Thanks for this! Definitely an important step that we handle `Int`/`UInt` soon, because that is the most common used integer type in Swift.

I do however, think we need to give this some more thought. Especially since the JNI mode will probably be used a lot by Android folks, which is commonly 32-bit. If you have the following Swift function:
```swift
func f(i: Int)
```
with this PR it gets translated to
```java
void f(long i)
```
however, the downcall to Swift would be `Int64(fromJNI: i, ...)`, which is not compatible with `Int` in Swift. so that would not work. I am not sure if its ok to just drop the top 32-bits, if we are running on a 32-bit machine. Maybe we need some kind of annotation mode like the unsigned stuff? Or throwing an error, if we are not in `Int32.min...Int32.max` on 32-bit? What do you think?

Also we should add some codegen tests and runtime tests while we are it, as well as add the `UInt` to the `@Unsigned` annotation handling

pelekon:
I am aware of Int64 -> Int32 conversion problem but I didn't think about it as problem needed to be addressed in the same PR. To be honest I thought 32-bit is rare for Android, so I have no problem to work on solution for it in this PR. 
I don't know much about annotations in Java, so for potential solutions I thought about SwiftInt java object which will throw IntegerOverflowException in those cases. 
Of course I will add tests for it too :)

ktoso:
Thanks for the PR, this would be good to address.

Overall the two "modes" should behave the same wherever possible and it is a problem that they're a bit independent currently so this would be good to align the two.

32bit is _very_ rare and AFAICS there is a number of prominent apps that choose to not support 32bit Android devices. 

The Java annotation mentioned is the https://github.com/swiftlang/swift-java/blob/main/SwiftKitCore/src/main/java/org/swift/swiftkit/core/annotations/Unsigned.java so please have a look how we're using it in the ffm mode to mark something was an "U"Int. The imported parameters or return types basically get an `@Unsigned long` etc, this is valuable because users can then handle it and notice in the API description, without having to guess

madsodgaard:
I know 32-bit is _rare_ these days. But since it's still a supported platform, for both Swift and the Swift Android SDK, I think we should still handle it. For example for us, most of our devices are 32-bit, and that is in the millions:)

OK by me to merge this and we can tackle it in another PR, just wanted to flag this, as we need to handle it somehow. 
An idea could be sourcegen something similar to this:

```swift
@_cdecl()
func $f(i: jlong) {
  let swiftInt = Int64(fromJNI: i, ...)
  #if _pointerBitWidth(_32)
  guard swiftInt <= Int32.max, swiftInt >= Int32.min else {
    // throw overflow exception
  }
  #endif
}
```

maybe this could even just be in the `extension Int: JavaValue` methods.

ktoso:
Well we can't merge anything this week, so yes, let's please polish this up before branches get unlocked next week.

pelekon:
I committed small change with a proposed solution for check and basic tests. Let me know what you think, and if the response is positive, I'll cover rest of JNI cases(enum cases etc) and apply it to FFM.

Sources/JExtractSwiftLib/JNI/JNIJavaTypeTranslator.swift
```diff
@@ -61,4 +56,38 @@ enum JNIJavaTypeTranslator {
       return nil
     }
   }
+
+  static func indirectConversionSetepSwiftType(for knownKind: SwiftKnownTypeDeclKind, from knownTypes: SwiftKnownTypes) -> SwiftType? {
```
ktoso: likely `indirectConversionStepSwiftType`?

Sources/JavaTypes/JavaException.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Describes a Java exception class (e.g. `SwiftIntegerOverflowException`)
+public struct JavaException: Equatable, Hashable {
```
ktoso: Probably need to call this `JavaExceptionType` so we don't confuse ourselfes with a JavaExpcetion that would be a @JavaClass

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftIntegerOverflowException.java
```diff
@@ -0,0 +1,21 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package org.swift.swiftkit.core;
+
+public class SwiftIntegerOverflowException extends RuntimeException {
```
ktoso, line 27: Please add some docs on it, the message is pretty good but we should have similar explanation in docs

Tests/JExtractSwiftTests/JNI/JNIConversionChecksTests.swift
```diff
@@ -0,0 +1,101 @@
...
+  @Test func generatesUnsignedSetterWithCheck() throws {
+    try assertOutput(input: unsignedSource, .jni, .swift, expectedChunks: [
+      """
+      @_cdecl("Java_com_example_swift_MyStruct__00024setUnsignedInt__JJ")
+      func Java_com_example_swift_MyStruct__00024setUnsignedInt__JJ(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, newValue: jlong, self: jlong) {
+        let indirect_newValue = UInt64(fromJNI: newValue, in: environment)
+        #if _pointerBitWidth(_32)
+          guard indirect_newValue >= UInt32.min && indirect_newValue <= UInt32.max else {
+            environment.throwJavaException(javaException: .integerOverflow)
+            return
```
ktoso: include the #endif in this chunk please

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -293,11 +293,43 @@ extension JNISwift2JavaGenerator {
...
+        case .check32BitIntOverflow:
+          int32OverflowChecks.append(parameter.conversionCheck!.render(&printer, "indirect_\(javaParameterName)"))
+        case nil:
+          break
+      }
+    }
+
+    // Make indirect variables
+    for (name, lowered) in indirectVariables {
+      printer.print("let indirect_\(name) = \(lowered)")
```
ktoso: we have a naming convention for variables to suffix with `$` so we don't clash with user defined variables; It's very unlikely anyone would do some "indirect_bla" but please let's be consistent.

This could be `name$indirect`?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -616,6 +643,12 @@ extension JNISwift2JavaGenerator {
 
     /// Represents how to convert the JNI parameter to a Swift parameter
     let conversion: NativeSwiftConversionStep
+
+    /// Represents swift type for indirect variable used in required checks, e.g Int64 for Int overflow check on 32-bit platforms
+    let indirectConversion: NativeSwiftConversionStep?
```
ktoso, line 667: Maybe let's add that "This will introduce a new name$indirect variabler" so it is easier to mentally connect the two? The "indirect" name sounds good!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -616,6 +643,12 @@ extension JNISwift2JavaGenerator {
...
+
+    /// Represents check operations executed in if/guard conditional block for check during conversion
+    let conversionCheck: NativeSwiftConversionCheck?
```
ktoso, line 670: Very nicely integrated here 👍

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -293,11 +293,43 @@ extension JNISwift2JavaGenerator {
...
+    }
+
+    if !int32OverflowChecks.isEmpty {
+      printer.print("#if _pointerBitWidth(_32)")
+
+      for check in int32OverflowChecks {
+        printer.printBraceBlock("guard \(check) else") { printer in 
+          let dummyReturn = dummyReturn(for: nativeSignature)
+          printer.print("environment.throwJavaException(javaException: .integerOverflow)")
+          printer.print(dummyReturn)
```
ktoso: I'd skip the dummyreturn variable:

```suggestion
          printer.print(dummyReturn(for: nativeSignature))
```

ktoso (commented):
Overall this looks quite neat, thank you.

Can you add some runtime tests please?

madsodgaard:
This looks great!

I only have one comment regarding the implementation for maintainability. Up until now all conversions have just been done in the `ConversionStep` types. This includes any checks as well (if, guards), for example for nullable checks. As more people become interested in collaborating on the project, perhaps we should make sure we unify the way we do conversions. Just a bit worried that we deviate and do things differently in the project. Makes it harder for new contributors!

Was there a specific reason for introducing the "indirect" and "check" types? And not just doing it inside a `ConversionStep`?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
pelekon: Sure, I like `name$indirect`

Tests/JExtractSwiftTests/JNI/JNIConversionChecksTests.swift
pelekon: I added endif in next chunk. I wanted to remove closing breace from this chunk as it contains comment generated by printer about file and line of invokation which isn't nice for those assertions.

pelekon:
Firstly, thank you @madsodgaard for sharing your concern :)

> Was there a specific reason for introducing the "indirect" and "check" types? And not just doing it inside a `ConversionStep`?

Of course, but I realise that for others this may not be a sufficient reason. As someone who just started to contribute I felt `NativeSwiftConversionStep` already does a lot and I didn't want to add too much to it. From my perspective, it should handle conversion between Java and Swift, while other things, such as checking and validation, should be handled by something else to separate their responsibilities. That is why I wanted to create a separate mechanism for handling this; I believe that this approach increases maintainability. Of course, I won't insist on this if you and @ktoso decide that the current approach is better, I have no objection to adapting to it :)

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
```diff
@@ -103,13 +103,24 @@ enum SwiftType: Equatable {
...
       switch nominal.nominalTypeDecl.knownTypeKind {
-      case .uint8, .uint16, .uint32, .uint64: true
+      case .uint8, .uint16, .uint32, .uint64, .uint: true
       default: false
       }
     default: false
     }
   }
 
+  var isArchDependingInteger: Bool {
```
ktoso, line 113: Typo

```suggestion
  var isArchDependentInteger: Bool {
```

Tests/JExtractSwiftTests/JNI/JNIIntConversionChecksTests.swift
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+  """
+  private let enumSource = """
+  public enum MyEnum {
+    case firstCase
+    case secondCase(UInt)
+  }
+  """
+
```
ktoso, line 57: Can you please add a test to `UnsignedNumberTests.swift` that the `.java` output in `.jni` mode also looks correct? The unsigned integers should appear as `void unsignedInt(@Unsigned int arg) {` to Java.

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftIntegerOverflowException.java
```diff
@@ -0,0 +1,41 @@
...
+ * thrown automatically by underlaying code for method.
+ * </p>
+ *
+ * <p>
+ * Inheritance hierarchy:
+ * <ul>
+ *   <li>{@link java.lang.RuntimeException}</li>
+ *   <li>SwiftIntegerOverflowException</li>
+ * </ul>
+ * </p>
```
ktoso: Thank you for the docs, ok to drop the inheritance though -- javadoc does this automatically very well

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -103,11 +103,16 @@ extension JNISwift2JavaGenerator {
...
+            let indirectStepType = JNIJavaTypeTranslator.indirectConversionStepSwiftType(for: knownType, from: knownTypes)
+            let indirectCheck = JNIJavaTypeTranslator.checkStep(for: knownType, from: knownTypes)
+
             return NativeParameter(
               parameters: [
                 JavaParameter(name: parameterName, type: javaType)
               ],
-              conversion: .initFromJNI(.placeholder, swiftType: type)
+              conversion: indirectStepType != nil ? .labelessAssignmentOfVariable(.placeholder, swiftType: type) : .initFromJNI(.placeholder, swiftType: type),
+              indirectConversion: indirectStepType.flatMap { .initFromJNI(.placeholder, swiftType: $0) },
```
ktoso, line 115: Hmm, would we be able to make this indirect conversion step the parameter the same as `conversionStep`? However, I do like the `conversionCheck`, so we'd keep that perhaps, WDYT? Makes sense? 

I don't love the fact we have two conversion steps and if one is nil we want the other one etc. The check is nice because it is always in addition to a conversion

ktoso:
A few comments, this is shaping up quite well

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
pelekon: Sure, I will move indirectConversion into conversion. I am not sure how soon I can do it cuz my PC died, but I will try to do it asap.

ktoso: Oh that sucks to hear :\ Hope you can get a functional computer soon enough. thanks for the info!

ktoso:
Happy new year!

I updated the PR however this is problematic since now the JNI ones are throwing but the FFM ones not when overflow happens... we should make sure both modes work the same 🤔 

Maybe we'll just have to follow up on it in a follow up PR

ktoso:
I'm inclined to accept this as is and we can continue fleshing out the differences in follow ups

pelekon:
> Happy new year!
> 
> I updated the PR however this is problematic since now the JNI ones are throwing but the FFM ones not when overflow happens... we should make sure both modes work the same 🤔
> 
> Maybe we'll just have to follow up on it in a follow up PR

Firstly sorry for late update. Opposite to my initial plan I went to visit family sooner and unfortunately new year party left me with infection which made be bedridden until this day. Finally I am starting to feel better so I will try to start working on FFM around monday.

ktoso:
No problem at all! I mostly didn’t want to waste your work so figured we can merge as is :-)

happy new year and thanks in advance!

PR #464, opened 2025-11-28
Fix JNI caching for native threads

### LINKED ISSUE #419: `swift-java` does not work on Android API 28-30, due to missing JavaVM JNI symbols
Currently, `swift-java` can crash runtime for Android API 28-30, if you try to use the `JavaVirtualMachine` class. This means that any JExtract generated sources _(JNI mode)_ only works on API 31+, since we rely on `JavaVirtualMachine.shared().environment()`.

This is due to the fact that Android first added _official_ support for the JavaVM JNI functions, such as `JNI_GetCreatedJavaVMs`, in API 31+.
We rely on these methods for getting the active JavaVM, through for example `JavaVirtualMachine.shared()`.

We have a shim for these JNI functions in https://github.com/swiftlang/swift-java/blob/main/Sources/CSwiftJavaJNI/AndroidSupport.cpp - this works on API 31+, because those Android versions have the `libnativehelper.so`. Previous versions did not. We cannot load `libart.so` or `libdvm.so` instead, since previous Android versions restricted dynamically loading these "private" libraries.

It seems like the **recommended** solution to this issue on Android, is to tap into `JNI_OnLoad` and store the VM from there. This would mean that we would need to update the shared VM at runtime:
https://github.com/swiftlang/swift-java/blob/cc0ff8f6e478a05079c7206d7d4014d9c6eef776/Sources/SwiftJava/JavaKitVM/JavaVirtualMachine.swift#L223 

We need to figure out which library to dynamically load, that contains the `JNI_OnLoad` and can set the shared VM. My concern is since we already statically link the `SwiftJava` package _(the generated sources depend on it)_, it would not be OK to just also dynamically load it on the Java side, but I am **not** sure.

An alternative solution could be _(if this works in practice):_
1. Define a function in `SwiftJava` with `@_silgen_name("_swift_java_setSharedVM")`, which takes in a JavaVM and updates the shared one.
2. Call that function in `JNI_OnLoad` from the already dynamically loaded `SwiftRuntimeFunctions` target to set the shared VM.

madsodgaard:
`environment.FindClass` does not work when ran on native threads, such as is in the case of when we bridge `async` functions. As that function uses the caller class loader _(which of course is not the application loader in native threads)_

Therefore, we need the application class loader in order to cache non-system classes. As recommended in [the Android docs](https://developer.android.com/ndk/guides/jni-tips#faq:-why-didnt-findclass-find-my-class), we store the application class loader on `JNI_OnLoad`, and use that to reference any classes. We also need this `JNI_OnLoad` in the future to fix #419 - so this is a nice start, but this PR does not fix that issue.

This is needed to make #462 work

### DIFF
--- a/Sources/SwiftJavaRuntimeSupport/_JNIMethodIDCache.swift
+++ b/Sources/SwiftJavaRuntimeSupport/_JNIMethodIDCache.swift
@@ -51,13 +51,18 @@ public final class _JNIMethodIDCache: Sendable {
     if let jniClass = environment.interface.FindClass(environment, className) {
       clazz = environment.interface.NewGlobalRef(environment, jniClass)!
       self.javaObjectHolder = nil
-    } else if let javaClass = try? JNI.shared.applicationClassLoader.loadClass(
-      className.replacingOccurrences(of: "/", with: ".")
-    ) {
-      clazz = javaClass.javaThis
-      self.javaObjectHolder = javaClass.javaHolder
     } else {
-      fatalError("Class \(className) could not be found!")
+      // Clear any ClassNotFound exceptions from FindClass
+      environment.interface.ExceptionClear(environment)
+
+      if let javaClass = try? JNI.shared.applicationClassLoader.loadClass(
+        className.replacingOccurrences(of: "/", with: ".")
+      ) {
+        clazz = javaClass.javaThis
+        self.javaObjectHolder = javaClass.javaHolder
+      } else {
+        fatalError("Class \(className) could not be found!")
+      }
     }
 
     self._class = clazz


Sources/SwiftJavaRuntimeSupport/_JNIMethodIDCache.swift
```diff
@@ -39,11 +39,29 @@ public final class _JNIMethodIDCache: Sendable {
...
 
-  public init(environment: UnsafeMutablePointer<JNIEnv?>!, className: String, methods: [Method]) {
-    guard let clazz = environment.interface.FindClass(environment, className) else {
-      fatalError("Class \(className) could not be found!")
+  /// An optional reference to a java object holder
+  /// if we cached this class through the class loader
+  /// This is to make sure that the underlying reference remains valid
+  nonisolated(unsafe) private let javaObjectHolder: JavaObjectHolder?
+
+  public init(className: String, methods: [Method], isSystemClass: Bool) {
```
ktoso: I'm not sure we need the "is system class" boolean, it'd be fine to just fallback to the system loeader in any case if we didn't find in the env classloader perhaps? 

Falling back to more general classloaders is a pretty typical thing to do, I think we can do the same here.

ktoso: Alternatively, we could pass in a specific classloader to use, and then pass the application on when we create those shared "known" objects so we don't hit the fail->succeed path then

madsodgaard: I think the fallback solution sounds fine 👍

ktoso:
Lgtm, we may want to consider making thr init throwing but that may be annoying for now. Class not found do happen and perhaps should not crash the whole process tbh, depending what were tryint to find

ktoso (approved):

Sources/SwiftJavaRuntimeSupport/_JNIMethodIDCache.swift
```diff
@@ -39,11 +39,33 @@ public final class _JNIMethodIDCache: Sendable {
...
+  public init(className: String, methods: [Method]) {
+    let environment = try! JavaVirtualMachine.shared().environment()
+
+    let clazz: jobject
+    if let jniClass = environment.interface.FindClass(environment, className) {
+      clazz = environment.interface.NewGlobalRef(environment, jniClass)!
+      self.javaObjectHolder = nil
+    } else {
+      // Clear any ClassNotFound exceptions from FindClass
+      environment.interface.ExceptionClear(environment)
```
ktoso, line 56: ah right, good catch

Sources/SwiftJavaRuntimeSupport/_JNIMethodIDCache.swift
```diff
@@ -39,11 +39,33 @@ public final class _JNIMethodIDCache: Sendable {
     self._class!
   }
 
-  public init(environment: UnsafeMutablePointer<JNIEnv?>!, className: String, methods: [Method]) {
-    guard let clazz = environment.interface.FindClass(environment, className) else {
-      fatalError("Class \(className) could not be found!")
+  /// An optional reference to a java object holder
+  /// if we cached this class through the class loader
+  /// This is to make sure that the underlying reference remains valid
+  nonisolated(unsafe) private let javaObjectHolder: JavaObjectHolder?
```
ktoso, line 45: Why was the nonisolated unsafe necessary actually?

PR #465, opened 2025-11-28
Use thread environment instead for JNI methods

### LINKED ISSUE #157: Passing objects between threads

madsodgaard:
Currently, interacting with any `AnyJavaObject` is only possible from the thread that created it. This is quite limiting, as its often needed to pass these objects across threads. This limitation also affects features like the protocol bridging, as we use wrap-Java to callback to Java from Swift, and they will cross threads.

The reason this is, is that currently all the method call helpers use the `javaEnvironment` from `JavaObjectHolder`, which is the environment that was used when the object was created. I don't see a reason why we should use this environment to interact with the object, after its creation. We should really be using the current threads environment instead, otherwise the JVM will crash.

This PR also fixes #157

### DIFF
--- a/Tests/SwiftJavaTests/BasicRuntimeTests.swift
+++ b/Tests/SwiftJavaTests/BasicRuntimeTests.swift
@@ -82,6 +82,17 @@ class BasicRuntimeTests: XCTestCase {
     let nullString = String(fromJNI: nil, in: environment)
     XCTAssertEqual(nullString, "")
   }
+
+  func testCrossThreadAccess() async throws {
+    let environment = try jvm.environment()
+    let url = try URL("https://swift.org", environment: environment)
+    let string = await Task.detached {
+      // This should be called on a different thread
+      url.toString()
+    }.value
+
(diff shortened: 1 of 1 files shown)


ktoso (approved):
Seems reasonable, got any test that failed before and will now work to add here tho?

madsodgaard:
@ktoso I added a simple test

ktoso:
Neat, lgtm!

ktoso (approved):

PR #468, opened 2025-12-01
Introduce traits guarding dependencies that macros and -tool use

ktoso:
This is a specific approach intended to remove the need to split the swift-java repository into more repositories.

The intended goal is to allow libraries to depend on swift-java WITHOUT having to depend on: 
- swift-syntax (which @Java... macros depend on)
- any of the swift-java command line tool's dependencies

This design effectively means that a library just using the "no dependencies" pieces of swift-java would depend on it this way:

```
dependencies: [
    .package(
        ... swift-java ...
         // Disable all the dependencies SwiftJava Tool,
         // and macros are pulling otherwise
        traits: []
    ),
],
```

By "not enabling" the default traits, such library will NOT built swift-syntax and any of the other dependencies.

If another dependency, e.g. an application, also used SwiftJava and DID use the macros, these traits would become enabled -- as the developer intended -- however that is _not_ a problem for the "middle" libraries, such as Skip which would not have to enable any of these traits nor care about the dependencies.

--

In addition to that, all dependencies of swift-java WILL promise to support Android shortly, as soon as we figure out some organizational things about them.


ktoso:
We're not doing this for the time being and will spin up a new repo

PR #473, opened 2025-12-02
extract `CustomStringConvertible` as `toString()` and fix extensions

madsodgaard:
If a type conforms to `CustomStringConvertible` we will extract the `toString()` function as calling down to `description`.

Additional fixes:
* Fixes extensions not being able to extract variables.
* Adds any types that we conform to in extensions, to the Java decl.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
@@ -265,6 +265,8 @@ extension JNISwift2JavaGenerator {
     }
 
     try printSwiftInterfaceWrapper(&printer, protocolWrapper)
+  }
+
   private func printToStringMethods(_ printer: inout CodePrinter, _ type: ImportedNominalType) {
     let selfPointerParam = JavaParameter(name: "selfPointer", type: .long)
     let parentName = type.qualifiedName


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -228,12 +228,32 @@ extension JNISwift2JavaGenerator {
...
       printer.println()
       printDestroyFunction(&printer, decl)
     }
   }
 
+
+  private func printToStringMethod(_ printer: inout CodePrinter, _ decl: ImportedNominalType) {
+    // We assume that we are implementing `CustomStringConvertible`
+    // and therefore have a `description` property.
+    guard let descriptionVariable = decl.variables.first(where: { $0.name == "description" }) else {
```
ktoso: Hmmm, I'm a little on the fence here; this isn't the worst but also it's a bit guesswork.

I wonder if we shouldn't also be able to have a thunk that string-describes any type, then we'd use `String(describing: self)"` in the Swift thunk and that would be more correct and handle all types.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -228,12 +228,32 @@ extension JNISwift2JavaGenerator {
         printer.println()
       }
 
+      printToStringMethod(&printer, decl)
```
ktoso: Should we also add a toDebugString() which would call the DebugStringConvertible (`String(reflecting: self)`)?

Tests/JExtractSwiftTests/JNI/JNIExtensionTests.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import JExtractSwiftLib
+import Testing
+
+@Suite
+struct JNIExtensionTests {
```
ktoso, line 19: Please add an FFM test, I think this would either just work or require very minimal work. The README also has flipped both, so let's make the impl match that 😉

madsodgaard: There is one already

https://github.com/swiftlang/swift-java/blob/main/Tests/JExtractSwiftTests/ExtensionImportTests.swift

ktoso: Ah thank you.

I guess we really have to start doing "in the same file" tests that do both modes (like in https://github.com/swiftlang/swift-java/pull/475), it is getting hard to track what is covered 🤔

PR #477, opened 2025-12-03
jextract/ffm: Support [UInt8] parameters

ktoso:
These cause severe copying as we must copy the Java array into an off heap memory segment and then we must copy AGAIN into the Swift Array (!), so this is deeply inefficient and you should always prefer pointers of `Data` which we had support for already earlier.

Edit: Returning an array is now efficient, we're able to return [UInt8] which gets directly copied from native memory into an `byte[]` on the heap 🥳 In a single copy.

We can do better with other types, but this is the best we can hope to achieve for native arrays on Swift an Java side.

We'll do even better with other types by keeping the native memory retained etc. This is the most convenient though.

However, this is useful for convenience and we should offer this anyway.

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/ConversionStep.swift
+++ b/Sources/JExtractSwiftLib/FFM/ConversionStep.swift
@@ -22,7 +22,6 @@ enum ConversionStep: Equatable {
   /// The value being lowered.
   case placeholder
   
-  /// FIXME: Workaround for picking a specific placeholder value; We should resolve how method() works with lowered closures instead
   case constant(String)
 
   /// A reference to a component in a value that has been exploded, such as


Tests/JExtractSwiftTests/ByteArrayTests.swift
```diff
@@ -0,0 +1,83 @@
...
+            );
+          """,
+          """
+            /**
+             * Downcall to Swift:
+             * {@snippet lang=swift :
+             * public func acceptArray(array: [UInt8])
+             * }
+             */
+            public static void acceptArray(byte[] array) {
```
ktoso: this should get @Unsigned

Tests/JExtractSwiftTests/ByteArrayTests.swift
```diff
@@ -0,0 +1,83 @@
...
+              }
+            }
+            """
+        ],
+        /* expected Swift chunks */
+        [
+          """
+          @_cdecl("swiftjava_SwiftModule_acceptArray_array")
+          public func swiftjava_SwiftModule_acceptArray_array(_ array_pointer: UnsafeRawPointer, _ array_count: Int) {
+            acceptArray(array: [UInt8](UnsafeRawBufferPointer(start: array_pointer, count: array_count)))
```
ktoso, line 60: Today we end up with 2 copies because of this.

Instead, we should accept a raw pointer that we know is an swift array of the rtight type here.

ktoso:
For some performance follow up work: https://github.com/swiftlang/swift-java/issues/478 because for now passing arrays will not be ideasl

Sources/JExtractSwiftLib/FFM/CDeclLowering/FFMSwift2JavaGenerator+FunctionLowering.swift
```diff
@@ -348,6 +348,46 @@ struct CdeclLowering {
     case .composite:
       throw LoweringError.unhandledType(type)
 
+    case .array(let wrapped) where wrapped == knownTypes.uint8:
+      // Lower an array as 'address' raw pointer and 'count' integer
+      let cdeclParameters = [
+        SwiftParameter(
+          convention: .byValue,
+          parameterName: "\(parameterName)_pointer",
```
ktoso, line 356: We have naming inconsistency; these should be $ and not _ but seems we got this inconsistency already from the Data work -- lets fix it separately, since this made the PR kinda big

Sources/JExtractSwiftLib/FFM/ConversionStep.swift
```diff
@@ -21,6 +21,9 @@ import SwiftSyntaxBuilder
 enum ConversionStep: Equatable {
   /// The value being lowered.
   case placeholder
+  
+  /// FIXME: Workaround for picking a specific placeholder value; We should resolve how method() works with lowered closures instead
```
ktoso:
```suggestion
```

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -63,7 +63,11 @@ extension FFMSwift2JavaGenerator {
         """
       )
       printJavaBindingDowncallMethod(&printer, cFunc)
-      printParameterDescriptorClasses(&printer, cFunc)
+      if let outCallback = translated.translatedSignature.result.outCallback {
+        printUpcallParameterDescriptorClasses(&printer, outCallback)
+      } else { // FIXME: not an "else"
```
ktoso, line 68: should probably just fix this right away

ktoso: will do later, after all

PR #485, opened 2025-12-08
jextract: remove unsigned integers mode, remove wrap-guava mode

madsodgaard:
Per the forums thread, we decided to just use the underlying bit representation of any unsigned integral types. On top of that we then always use the `@Unsigned` annotation in JExtract, and remove the unsigned integers mode, since noone is really using wrap-guava.

### DIFF
--- a/Sources/SwiftJava/BridgedValues/JavaValue+Integers.swift
+++ b/Sources/SwiftJava/BridgedValues/JavaValue+Integers.swift
@@ -354,9 +354,13 @@ extension UInt64: JavaValue {
   public static var jvalueKeyPath: WritableKeyPath<jvalue, JNIType> { \.j }
 
   public func getJNIValue(in environment: JNIEnvironment) -> JNIType {
-    // On 32-bit, standard JDK jlong is defined as long long = Int64
-    // On 64-bit, standard JDK jlong is defined as long = Int
-    // On Android it's correctly marked as int64_t
+    // `jlong` is always 64-bit, no matter the system pointer size.
+    // Due to differences in JNI headers between Android, JDK and how Swift sees these type imports
+    // we have to handle this a bit differently.
+    //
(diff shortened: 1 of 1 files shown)


Sources/JExtractSwiftLib/Configuration+Extensions.swift
```diff
@@ -1,25 +0,0 @@
...
-import JavaTypes
-
-extension Configuration {
-  public var effectiveUnsignedNumericsMode: UnsignedNumericsMode {
-    switch effectiveUnsignedNumbersMode {
-    case .annotate: .ignoreSign
-    case .wrapGuava: .wrapUnsignedGuava
-    }
-  }
-}
```
ktoso, line 25: lgtm on record that yeah i think we can remove this mode; if we'd ever want it back it'll be easy to bring back from past commits

Sources/SwiftJava/BridgedValues/JavaValue+Integers.swift
```diff
@@ -353,10 +353,26 @@ extension UInt64: JavaValue {
 
   public static var jvalueKeyPath: WritableKeyPath<jvalue, JNIType> { \.j }
 
-  public func getJNIValue(in environment: JNIEnvironment) -> JNIType { JNIType(self) }
+  public func getJNIValue(in environment: JNIEnvironment) -> JNIType {
+    // On 32-bit, standard JDK jlong is defined as long long = Int64
+    // On 64-bit, standard JDK jlong is defined as long = Int
+    // On Android it's correctly marked as int64_t
```
ktoso: The comment could be clearer about the "why" this is ok: is it because it is 64bit on all Android versions we're supporting?

madsodgaard: I updated the docs, is that better?

ktoso: Thanks that makes more sense

Issue #484, opened 2025-12-07
Jextract CLI doesn't have a way to pass .swiftinterface files


abdulowork:
The jextract CLI doesn't allow passing .swiftinterface files. I built the CLI with `swift build --product swift-java`, produced the `.swiftinterface` with:
```
xcodebuild clean build -workspace Samples/SwiftJavaExtractFFMSampleApp/ -scheme SwiftJavaExtractFFMSampleApp -destination generic/platform=macOS -derivedDataPath dd SWIFT_EMIT_MODULE_INTERFACE=YES ARCHS=arm64
```
And tried importing it with:
```
./.build/arm64-apple-macosx/debug/swift-java jextract --swift-module MySwiftLibrary --input-swift dd/Build/Products/Debug/MySwiftLibrary.swiftmodule/arm64-apple-macos.swiftinterface --java-package com.example.swift --output-swift S1 --output-java J1
```

This capability seems to be intended from the [sources](https://github.com/swiftlang/swift-java/blob/2534f41d61bf118096655d7e60286912a8b36282/Sources/JExtractSwiftLib/Swift2Java.swift#L115-L116). With a minor tweak [here](https://github.com/swiftlang/swift-java/blob/2534f41d61bf118096655d7e60286912a8b36282/Sources/JExtractSwiftLib/Swift2Java.swift#L52), I was able to get jextract to accept the .swiftinterface file, and it produced a Java API similar to the API it makes from sources.

ktoso:
We originally were driving everything from interfaces and then moved to swift files, but yes, it is intended to allow either!

Would you want to submit your small tweak https://github.com/swiftlang/swift-java/blob/2534f41d61bf118096655d7e60286912a8b36282/Sources/JExtractSwiftLib/Swift2Java.swift#L52
here as a PR please? We can also add some test-case for this in Samples probably ... best discussed in a PR and I can help out finish it up

PR #486, opened 2025-12-08
PoC SwiftPM import for Swift Java 

timofeys1:
This PR is a showcase for [the discussion on the forum](https://forums.swift.org/t/swift-java-interop-community-call-november-27th/83138/9)

https://github.com/user-attachments/assets/4e01a873-bdd4-41a3-ba1f-3cc71ddc8bea


ktoso:
Thank you a lot, this is super promising! I need to give it a proper review, sorry I was help up by some other things the last days. Will get through it as soon as possible :)

Samples/SwiftJavaExtractFFMSampleApp/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+//
+// Copyright (c) 2024 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
```
ktoso, line 13: Random note on the gradle -> gradle.kts, yeah perhaps it would be beneficial to just move everything to the kotlin style, it hopefully would yield more reliable autocompletion (though i had mixed results in the past, it has a higher chance of working :)) 👍

Samples/SwiftJavaExtractFFMSampleApp/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+        products = listOf(product("SwiftPMImport", importedModules = setOf("SwiftTarget", "TargetWithAlgorithms"))),
+    )
+
+    localPackage(
+        path = projectDir,
+        products = listOf("MySwiftLibrary")
+    )
+
+    swiftJavaRepository.set(rootDir)
+}
```
ktoso, line 111: I think that's very neat and definitely something we'd be happy to support here as part of swift-java! 👍

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -108,6 +108,7 @@ extension FFMSwift2JavaGenerator {
       // Generated by swift-java
 
       import SwiftRuntimeFunctions
+      import \(self.swiftModuleName)
```
ktoso, line 111: Nitpick since not the core of the PR for discussion; Unsure why we'd need that one? We are "in" that module here I'd think?

Sources/JExtractSwiftLib/Swift2Java.swift
```diff
@@ -49,7 +49,7 @@ public struct SwiftToJava {
     let inputPaths = inputSwift.split(separator: ",").map { URL(string: String($0))! }
     log.info("Input paths = \(inputPaths)")
 
-    let allFiles = collectAllFiles(suffix: ".swift", in: inputPaths, log: translator.log)
+    let allFiles = inputPaths // collectAllFiles(suffix: ".swiftinterface", in: inputPaths, log: translator.log)
```
ktoso, line 52: Yeap, agreed that we do want to allow interface files here too 👍 
So a TODO to filter for .swift || .swiftinterface

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportExtension.kt
```diff
@@ -0,0 +1,252 @@
...
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.jetbrains.kotlin.gradle.plugin.mpp.apple.swiftimport.SwiftPMDependency.Remote.Repository
+import org.jetbrains.kotlin.gradle.plugin.mpp.apple.swiftimport.SwiftPMDependency.Remote.Version
+import java.io.File
+import java.io.Serializable
+import javax.inject.Inject
+
+abstract class SwiftImportExtension @Inject constructor(
```
ktoso, line 13: That's great, thank you! That's the kind of how-to-structure plugins I wasn't quite sure about, so this is a great pattern we can follow then

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportExtension.kt
```diff
@@ -0,0 +1,252 @@
...
+                        SwiftPMDependency.CinteropClangModule(it, product.platformConstraints)
+                    }
+                },
+                traits = traits,
+            )
+        )
+    }
+
+    // FIXME: check if this is actually correct
+    private fun inferPackageName(url: String) = url.split("/").last().split(".git").first()
```
ktoso, line 168: The more correct way would be to do the equivalent of `swift package describe | grep 'Name:' | head -n1` (JSON format is available as well)

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportPlugin.kt
```diff
@@ -0,0 +1,1049 @@
...
+
+        val swiftJavaIntermediates = project.layout.buildDirectory.dir("swiftJavaIntermediates")
+
+        val importedModuleToSourcesDump = importedModules.map { modules ->
+            modules.map { module ->
+                ImportedModuleInfo(
+                    module,
+                    swiftcAndLdDumpTask.get().moduleSwiftSources(archs.single(), module).get().asFile,
+                    swiftJavaIntermediates.map { it.dir(module).dir("SwiftBridges") }.get().asFile,
+                    swiftJavaIntermediates.map { it.dir(module).dir("JavaBridges") }.get().asFile,
```
ktoso, line 110: minor, for consistency with the rest of the project we should keep calling these "Wrappers" rather than "Bridges" (that's not a word we're using in the project, but yeah it's the same thing)

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportPlugin.kt
```diff
@@ -0,0 +1,1049 @@
...
+            it.aggregateModuleName.set(aggregateTargetName)
+            it.xcodebuildPlatform.set(targetPlatform)
+            it.xcodebuildSdk.set(targetSdk)
+            it.filesToTrackFromLocalPackages.set(computeLocalPackageDependencyInputFiles.flatMap { it.filesToTrackFromLocalPackages })
+            it.hasSwiftPMDependencies.set(project.provider { swiftPMImportExtension.spmDependencies.isNotEmpty() })
+            it.importedSwiftModules.set(emptySet())
+            it.swiftPMDependenciesCheckout.set(fetchSyntheticImportProjectPackages.map { it.swiftPMDependenciesCheckout.get() })
+            it.syntheticImportProjectRoot.set(bridgesPackage.flatMap { it.syntheticImportProjectRoot })
+            it.architectures.set(archs)
+        }
```
ktoso, line 164: Would love to hear your take on the build tasks @dschaefer2

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportPlugin.kt
```diff
@@ -0,0 +1,1049 @@
...
+    }
+
+    private fun checkoutSwiftPMDependencies() {
+        execOps.exec {
+            it.workingDir(syntheticImportProjectRoot.get().asFile)
+            it.commandLine(
+                "xcodebuild", "-resolvePackageDependencies",
+                "-scheme", SYNTHETIC_IMPORT_TARGET_MAGIC_NAME,
+                XCODEBUILD_SWIFTPM_CHECKOUT_PATH_PARAMETER, swiftPMDependenciesCheckout.get().asFile.path,
+                "-derivedDataPath", swiftPMDependenciesCheckoutLogs.get().asFile.path,
```
ktoso, line 685: Another one that we'd need to express using swiftpm commands

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportPlugin.kt
```diff
@@ -0,0 +1,1049 @@
...
+                "CODE_SIGN_IDENTITY=-",
+                "COMPILER_INDEX_STORE_ENABLE=NO",
+                "SWIFT_INDEX_STORE_ENABLE=NO",
+                "SWIFT_USE_INTEGRATED_DRIVER=NO",
+                // "SWIFT_EMIT_MODULE_INTERFACE=YES", ended up not being useful
+                "-IDEPackageSupportCreateDylibsForDynamicProducts=YES",
+            )
+            exec.environment(KOTLIN_SWIFTC_ARGS_DUMP_FILE_ENV, swiftArgsDump)
+            exec.environment(KOTLIN_LD_ARGS_DUMP_FILE_ENV, ldArgsDump)
+        }
```
ktoso, line 828: TODO: expressing this using swiftpm.

This is from the existing infra from kt, we likely dont need this part?

ktoso (commented):
Overall I think this is great, and definitely in the spirit of this project so I hope we can collaborate on polishing this up! We'd need this to use swiftpm for building and resolving though... 

Let's chat more with Doug and SwiftPM folks and use our community sync to go over this. 

Thank you for the effort! This is a fantastic start

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+SwiftThunkPrinting.swift
timofeys1: I changed the model of compilation slightly: the wrappers are compiled ([1](https://github.com/swiftlang/swift-java/pull/486/files/c6361201c1d0da8ed620c1688fc76ef84edab4ae#diff-1b7e857df0e3c5b49dafd8d92ad52873f1be8a5c8330c3d69c344b40aaa83c39R541), [2](https://github.com/swiftlang/swift-java/pull/486/files/c6361201c1d0da8ed620c1688fc76ef84edab4ae#diff-1b7e857df0e3c5b49dafd8d92ad52873f1be8a5c8330c3d69c344b40aaa83c39R152)) in a separate Swift module and this is where this import is coming from.

Sources/JExtractSwiftLib/Swift2Java.swift
timofeys1: Sure, I will do this in a separate PR and add some tests in #484

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportExtension.kt
timofeys1: JiC: This plugin violates some rules you would expect a modern Gradle plugin to follow. One particular Gradle feature that is easier to support from the start is the configuration cache. It significantly alters the execution model, so in a future proper implementation, supporting proper execution with a configuration cache could be a top priority.

swiftPMImportPlugin/src/main/kotlin/SwiftPMImportPlugin.kt
timofeys1: The `SWIFT_EXEC` and `KOTLIN_SWIFTC_ARGS_DUMP_FILE_ENV` parts could definitely be reworked for Swift Java; they are unnecessarily complicated. There are easier APIs for extracting Swift sources via the `swift package` CLIs. Perhaps one edge case that might require thought: if, in the Swift target, you generate sources using `buildCommand` and expect Java APIs for the generated code, then the external build system needs to run the build command somehow and know where to search for the outputs to pass to `jextract`.

The `LD` part could rely on assumptions, of course, but then it would be difficult to guarantee that the resulting binaries work correctly. Perhaps there is a more robust mechanism for understanding linker behavior in `swift build` and `xcodebuild` - I would be very curious to know!

Samples/SwiftJavaExtractFFMSampleApp/build.gradle.kts
timofeys1: I will work on a separate PR to showcase the `kts` scripts. Then you could give it a try and decide if you see the benefits for the Swift Java project.

PR #487, opened 2025-12-09
command: fix argument parser usage, remove workaround for bool? opts

ktoso:


ktoso:
FYI @madsodgaard removing a workaround for argument parser usage; the error message is pretty bad because these are property wrappers but yeah, we can do this. I also filed an issue and we can improve this in upstream lib as well https://github.com/apple/swift-argument-parser/issues/835

PR #489, opened 2025-12-10
jextract: pass arena from Swift instead of default interface for callbacks

madsodgaard:
Previously, we generated `default` methods for the Java interface methods that require a `SwiftArena`. That was a way to basically use a global arena, instead of passing in one from Swift. 

I think this PR is a better solution, where we pass in the global arena from Swift _(a global cached ref)_ and not generate the default methods, unless the user has explicitly requested it through the `allowGlobalArena` memory management mode.

### DIFF
--- a/Sources/SwiftJava/AnyJavaObject.swift
+++ b/Sources/SwiftJava/AnyJavaObject.swift
@@ -77,7 +77,7 @@ extension AnyJavaObject {
     return String(seq)
   }
 
-  /// The  mangled name for this java class
+  /// The mangled name for this java class
   public static var mangledName: String {
     "L\(fullJavaClassNameWithSlashes);"
   }


Sources/SwiftJavaRuntimeSupport/JNI.swift
```diff
@@ -15,13 +15,28 @@
 import SwiftJava
 import CSwiftJavaJNI
 
-final class JNI {
-  static var shared: JNI!
+public final class JNI {
+  public fileprivate(set) static var shared: JNI!
```
ktoso, line 26: Since making it public now, let's document when this is initialized

Sources/SwiftJavaRuntimeSupport/JNI.swift
```diff
@@ -15,13 +15,28 @@
...
 
-  let applicationClassLoader: JavaClassLoader
+  public let applicationClassLoader: JavaClassLoader
+  public let globalArena: JavaSwiftArena
```
ktoso: Hmm, this makes me realize the naming of the constant is wrong as well. "[Global](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/foreign/Arena.html#global())" arena is specifically an arena that does not track references and things are just immortal, this isn't a global arena, this is an "[auto arena](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/foreign/Arena.html#ofAuto())".

Perhaps the `GLOBAL_SWIFT_JAVA_ARENA` should become `DEFAULT_SWIFT_JAVA_AUTO_ARENA` perhaps. If easy enough, we can get rid of the constant as well; just upcall to get the `SwiftArena.ofAuto()` here, but I don't mind us keeping the constant if it makes our lives much easier.

Sources/SwiftJavaRuntimeSupport/generated/JavaSwiftArena.swift
```diff
@@ -0,0 +1,18 @@
...
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftJava
+
+@JavaInterface("org.swift.swiftkit.core.SwiftArena")
+public struct JavaSwiftArena {}
```
ktoso, line 18: I guess we don't support static methods on java interfaces huh?
Because we'd want the static ofAuto here basically.

Sources/SwiftJavaRuntimeSupport/JNI.swift
```diff
@@ -15,13 +15,28 @@
...
+
+    self.applicationClassLoader = try! JavaClass<JavaThread>(environment: environment).currentThread().getContextClassLoader()
+
+    // Find global arena
+    let swiftMemoryClass = environment.interface.FindClass(environment, "org/swift/swiftkit/core/SwiftMemoryManagement")!
+    let arenaFieldID = environment.interface.GetStaticFieldID(
+        environment,
+        swiftMemoryClass,
+        "GLOBAL_SWIFT_JAVA_ARENA",
+        "Lorg/swift/swiftkit/core/SwiftArena;"
```
ktoso: Can we get the name off `JavaSwiftArena`?

ktoso (commented):
Let's clean this up some more while at it, goal looks good tho

Sources/SwiftJavaRuntimeSupport/generated/JavaSwiftArena.swift
madsodgaard: I don't think we really need/want to add any methods here. This is just a wrapper type that is needed to make the typechecker happy, such that we can pass arenas into `wrap-java` methods.

Sources/SwiftJavaRuntimeSupport/JNI.swift
madsodgaard: I still think we want to use the constant, as this is the arena that is also used when you enable `allowGlobalArena` in `memoryManagementMode`. I think it makes sense to just always use the same one, whenever we need to reach for a "global" one? 

But sure, I think `DEFAULT` would be better in terms of naming

ktoso: Sounds good, I don't mind the constant.

To clarify the other thread, I was thinking if we could do `JavaClass<JavaSwiftArena>.ofAuto()` basically instead of writing the manual JNI dance

Sources/SwiftJavaRuntimeSupport/JNI.swift
ktoso: To clarigy, if we have to keep the manual JNI dance, i was wondering if we can do `"L"\(JavaClass<JavaSwiftArena>.fullJavaClassNameWithSlashes);" or add mangledName there if we don't already

Sources/SwiftJavaRuntimeSupport/generated/JavaSwiftArena.swift
ktoso: But we could have the static `ofAuto()` here, this way we'd not have to do any manual JNI in the other piece of code.

Sources/SwiftJava/AnyJavaObject.swift
```diff
@@ -77,6 +77,11 @@ extension AnyJavaObject {
     return String(seq)
   }
 
+  /// The  mangled name for this java class
```
ktoso:
```suggestion
  /// The mangled name for this java class
```

I wonder if that's correct for array but let's run with this for now!

PR #490, opened 2025-12-10
misc: use upstream `Subprocess`

madsodgaard:
Now that we have dropped Swift 6.0, we can use the upstream version of `Subprocess` instead of vending our own.


ktoso (approved):
Hope it'll just work™, let's see what CI says :)

ktoso:
Blocked on CI updating the macOS images

PR #449, opened 2025-11-15
jextract: add support for implementing Swift protocols in Java

madsodgaard:
Adds support for implementing Swift `protocols` in Java, by implementing the corresponding extracted `interface`.

This approach changes the JExtract build plugin to also use `wrap-java` to wrap the extracted sources. This means we can generate Swift wrappers that use these `@JavaInterface` classes to interop between Swift and Java.

Due to the fact that we need to build `SwiftKitCore` using `javac`, this requires that we disable the SwiftPM sandbox mode (using `--disable-sandbox`). Because of this, this feature is currently gated behind the `--enable-java-callbacks` option.

PR #491, opened 2025-12-10
jextract: add support for escaping closures

gevartosky:
This PR introduces support for escaping closures with primitive types only.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/EscapingClosures.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/EscapingClosures.swift
@@ -39,12 +39,6 @@ public class CallbackManager {
   }
 }
 
-// public func delayedExecution(closure: @escaping (Int64) -> Int64, input: Int64) -> Int64 {
-//   // In a real implementation, this might be async
-//   // For testing purposes, we just call it synchronously
-//   return closure(input)
-// }
-
 public class ClosureStore {
   private var closures: [() -> Void] = []
   
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+InterfaceWrapperGeneration.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+InterfaceWrapperGeneration.swift
@@ -121,7 +121,7 @@ extension JNISwift2JavaGenerator {
       )
     }
 
-        /// Generates a synthetic closure function translation.
+    /// Generates a synthetic closure function translation.
     /// This treats the closure as if it were a protocol with a single `apply` method,
     /// allowing it to use the same conversion infrastructure for optionals, arrays, etc.
     func generateSyntheticClosureFunction(


ARCHITECTURE.md
```diff
@@ -0,0 +1,355 @@
...
+
+Swift-Java provides bidirectional interoperability between Swift and Java. The main tool, `jextract`, generates bindings that allow Java code to call Swift APIs.
+
+```
+┌─────────────────┐      ┌─────────────────┐      ┌─────────────────┐
+│   Swift Source  │      │   Intermediate  │      │ Generated Code  │
+│   (.swift)      │ ───▶ │ Representation  │ ───▶ │ (.java, .swift) │
+└─────────────────┘      └─────────────────┘      └─────────────────┘
+      Parse              ImportedDecls           Generator (JNI/FFM)
+```
```
ktoso: We need to explain `wrap-java` here too then I suppose. Its workflow is

```
The other direction, when starting from a Java library and wanting to access it from Swift is handled by the `wrap-java` command, which uses a classpath as its input:


┌──────────────────┐      ┌─────────────────┐      ┌───────────────────────┐
│  Java classpath  │      │    Analysis     │      │ Generated Code        │
│   (.*class)      │ ───▶ │                 │ ───▶ │ (.swift with @macros) │
└──────────────────┘      └─────────────────┘      └───────────────────────┘
      Parse              ImportedDecls           Generator (JNI/FFM)

```

ARCHITECTURE.md
```diff
@@ -0,0 +1,355 @@
...
+    // 3. Generate output based on mode (JNI or FFM)
+    switch config.effectiveMode {
+    case .jni:
+        let generator = JNISwift2JavaGenerator(...)
+        try generator.generate()
+    case .ffm:
+        let generator = FFMSwift2JavaGenerator(...)
+        try generator.generate()
+    }
+}
```
ktoso: I would be concerned having actual code snippets in this guide -- they WILL become outdated and confusing. How about we just explain the same with words please?

ARCHITECTURE.md
```diff
@@ -0,0 +1,355 @@
+# Swift-Java Architecture Guide
```
ktoso: Can we please split out this md file addition from the rest of the PR? Many comments here that are should not affect the rest of the PR

CONTRIBUTING.md
```diff
@@ -5,6 +5,34 @@ your contribution to Apple and the community, and agree by submitting the patch
...
+- Key components and what each file does
+- Where to make changes for different types of features
+- Data flow examples
+
+This will give you the mental model needed to navigate the codebase effectively.
+
+### Development Setup
+
+```bash
+# Prerequisites: Java 22+ and Swift 6.0+
```
ktoso:
```suggestion
# Prerequisites: Java 25+ and Swift 6.2+
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -472,14 +472,14 @@ extension JNISwift2JavaGenerator {
   ) {
     let apiParams = functionType.parameters.map({ $0.parameter.renderParameter() })
 
-    printer.print(
+      printer.print(
```
ktoso: indent by mistake?

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -91,8 +91,10 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Non-escaping `Void` closures: `func callMe(maybe: () -> ())`                                      | ✅        | ✅   |
 | Non-escaping closures with primitive arguments/results: `func callMe(maybe: (Int) -> (Double))`   | ✅        | ✅   |
 | Non-escaping closures with object arguments/results: `func callMe(maybe: (JavaObj) -> (JavaObj))` | ❌        | ❌   |
-| `@escaping` closures: `func callMe(_: @escaping () -> ())`                                        | ❌        | ❌   |
-| Swift type extensions: `extension String { func uppercased() }`                      | ✅       | ✅  |
+| `@escaping` `Void` closures: `func callMe(_: @escaping () -> ())`                                 | ❌        | ✅   |
+| `@escaping` closures with primitive arguments/results: `func callMe(_: @escaping (String) -> (String))`       | ❌        | ✅   |
+| `@escaping` closures with custom arguments/results: `func callMe(_: @escaping (Obj) -> (Obj))`       | ❌        | ❌   |
+| Swift type extensions: `extension String { func uppercased() }`                      | 🟡       | 🟡  |ommit to support primitive escaping closures)
```
ktoso: This isn't changing anything about extensions, why this change?
This is added outside the table?

Sources/SwiftJavaRuntimeSupport/JNIClosureContext.swift
```diff
@@ -0,0 +1,52 @@
...
+/// let context = JNIClosureContext(globalRef: env.NewGlobalRef(env, javaCallback)!)
+/// let swiftClosure = {
+///     // context.globalRef is captured and kept alive
+///     env.CallVoidMethod(env, context.globalRef, methodID)
+/// }
+/// // When swiftClosure is released, context is deallocated and GlobalRef is deleted
+/// ```
+public final class JNIClosureContext: @unchecked Sendable { // @unchecked Sendable is needed because the globalRef is a jobject
+    /// The JNI GlobalRef to the Java closure object
+    public let globalRef: jobject
```
ktoso: This is just the existing `JavaObjectHolder`, no need to introduce a new type

Tests/JExtractSwiftTests/JNI/JNIEscapingClosureTests.swift
```diff
@@ -0,0 +1,145 @@
...
+    let simpleSource =
+      """
+      public func setCallback(callback: @escaping () -> Void) {}
+      """
+    
+    try assertOutput(input: simpleSource, .jni, .java, expectedChunks: [
+      """
+      public static class setCallback {
+        @FunctionalInterface
+        public interface callback extends AutoCloseable {
```
ktoso: This is illegal Java code and would not compile, a functional interface cannot have multiple abstract methods

ktoso: It seems the generated code is more correct actually, the test isn't passing and you need to remove those auto closable ideas from here

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -917,6 +943,70 @@ extension JNISwift2JavaGenerator {
...
+        let closureParameters = !parameters.isEmpty ? "\(names.joined(separator: ", ")) in" : ""
+        printer.print("return { \(closureParameters)")
+        printer.indent()
+        
+        printer.print("guard let env$ = try? JavaVirtualMachine.shared().environment() else {")
+        printer.indent()
+        printer.print("fatalError(\"Failed to get JNI environment for escaping closure call\")")
+        printer.outdent()
+        printer.print("}")
+        printer.print("")
```
ktoso: Generally we do a big print with """ strings rather than this style when possible

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
gevartosky: Apologies, a mistake when solving merge conflicts, thank you

madsodgaard:
Awesome work 👍 

I had a thought about how we could approach this feature. It shares a lot of similarities with the protocol callback work I did in #449 - both escaping closures and protocols are a way to achieve callbacks.

Consider the following API
```swift
func f(closure: @escaping (Int64, MyClass) -> Void)
``` 
technically we could express this as:
```swift
protocol f_closure {
    func apply(_ arg0: Int64, _ arg1: MyClass)
 }

func f(_ closureObject: f_closure) {
    self.f {
        closureObject.apply($0, $1)
    }
 }
```
then if we could pass the `f_closure` protocol and the new helper-function `f` through the work I did in the protocol PR, then it would generate
```java
interface f_closure {
  void apply(long arg0, MyClass arg1)
}

private static void f(f_closure_protocol closure_protocol) {
  // Downcall to Swift that takes in the protocol
  $f(closure_protocol)
}
```
and the required Swift wrappers needed to call that interface from Swift.
Now, we could detect that the protocol is used as a closure, and just attach `@FunctionalInterface` to it
```java
@FunctionalInterface
interface f_closure {
  void apply(long arg0, MyClass arg1)
}

public static void f($f_closure closure) {
  // Call the`f` that takes in the "f_closure" Swift protocol:
  f($f_closure);
 }
```

Then we basically have achieved support for escaping closures, by "proxying" them through protocols. This means we don't have to do double the work when we need to add new features to either protocols or closures. For example, this would give us support for non-primitive types right away, since protocols already support them.

Let me know what you think!

gevartosky:
Hey, I'm back. Happy new year @madsodgaard.
Thank you for the input! I agree.

I'll look into it

madsodgaard:
awesome that you looked into sharing some functionality between closures and protocols! I think it helps a ton on the codebase and future features.

_Cross-posting this from slack:_
I actually thought we had to codegen the "fake" protocols in Swift and pass them through the JNI tool, but I guess you are just wrapping the already extracted "non-escaping" closures generated  - that ofc. means we are limited to the features of the non-escaping closures. So we would need to extend non-escaping closures to support optionals, async etc. If we could codegen the "fake" protocol and pass it through the JNI generator, we would already have support for optionals, async and so on, since it then just relies on the supported features for "normal" functions and protocols". Maybe we should think about doing that? What do you think? @gevartosky @ktoso - otherwise this might be a good start, and we should look into extend non-escaping closures instead, just trying to think how we could share as much code as possible.
So passing this "fake" protocol through the JNI generator:
```swift
protocol f_closure {
    func apply(_ arg0: Int64, _ arg1: MyClass)
 }
```

gevartosky:
@madsodgaard  thank you for the suggestion

I had an overall understanding of your suggestion and pushed a change.

Then I proceeded to trying to make **async** which was somewhat available thanks to the work on protocols but still required some changes and converting results back and forth. 
I am traveling tomorrow and will be back Jan 19, so I ended up making these changes on a separate branch in case we decided to merge the current PR.
https://github.com/gevartosky/swift-java/compare/feature/escaping-closures...gevartosky:swift-java:feature/escaping-closures-async?expand=1

ktoso:
Sorry for the delay here, I was traveling and ended up sick -- this is looking good, I'll just give it a final look and will merge, thank you

Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/EscapingClosures.swift
```diff
@@ -0,0 +1,83 @@
...
+  public func triggerIntCallback(value: Int64) -> Int64? {
+    return intCallback?(value)
+  }
+}
+
+// public func delayedExecution(closure: @escaping (Int64) -> Int64, input: Int64) -> Int64 {
+//   // In a real implementation, this might be async
+//   // For testing purposes, we just call it synchronously
+//   return closure(input)
+// }
```
ktoso: Enable or remove?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+InterfaceWrapperGeneration.swift
```diff
@@ -105,6 +121,34 @@ extension JNISwift2JavaGenerator {
       )
     }
 
+        /// Generates a synthetic closure function translation.
```
ktoso:
```suggestion
    /// Generates a synthetic closure function translation.
```

ktoso (approved):

PR #493, opened 2025-12-11
make `SwiftJava` a dynamic lib and add class lookup fallbacks

### LINKED ISSUE #419: `swift-java` does not work on Android API 28-30, due to missing JavaVM JNI symbols

madsodgaard:
Converts `SwiftJava` to a dynamic lib, that will be loaded in the JExtract libs. This will always us to tap into `JNI_OnLoad` to get the shared JVM, and also cache the application class loader, so that we can use that as a fallback in the `withJNIClass`, if for example running on a native thread.

Resolves #419

### DIFF
--- a/Sources/SwiftJava/AnyJavaObject.swift
+++ b/Sources/SwiftJava/AnyJavaObject.swift
@@ -134,6 +134,8 @@ extension AnyJavaObject {
     _ body: (jclass) throws -> Result
   ) throws -> Result {
     let resolvedClass = try classLoader.findClass(fullJavaClassName)
+    // OK to force unwrap, as classLoader will throw ClassNotFoundException
+    // if the class cannot be found.
     return try body(resolvedClass!.javaThis)
   }
 
--- a/Samples/SwiftAndJavaJarSampleLib/Package.swift
+++ b/Samples/SwiftAndJavaJarSampleLib/Package.swift
@@ -63,7 +63,9 @@ let package = Package(
     .target(
       name: "MySwiftLibrary",
       dependencies: [
-        .product(name: "SwiftRuntimeFunctions", package: "swift-java"),
+        .product(name: "SwiftJava", package: "swift-java"),
+        .product(name: "CSwiftJavaJNI", package: "swift-java"),
+        .product(name: "SwiftRuntimeFunctions", package: "swift-java")
       ],
       exclude: [
         "swift-java.config",
--- a/Tests/JExtractSwiftTests/JNI/JNIEnumTests.swift
+++ b/Tests/JExtractSwiftTests/JNI/JNIEnumTests.swift
@@ -50,6 +50,7 @@ struct JNIEnumTests {
           @SuppressWarnings("unused")
           private static final boolean INITIALIZED_LIBS = initializeLibs();
           static boolean initializeLibs() {
+            System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
             System.loadLibrary(LIB_NAME);
             return true;
           }
--- a/Tests/JExtractSwiftTests/JNI/JNIModuleTests.swift
+++ b/Tests/JExtractSwiftTests/JNI/JNIModuleTests.swift
@@ -53,6 +53,7 @@ struct JNIModuleTests {
         static final String LIB_NAME = "SwiftModule";
       
         static {
+          System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
           System.loadLibrary(LIB_NAME);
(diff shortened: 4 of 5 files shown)


Sources/SwiftJava/AnyJavaObject.swift
```diff
@@ -102,23 +102,39 @@ extension AnyJavaObject {
...
   private static func _withJNIClassFromCustomClassLoader<Result>(
     _ classLoader: JavaClassLoader,
     in environment: JNIEnvironment,
-    _ body: (jclass?) throws -> Result
+    _ body: (jclass) throws -> Result
   ) throws -> Result {
-    let resolvedClass = try? classLoader.findClass(fullJavaClassName)
-    return try body(resolvedClass?.javaThis)
+    let resolvedClass = try classLoader.findClass(fullJavaClassName)
+    return try body(resolvedClass!.javaThis)
```
ktoso, line 139: mini comment here as we do the `!` please? cannot ever be null because it'll throw a not found?

Sources/SwiftJavaRuntimeSupport/generated/JavaSwiftArena.swift
```diff
@@ -16,3 +16,11 @@ import SwiftJava
 
 @JavaInterface("org.swift.swiftkit.core.SwiftArena")
 public struct JavaSwiftArena {}
+
+extension JavaSwiftArena {
+  /// A cache for the default auto arena found in SwiftKitCore
+  public static internal(set) var defaultAutoArena: JavaSwiftArena = {
+    let swiftMemoryClass = try! JavaClass<SwiftJavaRuntimeSupport.JavaSwiftMemoryManagement>()
+    return swiftMemoryClass.defaultSwiftJavaAutoArena
```
ktoso, line 24: great, thx

Sources/SwiftJava/JNI.swift
```diff
@@ -21,36 +20,28 @@ import CSwiftJavaJNI
...
-    let localObject = environment.interface.GetStaticObjectField(environment, swiftMemoryClass, arenaFieldID)!
-    self.defaultAutoArena = JavaSwiftArena(javaThis: localObject, environment: environment)
-    environment.interface.DeleteLocalRef(environment, localObject)
   }
 }
 
-// Called by generated code, and not automatically by Java.
-public func _JNI_OnLoad(_ javaVM: JavaVMPointer, _ reserved: UnsafeMutableRawPointer) {
+@_cdecl("JNI_OnLoad")
+func SwiftJava_JNI_OnLoad(javaVM: JavaVMPointer, reserved: UnsafeMutableRawPointer) -> jint {
```
ktoso, line 44: don't we need this "lower", in a lib that does not have macros?

ktoso: note, we'll do that later

PR #497, opened 2025-12-16
jextract: extract swift func docs as Java docs

madsodgaard:
Adds support for extracting any Swift function documentation as javadoc.

It is just a string/regex parser, no fancy CLR parser or anything, which I think is fine. We can always improve this (markdown support etc), but at least this is very nice for any Java consumers, for now.


Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -387,16 +386,29 @@ extension FFMSwift2JavaGenerator {
...
+    } else {
+      printer.print(
+        """
+        /**
+         * Downcall to Swift:
+         * {@snippet lang=swift :
+         * \(decl.signatureString)
+         * }
+         */
+        """
```
ktoso: I think we likely want to keep the "downcall to ..." part of the docs always; it's a quick way to verify if this is calling what you think it is, and if we have any bad codegen etc

ktoso: In other words, I don't think there should be an if/else here, can we centralize "make the docs, as good as possible" in some docs type that may use the parser inside?

Sources/JExtractSwiftLib/SwiftDocumentationParsing.swift
```diff
@@ -0,0 +1,202 @@
...
+    if let summary = summary {
+      printer.print(" * \(summary)")
+    }
+
+    if let discussion = discussion {
+      let paragraphs = discussion.split(separator: "\n\n")
+      for paragraph in paragraphs {
+        printer.print(" * <p>")
+        printer.print(" * \(paragraph)")
+        printer.print(" * </p>")
```
ktoso: We could use the OpenJDK style here... which is just ```printer.print(" * <p>\(paragraph)")``` and that'll do the right thing. Examples https://github.com/srisatish/openjdk/blob/master/jdk/src/share/classes/java/util/LinkedList.java

Do we need to replace code quotes with `<pre>`?

Sources/JExtractSwiftLib/SwiftDocumentationParsing.swift
```diff
@@ -0,0 +1,202 @@
...
+      }
+    }
+
+    for parameter in parameters {
+      printer.print(" * @param \(parameter.name) \(parameter.description)")
+    }
+
+    if let returns = returns {
```
ktoso:
```suggestion
    if let returns {
```

Sources/JExtractSwiftLib/SwiftDocumentationParsing.swift
```diff
@@ -0,0 +1,202 @@
...
+enum SwiftDocumentationParser {
+  private enum State {
+    case summary
+    case discussion
+    case parameter(Int)
+    case returns
+  }
+
+  // Capture Groups: 1=Tag, 2=Arg(Optional), 3=Description
+  private static let tagRegex = try! NSRegularExpression(pattern: "^-\\s*(\\w+)(?:\\s+([^:]+))?\\s*:\\s*(.*)$")
```
ktoso, line 40: I'd leave a TODO to use Regex but we know it has some performance issues potentially...

Sources/JExtractSwiftLib/SwiftDocumentationParsing.swift
```diff
@@ -0,0 +1,202 @@
...
+    // Remove any trailing newlines in discussion
+    while doc.discussion?.last == "\n" {
+      doc.discussion?.removeLast()
+    }
+
+    return doc
+  }
+
+  /// This is a test
+  private static func appendLineToState(_ state: State, line: String, doc: inout SwiftDocumentation) {
```
ktoso, line 131: This is a test, should this live in this file then?

Tests/JExtractSwiftTests/MethodImportTests.swift
```diff
@@ -91,10 +91,7 @@ final class MethodImportTests {
       expected:
         """
         /**
-         * Downcall to Swift:
-         * {@snippet lang=swift :
-         * public func helloWorld()
-         * }
```
ktoso, line 97: Noted above, but I think we want to keep these

ktoso (approved):
Nice thank you!

PR #498, opened 2025-12-16
jextract: add support for desugared optional on Android 23

madsodgaard:
Certain Java APIs have been backported to older Android versions through desugaring, however that means that the underlying class names are different, when we need to call out to JNI.

This PR changes the `@JavaClass` macro, to use the desugared name if we are running on Android.


Sources/SwiftJavaMacros/JavaClassMacro.swift
```diff
@@ -73,13 +73,13 @@ extension JavaClassMacro: MemberMacro {
     let superclass = specifiedSuperclass ?? "JavaObject"
 
     // Check that the class name is fully-qualified, as it should be.
-    let className = classNameSegment.content.text
-    if className.firstIndex(of: ".") == nil {
-      throw MacroErrors.classNameNotFullyQualified(className)
+    let specifiedClassName = classNameSegment.content.text
+    if specifiedClassName.firstIndex(of: ".") == nil {
+      throw MacroErrors.classNameNotFullyQualified(specifiedClassName)
```
ktoso: nitpick but just "className" was fine here imho

Sources/SwiftJavaMacros/JavaClassMacro.swift
```diff
@@ -94,7 +94,13 @@ extension JavaClassMacro: MemberMacro {
     let classNameAccessSpecifier = isSwiftClass ? "open" : "public"
     members.append("""
       /// The full Java class name for this Swift type.
-      \(raw: classNameAccessSpecifier) \(raw: fullJavaClassNameMemberModifiers) var fullJavaClassName: String { \(literal: className) }
+      \(raw: classNameAccessSpecifier) \(raw: fullJavaClassNameMemberModifiers) var fullJavaClassName: String {
+        #if os(Android)
+          AndroidSupport.androidDesugarClassNameConversion(for: "\(raw: specifiedClassName)")
```
ktoso: yeah it has to be runtime like that... the os guard is good.

Kinda would like a test for this but we don't run tests on android yet hm

ktoso (approved):
Looks good to me

ktoso (approved):

PR #501, opened 2025-12-19
fix duplicated fields from superclass

diederich:
in case of duplicated properties, e.g. from a base class it was generating duplicate property definitions, e.g.:

```
extension JavaClass<PrivateKey> {
  @JavaStaticField(isFinal: true)
  public var serialVersionUID: Int64

  @JavaStaticField(isFinal: true)
  public var serialVersionUID: Int64
}

```
(see e.g. https://docs.oracle.com/javase/8/docs/api/java/security/PrivateKey.html)

which then fails the swift build.

Analog to methods, this now skips the property in case it wasn't directly from the class itself.


ktoso:
Tested locally while we debug the macOS CI problem -- lgtm, thank you!

PR #502, opened 2025-12-19
fix `init` naming mismatch

diederich:
Java's `init` methods are bridged as is and made swift compatible via the backticks.
At runtime though, this is forgotten and the java methods are looked up _with_ the backticks.

This approach of fixing adds a `@JavaMethod("init")` annotation, which fixes the issue.

(Though I'm not sure if this shouldn't be fixed in the generated macro code. Open for suggestions!)


diederich:
closing, superseded by #514 - Thanks Konrad! 🙏

PR #505, opened 2025-12-21
Improve JAVA_HOME detection with SDKMAN and PATH Fallback

### LINKED ISSUE #393: Improve locating JAVA_HOME to ask sdkman and other tools in addition to java_home
We should try more fallbacks if JAVA_HOME, and java_home cannot find a java.

Swift developers may not be too familiar with installing JDKs so we should try harder to find one when we fail using the most common methods

amanthatdoescares:
Adds extra fallbacks to locate a JDK when JAVA_HOME is not set.

In addition to existing behavior, the tool now:
- Detects Java installed via SDKMAN
- Falls back to java on PATH (resolving symlinks)

Tested on Linux (x86_64).

Fixes #393

### DIFF
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -1,9 +1,7 @@
 // swift-tools-version: 6.0
 
 import PackageDescription
-
-import class Foundation.FileManager
-import class Foundation.ProcessInfo
+import Foundation
 
 // Note: the JAVA_HOME environment variable must be set to point to where
 // Java is installed, e.g.,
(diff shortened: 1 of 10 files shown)


ktoso:
We have the same kind of logic in all Package.swift and they all need to get the same treatment. Sadly there’s no great way to share code with Package.swift so we’ll have to copy paste the code I fear 😢

ktoso:
I applied the change to all the other places, we sadly don't have a good way to share code between those Package.swift files...

The macOS Swift test CI will fail but I verified locally and it is OK -- we have some problem with CI infrastructure I'm debuggigng.

ktoso (approved):

ktoso:
Thank you @amanthatdoescares !

Sorry it took a moment to review, needed a real offline holiday over xmas :)

PR #506, opened 2025-12-26
jextract: extract `Foundation.Date`

madsodgaard:
Adds support to JNI mode for extracting `Foundation.Date` just like `Data`.

We also generate two helper methods on `Date` for converting back and forth between `java.time.Instant`, which represents the same thing semantically, but they do not contain the same precision. So we also alert the developer of this.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -727,7 +727,7 @@ extension JNISwift2JavaGenerator {
     printer.print(
       """
       /**
-       * Initializes a date from a Java {@link java.time.Instant}.
+       * Initializes a Swift {@code Foundation.Date} from a Java {@link java.time.Instant}.
        *
        * <h3>Warning: Precision Loss</h3>
        * <p>


Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -69,6 +69,7 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Existential parameters `f(x: any (A & B)) `                                          | ❌        | ✅   |
 | Existential return types `f() -> any Collection `                                    | ❌        | ❌   |
 | Foundation Data and DataProtocol: `f(x: any DataProtocol) -> Data`                   | ✅        | ❌   |
+| Foundation Date: `f(date: Date) -> Date`                                             | ❌        | ✅   |
```
ktoso, line 72: because it's an unusual conversion I think we have to add actual docs about it

ktoso:
I keep thinking about this from an overall project perspective, it is a bit random to bless a few types, but yes this one would work kinda well...

madsodgaard:
@ktoso Hm, imo this project exists to provide a seamless Swift/Java interop story. Since we are generating Java code meant to be used on Java platforms, perhaps also by Java developers, I think it makes sense to bridge any matching types, such as `UUID`, `Date`,`URL` etc... 

The alternative is that the developer has to use a generated type `Swift.Date` like we do for `Data` in FFM. That will probably often (for these kind of "utility" types") result in the developer having to convert their Java types to the Swift wrapped types. I understand why it makes sense for `Data` because of memory copying etc, but not sure that its the same motivation for the above types.

ktoso:
> Hm, imo this project exists to provide a seamless Swift/Java interop story.

Primarily it must be: correct, safe, and then seamless. We must not sacrifice correctness just because it's convenient. The conversion is lossy, and could result in bugs in practice, e.g. if you're trying to key anything off the date.

Such echo function showcases how this is a lossy conversion:

```swift
import Foundation

public func echoDate(_ date: Date) -> Date {
  return date
}
```

and in Java + jextracted Date-as-Instant:

```java
        Instant now = Instant.now();
        Instant back = MySwiftLibrary.echoDate(now);

        if (!now.equals(back)) {
            System.out.println("NOT EQUAL! " + now + " != " + back);
        } else {
            System.out.println("EQUAL");
        }
```

gives:

```bash
NOT EQUAL! 2026-01-14T05:45:55.764682Z != 2026-01-14T05:45:55.764682054Z
DONE.
```

That's not something we should be just doing by default.

Instead I propose:

- import `Foundation.Date` as a wrapped type, same as we do Data
- add a method on it called `Instant toJavaInstant()` which performs the conversion so people don't have to do it manually.

Sounds good?

ktoso (changes requested):
Yeah after all let's not do the automatic conversion because it could be a source of equality bugs

madsodgaard:
@ktoso I changed it to extract `Date` as its own class, and also added generation of helper methods for converting back and forth between `Instant`

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -694,4 +708,57 @@ extension JNISwift2JavaGenerator {
...
+          long nanos = Math.round((this.getTimeIntervalSince1970() - seconds) * 1_000_000_000);
+          return java.time.Instant.ofEpochSecond(seconds, nanos);
+      }
+      """
+    )
+    printer.println()
+    printer.print(
+      """
+      /**
+       * Initializes a date from a Java {@link java.time.Instant}.
```
ktoso: "Initializes a Swift Foundation.Date" maybe?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -689,4 +703,57 @@ extension JNISwift2JavaGenerator {
...
```
```suggestion
       * Initializes a Swift {@code Foundation.Date} from a Java {@link java.time.Instant}.
```

maybe?

PR #511, opened 2025-12-30
Try loading Swift libraries from JAR resources

### LINKED ISSUE #510: Load Swift Libraries manually instead of System.loadLibrary
I'm trying to use swift-java to include swift code in a java image processing application.
https://github.com/PhotonVision/photonvision/pull/2158

I'm using FFM mode.
I've got it building when using ./gradlew run but it fails when packaging into a jar.

The generated java code uses System.loadLibrary but this doesn't work from a jar because
1. /usr/bin/swift is not on the java path
2. SwiftJava and SwiftRuntimeFunctions are still inside the jar

I think a solution would be to extract the libs in the initializer and use System.load.

TheTripleV:
closes #510 

For swiftCore:
1. Try loading by name
2. Try loading from jar
3. Try loading from absolute path

For the rest:
1. Try loading by name
2. Try loading from jar

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
@@ -194,46 +194,10 @@ extension FFMSwift2JavaGenerator {
         @SuppressWarnings("unused")
         private static final boolean INITIALIZED_LIBS = initializeLibs();
         static boolean initializeLibs() {
-            // Load swiftCore: try path, then JAR, then /usr/lib/swift on macOS
-            try {
-                System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_CORE);
-            } catch (Throwable e) {
-                try {
-                    SwiftLibraries.loadResourceLibrary(SwiftLibraries.LIB_NAME_SWIFT_CORE);
-                } catch (Throwable e2) {
-                    if (PlatformUtils.isMacOS()) {
-                        try {
-                            System.load("/usr/lib/swift/libswiftCore.dylib");
-                        } catch (Throwable e3) {
-                            throw new RuntimeException("Failed to load swiftCore from java.library.path, JAR, and /usr/lib/swift", e3);
-                        }
-                    } else {
-                        throw e2;
-                    }
-                }
-            }
-
-            // Load SwiftJava: try path, then JAR
-            try {
-                System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
-            } catch (Throwable e) {
-                SwiftLibraries.loadResourceLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
-            }
-
-            // Load SwiftRuntimeFunctions: try path, then JAR
-            try {
-                System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_RUNTIME_FUNCTIONS);
-            } catch (Throwable e) {
-                SwiftLibraries.loadResourceLibrary(SwiftLibraries.LIB_NAME_SWIFT_RUNTIME_FUNCTIONS);
-            }
-
-            // Load module library: try path, then JAR
-            try {
-                System.loadLibrary(LIB_NAME);
-            } catch (Throwable e) {
-                SwiftLibraries.loadResourceLibrary(LIB_NAME);
-            }
-
+            SwiftLibraries.loadLibraryWithFallbacks(SwiftLibraries.LIB_NAME_SWIFT_CORE);
+            SwiftLibraries.loadLibraryWithFallbacks(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
+            SwiftLibraries.loadLibraryWithFallbacks(SwiftLibraries.LIB_NAME_SWIFT_RUNTIME_FUNCTIONS);
+            SwiftLibraries.loadLibraryWithFallbacks(LIB_NAME);
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


TheTripleV:
This PR does work for me for when I launch my app from a jar.

The java project building swift wrappers must add the build binaries as resources (see https://github.com/TheTripleV/photonvision/blob/mac/photon-apple/build.gradle#L73).
They must be at the root level of the jar.

Then, the generated wrappers will load binaries from jar if they are not available on the java library path.

--
For swiftCore, the added behavior of falling back to /usr/bin/swift means that an application jar using wrappers built with swift-java:
1. doesn't have to vendor swiftCore.dylib
2. can be launched using `java -jar [file]` without any additional library paths passed in as parameters to the jvm



For reference, the directory structure of my built jar:
```bash
LICENSE
META-INF
ResourceInformation.json
boarddefs
calibration
com
controller.proto
edu
geometry2d.proto
geometry3d.proto
io
jakarta
javax
jetty-dir.css
kinematics.proto
kotlin
lib
libPhotonAppleLibrary.dylib
libSwiftJava.dylib
libSwiftRuntimeFunctions.dylib
models
org
osx
pabeles
photon.proto
plant.proto
soc_mapping.properties
spline.proto
sqlite-jdbc.properties
system.proto
testimages
tinylog.properties
trajectory.proto
us
web
wpimath.proto
```

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -194,10 +194,46 @@ extension FFMSwift2JavaGenerator {
...
-            System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_RUNTIME_FUNCTIONS);
-            System.loadLibrary(LIB_NAME);
+            // Load swiftCore: try path, then JAR, then /usr/lib/swift on macOS
+            try {
+                System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_CORE);
+            } catch (Throwable e) {
+                try {
+                    SwiftLibraries.loadResourceLibrary(SwiftLibraries.LIB_NAME_SWIFT_CORE);
+                } catch (Throwable e2) {
+                    if (PlatformUtils.isMacOS()) {
```
ktoso: We should also handle linux here; it would be under `/usr/lib/swift/linux/libswiftCore.so`

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -194,10 +194,46 @@ extension FFMSwift2JavaGenerator {
...
+                        try {
+                            System.load("/usr/lib/swift/libswiftCore.dylib");
```
ktoso: Make this a function like `SwiftLibraries.libSwiftCorePath()` and inside it make the `is....` platform checks, and return the appropriate path for the platform.

The func there should then find the right prefix (`/usr/lib/swift/linux/`, `/usr/lib/swift`) based on platform, and do the same to the suffix (`so`, `dylib`). We don't want those paths just hardcoded in many places like this

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator.swift
```diff
@@ -194,10 +194,46 @@ extension FFMSwift2JavaGenerator {
...
+                    }
+                }
+            }
+
+            // Load SwiftJava: try path, then JAR
+            try {
+                System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
+            } catch (Throwable e) {
+                SwiftLibraries.loadResourceLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
+            }
```
ktoso: This is pretty repetitive, please make a function inside `SwiftLibraries` that does the "attempt path, then resource" along with nice error reporting

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftLibraries.java
```diff
@@ -66,12 +86,19 @@ public static void loadResourceLibrary(String libname) {
                 throw new RuntimeException("Expected library '" + libname + "' ('" + resourceName + "') was not found as resource!");
             }
 
-            // TODO: we could do an in memory file system here
```
ktoso, line 69: The TODO is still valid, don't remove it

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftLibraries.java
```diff
@@ -41,21 +41,41 @@ public final class SwiftLibraries {
...
     }
 
     // ==== ------------------------------------------------------------------------------------------------------------
     // Loading libraries
 
-    public static void loadResourceLibrary(String libname) {
+    // Cache of already-loaded libraries to prevent duplicate extraction
+    private static final java.util.Map<String, File> loadedLibraries = new java.util.HashMap<>();
+
+    public static synchronized void loadResourceLibrary(String libname) {
```
ktoso: Please don't use `synchronized` this'll bottleneck loading libraries while there's no reason to do so at all -- we can use the loadded libraries map but please back it with a `ConcurrentHashMap` for safety and remove the synchronized.

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftLibraries.java
```diff
@@ -66,12 +86,19 @@ public static void loadResourceLibrary(String libname) {
...
-            // TODO: we could do an in memory file system here
+            // Extract to temp file
             File tempFile = File.createTempFile(libname, "");
             tempFile.deleteOnExit();
             Files.copy(libInputStream, tempFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
 
             System.load(tempFile.getAbsolutePath());
+
+            // Cache the loaded library
+            loadedLibraries.put(libname, tempFile);
```
ktoso: make use of concurrent hashmap's computeIfAbsent here please

SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftLibraries.java
```diff
@@ -41,21 +41,41 @@ public final class SwiftLibraries {
             : Boolean.getBoolean("swiftkit.auto-load-libraries");
 
     @SuppressWarnings("unused")
-    private static final boolean INITIALIZED_LIBS = loadLibraries(false);
+    private static final boolean INITIALIZED_LIBS = AUTO_LOAD_LIBS ? loadLibraries(false) : true;
```
ktoso, line 44: that seems right, thanks

ktoso (changes requested):
Thank you! 

Needs some cleanup, thank you for the work though. Marking as needs changes until comments addressed

ktoso:
Thanks!

Issue #521, opened 2026-01-15
System.loadLibrary uses the module name instead of the library (product) name


sidepelican:
The name of the dynamic library (`.so` file) generated by SwiftPM depends on the `name` in the `.library` product.

```swift
.library(
    name: "my-module", // This name is used for the file: libmy-module.so
    type: .dynamic,
    targets: ["MyModule"]
)
```

However, SwiftJava currently uses the **module name** for `System.loadLibrary`. In the example above, it tries to load `libMyModule.so`, which causes an error because the file does not exist.

```swift
public final class MySwiftClass implements JNISwiftInstance {
  static final String LIB_NAME = "MyModule"; // Should be "my-module"
  
  static boolean initializeLibs() {
      System.loadLibrary(SwiftLibraries.LIB_NAME_SWIFT_JAVA);
      System.loadLibrary(LIB_NAME); // Error
      return true;
  }
}

```

**Expected Behavior:**
I think we should have one of the following:

1. Use the **library (product) name** for code generation instead of the module name.
2. Or, show an error if the library name and module name are different (or if there are multiple modules).

ktoso:
The right thing here is probably sticking to product name, if you'd like to pick this up -- thank you

PR #525, opened 2026-01-19
Migrate build scripts from Groovy to Kotlin

abdulowork:
In #486, we discussed that the Kotlin build scripts might offer better DevX. Logic-wise I tried to introduce as little diff as possible, but I moved the buildSrc code into the BuildLogic.

Please let me know if you see the DX benefits from the migration!

### DIFF
--- a/BuildLogic/src/main/kotlin/utilities/registerJextractTask.kt
+++ b/BuildLogic/src/main/kotlin/utilities/registerJextractTask.kt
@@ -19,12 +19,11 @@ import org.gradle.api.tasks.Exec
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.kotlin.dsl.register
 import java.io.File
-import java.nio.file.Files
 
-private fun Project.swiftProductsWithJExtractPlugin(): List<String> = swiftPMPackage().targets.filter {
+private fun Project.swiftTargetsWithJExtractPlugin(): List<String> = swiftPMPackage().targets.filter {
     it.productDependencies.contains("JExtractSwiftPlugin")
-}.flatMap {
-    it.productMemberships
+}.map {
+    it.name
 }
 
 private fun Project.registerSwiftCheckValidTask(): TaskProvider<*> = tasks.register<Exec>("swift-check-valid") {
@@ -50,7 +49,7 @@ fun Project.registerJextractTask(
         inputs.file(File(projectDir, "Package.swift"))
 
         // monitor all targets/products which depend on the JExtract plugin
-        swiftProductsWithJExtractPlugin().forEach { targetName ->
+        swiftTargetsWithJExtractPlugin().forEach { targetName ->
             logger.info("[swift-java:jextract (Gradle)] Swift input target: ${targetName}")
             inputs.dir(File(layout.projectDirectory.asFile, "Sources/${targetName}"))
             outputs.dir(
--- a/BuildLogic/src/main/kotlin/utilities/SwiftPMTarget.kt
+++ b/BuildLogic/src/main/kotlin/utilities/SwiftPMTarget.kt
@@ -19,6 +19,7 @@ import kotlinx.serialization.Serializable
 
 @Serializable
 internal data class SwiftPMTarget(
+    val name: String,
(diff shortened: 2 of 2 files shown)


BuildLogic/src/main/kotlin/swiftPackageDescribe.kt
```diff
@@ -0,0 +1,33 @@
+import org.gradle.api.Project
```
ktoso, line 15: All source files need to have the copyright header (see the license validation test https://github.com/swiftlang/swift-java/actions/runs/21143842631/job/60835980516?pr=525 ) would you mind adding those?

timofeys1: Forgot about those. I pushed the license update

BuildLogic/src/main/kotlin/utilities/registerJextractTask.kt
```diff
@@ -50,21 +50,14 @@ fun Project.registerJextractTask(
...
-        val baseSwiftPluginOutputsDir = layout.buildDirectory.dir("../.build/plugins/outputs/").get().asFile
-        if (!baseSwiftPluginOutputsDir.exists()) {
-            baseSwiftPluginOutputsDir.mkdirs()
-        }
-        Files.walk(layout.buildDirectory.dir("../.build/plugins/outputs/").get().asFile.toPath()).forEach {
-            // Add any Java sources generated by the plugin to our sourceSet
-            if (it.endsWith("JExtractSwiftPlugin/src/generated/java")) {
-                outputs.dir(it)
-            }
+        swiftProductsWithJExtractPlugin().forEach { targetName ->
```
timofeys1: I had to change this logic a bit. CI builds failed because for the Kotlin build scripts, Gradle generates some `.java` sources, which were accidentally picked up by the `compileJava` task because of [a very general output path](https://github.com/swiftlang/swift-java/pull/525/commits/8ef437c76cea10c415e3d2f2236ea611235f8da0#diff-c7d443e975e400c2b03c18730e9a04a87accd5bc9560fca1db1c0a755a1eed57L57). The configuration-time `Files.walk`, which declared outputs, seems incorrect because on a clean run the outputs will not be there so I just declare the generated output path ahead of time.

timofeys1: I think this code:
```
swiftProductsWithJExtractPlugin().forEach { targetName -> 
```
inaccurately relied on product name matching the target name. This was used to [declare inputs](https://github.com/swiftlang/swift-java/blob/680a3fce55d23e7ceabfecfae6a206100adbc024/Samples/SwiftJavaExtractJNISampleApp/build.gradle#L88) and now also to declare outputs. Should I fix this?

ktoso: You're probably right, fix would be welcome!

BuildLogic/src/main/kotlin/swiftPackageDescribe.kt
timofeys1: I ran the license check scripts locally through `act`, and we will also need swiftlang/github-workflows#231

BuildLogic/src/main/kotlin/utilities/registerJextractTask.kt
timofeys1: I updated this code: if you change the target name now, the Java compilation will work, but there will be a runtime error because, iiuc, `jextract` assumes the target name where the plugin runs is going to match the `product` name, and so it emits the `LIB_NAME` with this assumption. I think that's what #521 is about

ktoso: Yeah that makes sense, we'll follow up on that one

BuildLogic/src/main/kotlin/swiftPackageDescribe.kt
ktoso: https://github.com/swiftlang/github-workflows/pull/231 merged, thank you!

PR #528, opened 2026-01-20
Fix crash when `Any` type parsing

sidepelican:
When reading the following code causes the jextract command to crash.

```swift
public func f() -> Any {
    return 0
}
```

> JExtractSwiftLib/SwiftType.swift:368: Fatal error: Unexpectedly found nil while unwrapping an Optional value

According to swift-syntax, `Any` is not an `Identifier` and thus cannot be handled by the `lookup`.

https://github.com/swiftlang/swift-syntax/blob/65b02a90ad2cc213e09309faeb7f6909e0a8577a/Sources/SwiftSyntax/Identifier.swift#L31

Since `Any` is existential, neither JNI mode nor FFM mode is currently supported. Therefore, I'll modify the implementation to simply ignore it.

### DIFF
--- a/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
+++ b/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
@@ -65,9 +65,9 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Tuples: `(Int, String)`, `(A, B, C)`                                                 | ❌        | ❌   |
 | Protocols: `protocol`                                                                | ❌        | ✅   |
 | Protocols: `protocol` with associated types                                          | ❌        | ❌   |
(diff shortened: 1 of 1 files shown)


sidepelican:
Ah, I've discovered that existentials in argument position are supported in JNI mode.
I need to reconsider how `Any` should be represented.

ktoso:
Edit with more ideas: I guess we'd have to defensively copy into a new buffer since we don't know if the value is a value or an instance to retain... that'd force us into arena use as well (tho in JNI i guess going global by defalt may be fine).

It's a tricky question, ok to leave as is for now

sidepelican:
Implementing full support for `Any` seems take a time.
My main focus is to avoid the fatal error, so I’d like to handle `Any` as unsupported.

ktoso:
Yeah that works for me, thanks

ktoso:
Whoops misclicked, tests passed

ktoso (approved):

PR #537, opened 2026-01-22
@JavaMethod which crashes now includes whole exception stacktrace

ktoso:
Previously we'd get a single line or exception's "message". This is not actionable at all usually as the `reason` may contain more information about the issue.

Making this change was difficult because we need `SwiftJava/Throwable` to gain `open func printStackTrace(_ writer: PrintWriter?)` and in order to do that we need to get `PrintWriter` into that module. So we take it from `JavaIO` which then causes a cascade of problems related to missing types, so we need to lower them all into `SwiftJava/`.

As such, this change is 4 lines in the generator, and then TONS OF moving around types and adding `StringWriter` and `PrintWriter` so we can use the new API. This is effectively just moving around where we generate the wrappers, and then making use of it.

Since the wrap-java command now is adding comments, some existing types got those as well.

We did not lose any methods in this process.

### DIFF
--- a/Sources/SwiftJavaMacros/JavaMethodMacro.swift
+++ b/Sources/SwiftJavaMacros/JavaMethodMacro.swift
@@ -126,23 +126,15 @@ extension JavaMethodMacro: BodyMacro {
       }
 
     let resultSyntax: CodeBlockItemSyntax =
-      if canRethrowError {
-        """
-        try {
+      """
+      \(raw: canRethrowError ? "try " : ""){
+        do {
           return try dynamicJava\(raw: isStatic ? "Static" : "")MethodCall(methodName: \(literal: funcName)\(raw: parametersAsArgs)\(raw: resultType))
-        }()
-        """
-      } else {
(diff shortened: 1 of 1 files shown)


Sources/SwiftJavaMacros/JavaMethodMacro.swift
```diff
@@ -109,18 +109,47 @@ extension JavaMethodMacro: BodyMacro {
...
+      } else {
+        """
+        if let throwable = error as? Throwable {
+          let sw = StringWriter()
+          let pw = PrintWriter(sw)
+          throwable.printStackTrace(pw)
+          fatalError("Java call threw unhandled exception: \\(error)\\n\\(sw.toString())")
+        }
+        fatalError("Java call threw unhandled exception: \\(error)")
+        """
```
ktoso, line 125: The actual™ change, this is what we want to make possible -- everything else in this patch is just moving around types so we're able to generate and use this code.

PR #544, opened 2026-01-29
Fix Swift overloads producing uncompilable Java wrappers

### LINKED ISSUE #526: Swift overloads produce uncompilable wrappers
# Reproduction

It seems that overloads currently produce incorrect Java code. This is reproducible by creating:
```
public class OverloadSample {
    public func takeValue(a: String) {}
    public func takeValue(b: String) {}
}
```
in `Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/OverloadSample.swift` and running `./gradlew :Samples:SwiftJavaExtractFFMSampleApp:compileJava`.

The compilation fails with:
```
OverloadSample.java:129: error: method takeValue(String) is already defined in class OverloadSample
    public void takeValue(java.lang.String b) {
```

Clemo97:
Fixes #526 

## Problem

Swift allows method overloading using parameter labels (argument labels), where methods can have the same base name but different parameter labels:

```swift
public func takeValue(a: String) {}
public func takeValue(b: String) {}
```

In Swift, these are distinct methods because parameter labels are part of the function signature. However, when translated to Java, both were being generated with the same signature, causing compilation failures:

```java
public void takeValue(String) //   Compilation error: duplicate method
```

**Error:**
```
error: method takeValue(String) is already defined in class OverloadSample
```

## Solution

Modified the Java method name generation in `FFMSwift2JavaGenerator+JavaTranslation.swift` to append parameter labels as suffixes, creating unique method names that preserve Swift semantics while ensuring valid Java code.

**Implementation:**
- Parameter labels are joined with underscores (e.g., `_a`, `_b`)
- Applied consistently to all methods (functions, initializers)
- Getters and setters are excluded to maintain property access patterns

**Example transformation:**

```swift
// Swift
public func takeValue(a: String) {}
public func takeValue(b: String) {}
```

```java
// Generated Java (Before)
public void takeValue(String a) //   Duplicate!
public void takeValue(String b) //   Compilation error

// Generated Java (After)
public void takeValue_a(String a) //   Unique
public void takeValue_b(String b) //   Compiles successfully
```

## Changes

- **`Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift`**: Added logic to append parameter labels to method names for disambiguation
- **`Samples/SwiftJavaExtractFFMSampleApp/src/main/java/com/example/swift/HelloJava2Swift.java`**: Updated sample code to use new method naming convention (e.g., `globalTakeInt_i()`, `init_len_cap()`)
- **`Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/OverloadSample.swift`**: Added test cases with overloaded methods.


ktoso:
Thanks for the request.what’s a reasonable direction however we should only be doing this if we detect a method has conflicts, otherwise it could be pretty annoying. Many members don’t have overloads and we can still call them with her normal names after all.

ktoso (changes requested):
marking as requesting changes

Clemo97:
> Thanks for the request.what’s a reasonable direction however we should only be doing this if we detect a method has conflicts, otherwise it could be pretty annoying. Many members don’t have overloads and we can still call them with her normal names after all.

Yeah, my thought is to add a conflict detection logic one of the FFM directories that will collect all methods that will be generated from a class, compute each methods java signtature without suffixes then use that to detect duplicates from overloading and add parameter label suffixes to conflicting ones. @copilot

ktoso:
I appreciate the will to help out but please don’t just at a coding assistant and expect it to figure out the implementation.

when you’re contributing you are ultimately responsible for the quality of the code. Feel free to use coding assistants but please verify and adjust their results such that they are passing our standards for quality and achieve the goals set out.

Back on the topic of implementation: yeah basically we’ll need such “conflicting names” mapping and then do another pass over them applying the renaming

Clemo97:
> I appreciate the will to help out but please don’t just at a coding assistant and expect it to figure out the implementation.
> 
> when you’re contributing you are ultimately responsible for the quality of the code. Feel free to use coding assistants but please verify and adjust their results such that they are passing our standards for quality and achieve the goals set out.
> 
> Back on the topic of implementation: yeah basically we’ll need such “conflicting names” mapping and then do another pass over them applying the renaming

Thanks for the feedback.

ktoso:
@Clemo97 are you working on this actively? I'm a bit confused with the periodic merges here but no action taken on the requested change?

Samples/SwiftJavaExtractFFMSampleApp/src/main/java/com/example/swift/HelloJava2Swift.java
```diff
@@ -68,36 +68,36 @@ static void examples() {
...
         }
 
         try (var arena = AllocatingSwiftArena.ofConfined()) {
             var bytes = arena.allocateFrom("hello");
-            var dat = Data.init_bytes_count(bytes, bytes.byteSize(), arena);
-            MySwiftLibrary.globalReceiveSomeDataProtocol_data(dat);
-            MySwiftLibrary.globalReceiveOptional_o1_o2(OptionalLong.of(12), Optional.of(dat));
+            var dat = Data.init(bytes, bytes.byteSize(), arena);
+            MySwiftLibrary.globalReceiveSomeDataProtocol(dat);
+            MySwiftLibrary.globalReceiveOptional(OptionalLong.of(12), Optional.of(dat));
```
ktoso, line 100: cool, now we need actual tests for the conflicting behavior please :)

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -13,9 +13,78 @@
...
 
 import JavaTypes
-import SwiftJavaConfigurationShared
+import SwiftJavaConfigurationShared 
 
 extension FFMSwift2JavaGenerator {
+  /// Detects method name conflicts within a nominal type and marks methods that need parameter label suffixes.
+  /// This implements a two-pass approach: first detect conflicts, then apply suffixes only where needed.
+  /// Only marks methods as conflicting if they would have identical Java signatures (same parameter types).
+  func detectMethodNameConflicts(for decl: ImportedNominalType) {
```
ktoso: Can this be "detectJavaMethodNameDupes"? And can it please return a value instead of mutating self directly?

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -13,9 +13,78 @@
...
+        
+        if !foundGroup {
+          signatureGroups.append([method])
+        }
+      }
+      
+      // Only mark methods as conflicting if they share the same Java signature
+      for group in signatureGroups where group.count > 1 {
+        for method in group {
+          methodsNeedingSuffixes.insert(ObjectIdentifier(method))
```
ktoso: I don't love the ObjectIdentifier because it will be hard to debug and print those; can you please use some readable representation for the storage? Can be string names I guess

ktoso: Maybe all this needs to become a class which is "DuplicateNames" and inside there do the storage, detection and everything -- that'll be much cleaner

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -145,10 +218,12 @@ extension FFMSwift2JavaGenerator {
   struct JavaTranslation {
     let config: Configuration
     var knownTypes: SwiftKnownTypes
+    let shouldAppendParameterLabels: Bool
```
ktoso: Do we need this explicitly or should we pass the `DuplicateNames` call then during translate(), that might be cleaner than adding bools

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -163,14 +238,21 @@ extension FFMSwift2JavaGenerator {
...
         suffix = ""
       default:
-        suffix = decl.functionSignature.parameters
-          .map { "_" + ($0.argumentLabel ?? "_") }
-          .joined()
+        if shouldAppendParameterLabels {
+          // Conflict detected: append parameter labels for disambiguation
+          suffix = decl.functionSignature.parameters
+            .map { "_" + ($0.argumentLabel ?? "_") }
+            .joined()
```
ktoso: Make this into a private method `makeMethodNameWithParamsSuffix()`

ktoso: The whole "javaName = makeJavaMethodName(decl)" and put the logic there

ktoso:
Still conflicts :)

Clemo97:
> Still conflicts :)

I have managed to resolve the conflicts.

ktoso:
I'm sorry but this still isn't passing basic `swift test` and other tests as well.

Please spend some effort on making sure tests pass or we'll have to close the pull request.

Clemo97:
> I'm sorry but this still isn't passing basic `swift test` and other tests as well.
> 
> Please spend some effort on making sure tests pass or we'll have to close the pull request.

Noted.
Let me get on that.

ktoso:
Should I try to take this over the finish line?

ktoso:
I'm going to close this since it doesn't seem like it's getting closer to completion.
I'll co-attribute when I do the solution, but please spend the effort to complete PRs next time.

Issue #517, opened 2026-01-08
jextract jni/ffm modes alignment: make FFM wrapper func throw if integer overflow would happen


ktoso:
Follow up to https://github.com/swiftlang/swift-java/pull/460 which did this in JNI mode, but now the FFM mode does not do this so we have inconcistency between the modes.

We should probably move the check into Java BEFORE doing the downcall, rather than throwing from the swift thunk -- then the code can be shared betweeh the two modes.

ktoso:
FYI @pelekon if you'd be up for this follow up

DPrakashhh:
Hi @ktoso! I'm really interested in this issue as a way to start contributing to the project. Full disclosure: I'm currently a Java beginner, but I'm eager to learn by doing.

I've looked at the description and I understand the goal is to move the overflow checks from the native side to the Java side for FFM consistency. If you're open to it, I'd love to try and tackle this—I'll likely need a little guidance on where the specific Java-side generation logic lives, but I'm happy to put in the work to figure it out.

Is it okay if I give this a shot?

pelekon:
I am fine with that, I am using all scraps of my free time for topics from last meeting, so I don't have this issue on my list rn.

PR #546, opened 2026-01-31
jextract: Implement Java-side integer overflow checks for FFM mode (#517)

DPrakashhh:
**Description**
**Summary:** This PR aligns FFM mode with JNI by moving integer validation to the Java side. This prevents native crashes by catching out-of-range values before they reach Swift.

**Key Changes**
1-**Java-Side Validation**: Moved range checks from native thunks to Java wrappers using a shared `ArithmeticException` helper.
2-**Type Widening**: Parameters (e.g., `Int8/16/32`) are now widened in Java to allow safe bounds checking against Swift's target limits.
3-**Unsigned Handling**: Added specific checks to bridge the gap between Java's signed types and Swift's `UInt` types.
4-**Updated Tests**: Adjusted "golden" expectations for `UnsignedNumberTests`,` MethodImportTests`, and others to reflect new safety logic.

**Verification**
1-**Full Suite:** All 201 tests passed across 41 suites.
2-**Toolchain**: Verified on JDK 25 and Swift 6.2.1.

### DIFF
--- a/Sources/JExtractSwiftLib/Swift2JavaTranslator.swift
+++ b/Sources/JExtractSwiftLib/Swift2JavaTranslator.swift
@@ -110,6 +110,7 @@ extension Swift2JavaTranslator {
       let dataProtocolDecl = (self.symbolTable[.foundationDataProtocol] ?? self.symbolTable[.essentialsDataProtocol])!
       if self.isUsing(where: { $0 == dataDecl || $0 == dataProtocolDecl }) {
         visitor.visit(nominalDecl: dataDecl.syntax!.asNominal!, in: nil, sourceFilePath: "Foundation/FAKE_FOUNDATION_DATA.swift")
+        visitor.visit(nominalDecl: dataProtocolDecl.syntax!.asNominal!, in: nil, sourceFilePath: "Foundation/FAKE_FOUNDATION_DATAPROTOCOL.swift")
       }
     }
 


Tests/JExtractSwiftTests/VariableImportTests.swift
```diff
@@ -103,6 +103,9 @@ final class VariableImportTests {
          */
         public void setCounterInt(long newValue) {
           $ensureAlive();
+          if (SwiftValueLayout.SWIFT_INT == ValueLayout.JAVA_INT) {
```
ktoso: This feels like we should introduce some "`SwiftValueLayout.has32bitSwiftInt`" or something similar?

DPrakashhh:
I’ve updated the logic to use the new `SwiftValueLayout.has32bitSwiftInt` helper as suggested. I also included some minor formatting/indentation fixes to keep the generated code consistent.

Please let me know if you’d prefer the formatting changes removed; otherwise, the logic is ready for another look! 

Sir, its ready for another look :)

ktoso:
That's a lot of unrelated changes, may I ask if you're using some agent to write these changes? It makes it difficult to review, please try to keep changes committed to a minimum so that that they are easier to review.

I ended up sick at a conference just now... I'll review as soon as I'm able to in a few days

DPrakashhh:
I am closing this PR - ig starting a new one will be better after all this mess :/

Closing this in favor of a fresh, clean PR with the `has32bitSwiftInt` helper and refined signed/unsigned logic. New PR is here: https://github.com/swiftlang/swift-java/pull/549

PR #548, opened 2026-02-02
Import DataProtocol's decl along with Data

sidepelican:
Currently, using `Foundation.Data` in JNI mode prevents Java code from compiling.

```
/..../intermediates/swift/generatedJava/debug/Data.java:12: error: cannot find symbol
public final class Data implements JNISwiftInstance, DataProtocol {
                                                     ^
  symbol: class DataProtocol
```

This occurs because in
https://github.com/swiftlang/swift-java/blob/291fb8a9c6a3218bc30f1d3504b2609be0cf6744/Sources/JExtractSwiftLib/Swift2JavaTranslator.swift#L109-L114

`dataDecl` uses `DataProtocol` but does not visit `dataProtocolDecl`.
In FFM mode, protocols were not yet supported, `DataProtocol` would automatically be eliminated without causing issues.


sidepelican:
I understand that this fix isn't a complete support for `Data`.
All it does is resolve the broken compilation issue.

sidepelican:
I found a test failure in mylocal, but realized it might not be an issue with my environment, so I marked it as a draft.

sidepelican:
Seeing CI results, everything seems fine.
It seems my local environment is unstable.

ktoso:
Cool, happy to accept this as is

PR #553, opened 2026-02-06
Skip static requirements in protocols

sidepelican:
Currently, when generating Java code from a Swift protocol (JNI only), `static` requirements are mapped directly to `static` methods in a Java `interface`.
However, unlike Swift protocols, Java interfaces do not support 'static requirements'. This leads to the following compilation errors:

```
/Users/iceman/github/swift-java/Samples/SwiftJavaExtractJNISampleApp/.build/plugins/outputs/swiftjavaextractjnisampleapp/MySwiftLibrary/destination/JExtractSwiftPlugin/src/generated/java/com/example/swift/ProtocolWithStatic.java:23: error: missing method body, or declare abstract
  public static long myFunc();
                     ^
/Users/iceman/github/swift-java/Samples/SwiftJavaExtractJNISampleApp/.build/plugins/outputs/swiftjavaextractjnisampleapp/MySwiftLibrary/destination/JExtractSwiftPlugin/src/generated/java/com/example/swift/ProtocolWithStatic.java:35: error: missing method body, or declare abstract
  public static long getValue();
                     ^
```

While there might be a design pattern to introduce 'static requirements' in Java, this feels like a complex change with many points of contention that would require significant time.

To ensure the generated code is compilable, this PR updates the generator to ignore them.

### DIFF
--- a/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
+++ b/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
@@ -65,6 +65,7 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Tuples: `(Int, String)`, `(A, B, C)`                                                 | ❌        | ❌   |
 | Protocols: `protocol`                                                                | ❌        | ✅   |
 | Protocols: `protocol` with associated types                                          | ❌        | ❌   |
+| Protocols static requirements: `static func`, `init(rawValue:)`                      | ❌        | ❌   |
 | Existential parameters `f(x: any SomeProtocol)` (excepts `Any`)                      | ❌        | ✅   |
 | Existential parameters `f(x: any (A & B)) `                                          | ❌        | ✅   |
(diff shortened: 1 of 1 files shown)


ktoso (approved):
Lgtm but please add a note to the features list that this isn't supported

sidepelican:
Actually, I realized there is a slight gap in this implementation.
I missed the case where JavaInterfaceProtocolWrapper is generated.

JavaInterfaceProtocolWrapper cannot be generated when there are static requirements, we will need to skip it in those cases as well.

https://github.com/swiftlang/swift-java/blob/1fbab460ed062dec93133ee7cdfe55c87157d90e/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator%2BInterfaceWrapperGeneration.swift#L80

That said, this PR still provides an improvement as-is. 
We could merge this first, and I can address the remaining JavaInterfaceProtocolWrapper issue in a separate PR.

ktoso:
That's fine, let's merge this and follow up then

PR #555, opened 2026-02-09
Skip InterfaceWrapper generation for protocols with static requirements

sidepelican:
- Fix https://github.com/swiftlang/swift-java/pull/553#issuecomment-3858931173

InterfaceWrappers are generated to allow Java classes to conform to Swift protocols. However, if a protocol includes static requirements, the generated code results in a compilation error.

Currently, this hasn't been a major issue because InterfaceWrappers are only generated for types where properties are not defined.
Since it is practically difficult to implement Swift's metatype from Java, we should skip generating InterfaceWrappers for such protocols.

I've updated the generator to ignore protocols with static requirements during the InterfaceWrapper generation process.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+InterfaceWrapperGeneration.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+InterfaceWrapperGeneration.swift
@@ -23,7 +23,7 @@ extension JNISwift2JavaGenerator {
   ) -> [ImportedNominalType: JavaInterfaceSwiftWrapper] {
     var wrappers = [ImportedNominalType: JavaInterfaceSwiftWrapper]()
 
-    for type in types {
+    for type in types where type.swiftNominal.kind == .protocol {
       do {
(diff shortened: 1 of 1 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+InterfaceWrapperGeneration.swift
```diff
@@ -23,7 +23,7 @@ extension JNISwift2JavaGenerator {
   ) -> [ImportedNominalType: JavaInterfaceSwiftWrapper] {
     var wrappers = [ImportedNominalType: JavaInterfaceSwiftWrapper]()
 
-    for type in types {
+    for type in types where type.swiftNominal.kind == .protocol {
```
sidepelican, line 26: I noticed that JavaInterfaceProtocolWrapperGenerator was being used for all types erroneously.

ktoso: I see, nice catch

PR #557, opened 2026-02-09
jextract/jni: Add initial Foundation.Data support

ktoso:
This also improves the Array.getJNIValue(in:) when the array is of bytes, addressing a long standing TODO.

I also worked on optimizing the `toByteArray` some more and now we're able to get such performance when doing the copying (by avoiding iterating and mapping data into jbytes):

```
Benchmark                                      (dataSize)  (taskCount)  Mode  Cnt      Score      Error  Units
// how you'd do it naively:
JNIDataBenchmark.jni_data_toByteArrayIndirectCopy       4          N/A  avgt   30    616.328 ±   13.157  ns/op
JNIDataBenchmark.jni_data_toByteArrayIndirectCopy     100          N/A  avgt   30   4030.721 ±   48.059  ns/op
JNIDataBenchmark.jni_data_toByteArrayIndirectCopy    1000          N/A  avgt   30  35415.920 ±  253.776  ns/op <<< "what you'd do naively"

// how we do it now:
JNIDataBenchmark.jni_data_toByteArray                   4          N/A  avgt   30    264.089 ±   10.726  ns/op
JNIDataBenchmark.jni_data_toByteArray                 100          N/A  avgt   30    256.012 ±    1.972  ns/op
JNIDataBenchmark.jni_data_toByteArray                1000          N/A  avgt   30    282.767 ±    4.469  ns/op << _nice_
```

These work because uint8 and jbyte should be really have the same layout AFAIK so we don't need to iterate, but just rebind the memory and use the JNI funcs to initialize the java array. We could do even better by initializing into an existing array, but I didn't do that yet.

I did use the help of Claude (with lots of hand holding) to get some of the boilerplate done here, and especially make sure the FFM and JNI modes are roughly equivalent. I've read and vetted all the code and tests, and most documentation was written by myself.

### DIFF
--- a/Sources/SwiftJava/BridgedValues/JavaValue+Array.swift
+++ b/Sources/SwiftJava/BridgedValues/JavaValue+Array.swift
@@ -23,29 +23,56 @@ extension Array: JavaValue where Element: JavaValue {
 
   public init(fromJNI value: JNIType, in environment: JNIEnvironment) {
     let jniCount = environment.interface.GetArrayLength(environment, value)
-    let jniArray: [Element.JNIType] =
-      if let value {
-        .init(
-          unsafeUninitializedCapacity: Int(jniCount)
-        ) { buffer, initializedCount in
-          Element.jniGetArrayRegion(in: environment)(
+    let count = Int(jniCount)
+
+    guard let value else {
+      self = []
+      return
+    }
+
+    // Fast path for byte types: Since the memory layout of `jbyte` (Int8) and UInt8/Int8 is identical,
+    // we can rebind the memory and fill it directly without creating an intermediate array.
+    // This mirrors the optimization in `getJNIValue` in the reverse direction.
+    if Element.self == UInt8.self {
+      let result = [UInt8](unsafeUninitializedCapacity: count) { buffer, initializedCount in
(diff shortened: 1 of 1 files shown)


madsodgaard:
Very nice! 

Could we add the same mechanism for `init(fromJNI:)`? Such that we also only incur a single copy for `Data.fromByteArray`?

ktoso:
Yep we can do that, following up

Sources/SwiftJava/BridgedValues/JavaValue+Array.swift
```diff
@@ -23,29 +23,56 @@ extension Array: JavaValue where Element: JavaValue {
...
+        initializedCount = count
       }
-
-    // FIXME: If we have a 1:1 match between the Java layout and the
-    // Swift layout, as we do for integer/float types, we can do some
-    // awful alias tricks above to have JNI fill in the contents of the
-    // array directly without this extra copy. For now, just map.
-    self = jniArray.map { Element(fromJNI: $0, in: environment) }
+      self = result as! Self
+    } else if Element.self == Int8.self {
```
ktoso, line 50: Could not nicely pull this off in a single branch... doing some small local func was even more annoying tbh -- since we need the concrete Int8.jniGetArrayRegion

PR #558, opened 2026-02-09
try to enable format check again

ktoso:
This used to just hang, maybe we're more lucky with newer swift now


.github/workflows/pull_request.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       # Not API stable package (yet)
       api_breakage_check_enabled: false
       # FIXME: Something is off with the format task and it gets "stuck", need to investigate
```
ktoso, line 16: remove the fixme if it works now

PR #559, opened 2026-02-10
Suppress warning for "inferred to have type '()',"

sidepelican:
This is a minor improvement to silence a compiler warning in the generated code like below.

```
/Users/iceman/github/swift-java/Samples/SwiftJavaExtractJNISampleApp/.build/plugins/outputs/swiftjavaextractjnisampleapp/MySwiftLibrary/destination/JExtractSwiftPlugin/Sources/MySwiftLibraryModule+SwiftJava.swift:864:13: warning: constant 'swiftResult$' inferred to have type '()', which may be unexpected
 862 |       } // printTaskBody(printer:) @ JExtractSwiftLib/JNISwift2JavaGenerator+NativeTranslation.swift:1151
 863 |       do {
 864 |         let swiftResult$ = try await MySwiftLibrary.asyncThrows()
     |             |- warning: constant 'swiftResult$' inferred to have type '()', which may be unexpected
     |             `- note: add an explicit type annotation to silence this warning
 865 |         environment = try! JavaVirtualMachine.shared().environment()
 866 |         _ = environment.interface.CallBooleanMethodA(environment, globalFuture, _JNIMethodIDCache.CompletableFuture.complete, [jvalue(l: nil)])
```

Since the generator already checks for `.isVoid` at this location, I have slightly adjusted the logic to avoid assigning the result.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1227,12 +1227,14 @@ extension JNISwift2JavaGenerator {
             } else {
               "await"
             }
-          printer.print("let swiftResult$ = \(tryAwaitString) \(placeholderWithoutTry)")
-          printer.print("environment = try! JavaVirtualMachine.shared().environment()")
-          let inner = nativeFunctionSignature.result.conversion.render(&printer, "swiftResult$")
           if swiftFunctionResultType.isVoid {
+            printer.print("\(tryAwaitString) \(placeholderWithoutTry)")
+            printer.print("environment = try! JavaVirtualMachine.shared().environment()")
             printer.print("_ = environment.interface.CallBooleanMethodA(environment, globalFuture, \(completeMethodID), [jvalue(l: nil)])")
           } else {
+            printer.print("let swiftResult$ = \(tryAwaitString) \(placeholderWithoutTry)")
+            printer.print("environment = try! JavaVirtualMachine.shared().environment()")
+            let inner = nativeFunctionSignature.result.conversion.render(&printer, "swiftResult$")
             let result: String
             if nativeFunctionSignature.result.javaType.requiresBoxing {
               printer.print("let boxedResult$ = SwiftJavaRuntimeSupport._JNIBoxedConversions.box(\(inner), in: environment)")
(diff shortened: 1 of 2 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1227,12 +1227,14 @@ extension JNISwift2JavaGenerator {
...
-          printer.print("let swiftResult$ = \(tryAwaitString) \(placeholderWithoutTry)")
-          printer.print("environment = try! JavaVirtualMachine.shared().environment()")
-          let inner = nativeFunctionSignature.result.conversion.render(&printer, "swiftResult$")
           if swiftFunctionResultType.isVoid {
+            printer.print("\(tryAwaitString) \(placeholderWithoutTry)")
+            printer.print("environment = try! JavaVirtualMachine.shared().environment()")
             printer.print("_ = environment.interface.CallBooleanMethodA(environment, globalFuture, \(completeMethodID), [jvalue(l: nil)])")
           } else {
+            printer.print("let swiftResult$ = \(tryAwaitString) \(placeholderWithoutTry)")
+            printer.print("environment = try! JavaVirtualMachine.shared().environment()")
```
ktoso, line 1236: no need to move the environment initialization?

sidepelican: The only difference is whether we write the `if` statement twice or not.
I was okay with a single line of duplication, but we could also write the `if` statement twice if preferred.

ktoso: I mean, you can keep the `printer.print("environment = try! JavaVirtualMachine.shared().environment()")` before the if entirely so we don't duplicate the string in both branches

sidepelican: I thought it would be better not to change the timing of fetching `environment` as it might span across a suspention point (`await`).

ktoso: Hmm... let's go with this, though arguably before might be safer tbh...

PR #560, opened 2026-02-10
fix comments style in generated swift files

ktoso:
We wrongly generated comments in Swift files to be /** style, but it should be /// style.

I used Claude to fix all the assertions and existing files since it was a lot to do manually


ktoso:
Forgot to update tests ofc duh

PR #561, opened 2026-02-11
Reenable formatting checker

ktoso:
Just a baseline, we'll adjust the rules more


ktoso:
This project used to hang the formatter, seems it has improved and can handle it now :-)

PR #563, opened 2026-02-16
Don't generate int returning functions for now

madsodgaard:
Returning `Int` is not supported in JNI yet, and will fail to compile because of the support we added in parameter position. 

I might have time to look at it some point, but for now, let's just make the generated code compile.


ktoso:
Please make sure to file an issue about this if we don’t have it yet

madsodgaard:
Closing, as this will break for people relying on it working on 64-bit... I'll do a proper fix instead

PR #567, opened 2026-02-17
jextract/jni: Refactor `toString` generation to use the standard method generation path

sidepelican:
## Motivation

Currently, the implementations for `toString` and `toDebugString` methods are handled by dedicated, hardcoded logic.
However, these methods are fundamentally identical to standard function calls in terms of their underlying mechanism.

This inconsistency makes it difficult to maintain parity with other parts of the generator. Specifically, as I work on implementing support for function calls in generic type, I find myself having to write redundant specialized logic just for `toString`, which adds unnecessary complexity.

## Changes

Treat `toString` and `toDebugString` as `ImportedFunc`.
This allows them to be processed through the general method generation path.

## Note 

While this change also affects FFM mode, it will currently be ignored because `String` return values are not yet supported in that mode.

### DIFF
--- a/Tests/JExtractSwiftTests/JNI/JNIToStringTests.swift
+++ b/Tests/JExtractSwiftTests/JNI/JNIToStringTests.swift
@@ -35,7 +35,7 @@ struct JNIToStringTests {
         public java.lang.String toString() {
           return MyType.$toString(this.$memoryAddress());
         }
-        private static native java.lang.String $toString(long self);
+        private static native java.lang.String $toString(long selfPointer);
         """
       ]
     )
@@ -51,9 +51,9 @@ struct JNIToStringTests {
       expectedChunks: [
         """
         @_cdecl("Java_com_example_swift_MyType__00024toString__J")
-        public func Java_com_example_swift_MyType__00024toString__J(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, self: jlong) -> jstring? {
+        public func Java_com_example_swift_MyType__00024toString__J(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, selfPointer: jlong) -> jstring? {
           ...
-          return String(describing: self$.pointee).getJNIValue(in: environment)
+          return String(describing: selfPointer$.pointee).getJNIValue(in: environment)
         }
         """
       ]
@@ -72,7 +72,7 @@ struct JNIToStringTests {
         public java.lang.String toDebugString() {
           return MyType.$toDebugString(this.$memoryAddress());
         }
-        private static native java.lang.String $toDebugString(long self);
+        private static native java.lang.String $toDebugString(long selfPointer);
         """
       ]
     )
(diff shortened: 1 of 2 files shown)


Sources/JExtractSwiftLib/ImportedDecls.swift
```diff
@@ -25,6 +25,8 @@ package enum SwiftAPIKind {
   case enumCase
   case subscriptGetter
   case subscriptSetter
+  case toString
+  case toDebugString
```
madsodgaard: Instead of adding these to `SwiftAPIKind`, perhaps we should introduce a new type `SynethesizedAPI` (or something) else to describe these sort of functions like `toString` etc?

sidepelican: Thank you for the feedback. Do you have a specific vision or any examples of how you'd like to implement `SynthesizedAPI`?

madsodgaard: Hm, I was just thinking of adding it to the `ImportedFunc` as an optional and then we would do
```
enum SynthesizedAPI {
  case toString
  case toDebugString
```

Just an idea to not pollute the `SwiftAPIKind`, as it seems like it should only concern itself with Swift language APIs

sidepelican: I believe it's important to ensure we can handle the branching logic correctly in the following locations:

https://github.com/swiftlang/swift-java/blob/292af7604d6637e09b9bd2cab54afae9bdcb0f41/Sources/JExtractSwiftLib/FFM/CDeclLowering/FFMSwift2JavaGenerator%2BFunctionLowering.swift#L986

https://github.com/swiftlang/swift-java/blob/292af7604d6637e09b9bd2cab54afae9bdcb0f41/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator%2BSwiftThunkPrinting.swift#L614

Considering this, I think `SynthesizedAPI` should be integrated into `SwiftAPIKind` as an associated value, like this:

```swift
enum SynthesizedAPI { 
  case toString, toDebugString
}

enum SwiftAPIKind {
  case function(SynthesizedAPI?)
  case initializer
  case getter
  case setter
  case enumCase
  case subscriptGetter
  case subscriptSetter
}
```

Does this approach look good to you?

(In this case, `SynthesizedFunction` might be a better than `SynthesizedAPI`.)

ktoso: maybe it's `function` and `synthesizedFunction(SynthesizedAPI)` in that case?

sidepelican: Agreed, that looks better. I've updated it in bef0d3b9d416d6e88490c9bcbe7f35fae68638e2

sidepelican:
I have addressed the feedback. Are there any remaining concerns ?

Tests/JExtractSwiftTests/JNI/JNIToStringTests.swift
```diff
@@ -89,7 +88,7 @@ struct JNIToStringTests {
       expectedChunks: [
         """
         @_cdecl("Java_com_example_swift_MyType__00024toDebugString__J")
-        public func Java_com_example_swift_MyType__00024toDebugString__J(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, selfPointer: jlong) -> jstring? {
+        public func Java_com_example_swift_MyType__00024toDebugString__J(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, self: jlong) -> jstring? {
```
ktoso: we want to keep those as "selfPointer"

ktoso (approved):
Looks fine now, thank you

Tests/JExtractSwiftTests/JNI/JNIToStringTests.swift
ktoso: i'll apply that

ktoso:
Thank you!

sidepelican:
Thank you too! Happy to see this merged.

PR #569, opened 2026-02-23
jextract/jni: fix support for optional strings in protocols

madsodgaard:
Fixes support for protocols that accept or return optional strings:

```swift
protocol MyProtocol {
  func takeAndReturnString(_ string: String?) -> String?
}
```


madsodgaard:
@ktoso maybe the Android CI needs a retry

Issue #533, opened 2026-01-21
Classes with a type parameter fail to compile


abdulowork:
I searched through the issues, but couldn't find an explicit issue about the missing support for generics. In particular if you have a public class such as:
```
public class Foo<T> {
    public init() {}
}
```
in `Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/Foo.swift`, then the Swift wrappers compilation currently fails with:
```
error: generic parameter 'T' could not be inferred
 26 | @_cdecl("swiftjava_getType_MySwiftLibrary_Foo")
 27 | public func swiftjava_getType_MySwiftLibrary_Foo() -> UnsafeMutableRawPointer /* Any.Type */ {
 28 |   return unsafeBitCast(Foo.self, to: UnsafeMutableRawPointer.self)
    |                        |- error: generic parameter 'T' could not be inferred
    |                        `- note: explicitly specify the generic arguments to fix this issue
```

sidepelican:
@ktoso  I’ve outlined a potential approach to resolve this issue. I’d appreciate your feedback on the following plan.

## Investigation of Existing Interoperability

I first looked into how generics are handled in existing Swift interoperability layers:

* Objective-C: Generic types are not exported at all.
* C++: Only `struct` types support generics; `class` types do not. The implementation directly calls functions using mangled Swift symbols according to the calling convention. It leverages advanced C++ template features to implement copy semantics that match Swift’s generic logic. It allows Swift-derived types and some primitives as generic parameters, with static enforcement via templates (invalid types cause compilation errors).

## Proposed Strategy for swift-java

While I would like to follow the C++ implementation's lead, it is difficult for two reasons:

1. `jextract` cannot retrieve mangled names.
2. Java generics are more limited compared to C++ templates.

Therefore, I believe we need a different approach. I propose the following phased improvements:

---

## 1. Retaining Metatypes for Type Parameters

This phase focuses on fixing the current error:

```swift
error: generic parameter 'T' could not be inferred
 26 | @_cdecl("swiftjava_getType_MySwiftLibrary_Foo")
 27 | public func swiftjava_getType_MySwiftLibrary_Foo() -> UnsafeMutableRawPointer /* Any.Type */ {
 28 |   return unsafeBitCast(Foo.self, to: UnsafeMutableRawPointer.self)
    |                        |- error: generic parameter 'T' could not be inferred
    |                        `- note: explicitly specify the generic arguments to fix this issue
```

[...]

PR #572, opened 2026-02-25
jextract/jni: Add basic support for generic types

sidepelican:
## Purpose

This PR introduces basic support for accessing generic Swift types from Java.
As this is the initial step, the functionality is currently limited.

Specifically, the following are not supported in this PR:

* The Java-side wrappers do not yet have generic signatures.
* No access to static members. (Since there is no mechanism to specify type parameters from Java yet.)
* No access to members using type parameters

## Implementation

The general strategy follows the direction discussed in [this issue comment](https://github.com/swiftlang/swift-java/issues/533#issuecomment-3894556886).

In this implementation, the Swift metatype is stored as a field within the Java wrapper class.
When a function is called, this metatype is passed as an argument to the native function.

Regarding the usage of the metatype, I have adopted a different approach than the one originally mentioned in the issue.
When a type is generic, the generator now produces an "opener protocol".
This is a technique to handle the `Self` type correctly from a metatype. By conforming the original type to this opener protocol, we can indirectly extract and utilize the type parameters within Swift.

```swift
public struct MyID<T> {
  public var rawValue: T
  public var description: String {
    "\(rawValue)"
  }
}
```

```swift
protocol _MySwiftLibrary_MyID_opener {
  static func _get_description(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, self: jlong) -> jstring?
  ...
}

extension MyID: _MySwiftLibrary_MyID_opener {
  static func _get_description(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, self: jlong) -> jstring? {
    assert(self != 0, "self memory address was null")
    let selfBits$ = Int(Int64(fromJNI: self, in: environment))
    let self$ = UnsafeMutablePointer<MyID>(bitPattern: selfBits$)
    guard let self$ else {
      fatalError("self memory address was null in call to \(#function)!")
    }
    return self$.pointee.description.getJNIValue(in: environment)
  }
  ...
}

@_cdecl("Java_com_example_swift_MyID__00024getDescription__JJ")
public func Java_com_example_swift_MyID__00024getDescription__JJ(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, self: jlong, selfType: jlong) -> jstring? {
  let selfTypeBits$ = Int(Int64(fromJNI: selfType, in: environment))
  guard let selfType$ = UnsafeRawPointer(bitPattern: selfTypeBits$) else {
    fatalError("selfType metadata address was null")
  }
  let openerType = unsafeBitCast(selfType$, to: Any.Type.self) as! (any _MySwiftLibrary_MyID_opener.Type)
  return openerType._get_description(environment: environment, thisClass: thisClass, self: self)
}
```

<details>
<summary>Why not use the original pattern?</summary>
This is because the original pattern cannot handle certain cases successfully.
It cannot expand when there are multiple type parameters and constraints.

</details>

## Others

The following functions cause branching in code generators based on whether they are generic or not:

* `long $typeMetadataAddress()`
* `Runnable $createDestroyFunction()`
* `Discriminator getDiscriminator()`

Similar to the refactoring in #567 for `toString`, I plan to register these as `SynthesizedAPI` to remove this redundancy. However, to avoid over-complicating this PR, I have deferred that refactoring to a future update.

### DIFF
--- a/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
+++ b/Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
@@ -362,7 +362,7 @@ or set the `asyncFuncMode` configuration value in `swift-java.config`
 
 > Note: Generic types are currently only supported in JNI mode. 
 
-Support for generic types is limited.
+Support for generic types is still work-in-progress and limited.
 The generated Java classes do not have generic signatures.
 Any members containing type parameters (such as T) are not exported.
 


Tests/JExtractSwiftTests/JNI/JNIGenericTypeTests.swift
```diff
@@ -0,0 +1,190 @@
...
+              "this", this,
+              "self", self$);
+          }
+          return new Runnable() {
+            @Override
+            public void run() {
+              if (CallTraces.TRACE_DOWNCALLS) {
+                CallTraces.traceDowncall("MyID.$destroy", "self", self$);
+              }
+              MyID.$destroy(self$, selfType$);
```
ktoso, line 93: 👍

Tests/JExtractSwiftTests/JNI/JNIGenericTypeTests.swift
```diff
@@ -0,0 +1,190 @@
...
+        """
+        private static native void $destroy(long selfPointer, long selfType);
+        @Override
+        public Runnable $createDestroyFunction() {
+          long selfType$ = this.$typeMetadataAddress();
+          long self$ = this.$memoryAddress();
+          if (CallTraces.TRACE_DOWNCALLS) {
+            CallTraces.traceDowncall("MyID.$createDestroyFunction",
+              "this", this,
+              "self", self$);
```
ktoso: Let's include the selfType$ now as well

Tests/JExtractSwiftTests/JNI/JNIGenericTypeTests.swift
```diff
@@ -0,0 +1,190 @@
...
+  func generateSwiftThunk() throws {
+    try assertOutput(
+      input: genericFile,
+      .jni,
+      .swift,
+      detectChunkByInitialLines: 2,
+      expectedChunks: [
+        """
+        protocol _SwiftModule_MyID_opener {
+          static func _get_description(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, self: jlong) -> jstring?
```
ktoso, line 112:
```suggestion
          static func _get_description(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, selfPointer: jlong) -> jstring?
```

sorry to be annoying about the `selfPointer` every time 😉 

I don't want it to be confusing when we see a `self` it should be the object and the `selfPointer` should be the `jlong` or other pointer repr

SwiftKitCore/src/main/java/org/swift/swiftkit/core/OutSwiftGenericInstance.java
```diff
@@ -0,0 +1,20 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package org.swift.swiftkit.core;
+
+public final class OutSwiftGenericInstance {
```
ktoso: Please document what this is for

SwiftKitCore/src/main/java/org/swift/swiftkit/core/_OutSwiftGenericInstance.java
```diff
@@ -0,0 +1,20 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package org.swift.swiftkit.core;
+
+public final class OutSwiftGenericInstance {
+    public long selfPointer;
+    public long selfTypePointer;
```
ktoso, line 27:
```suggestion
    public final long selfTypePointer;
```

If making them public please make them final

SwiftKitCore/src/main/java/org/swift/swiftkit/core/_OutSwiftGenericInstance.java
```diff
@@ -0,0 +1,20 @@
...
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package org.swift.swiftkit.core;
+
+public final class OutSwiftGenericInstance {
+    public long selfPointer;
```
ktoso, line 26:
```suggestion
    public final long selfPointer;
```

If making them public please make them final

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -60,6 +60,9 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Async functions `func async` and properties: `var { get async {} }`                  | ❌        | ✅   |
 | Arrays: `[UInt8]`, `[MyType]`, `Array<Int64>` etc                                    | ❌        | ✅   |
 | Dictionaries: `[String: Int]`, `[K:V]`                                               | ❌        | ❌   |
+| Generic type: `struct S<T>`                                                          | ❌        | ✅   |
+| Functions or properties using generic type param: `struct S<T> { func f(_: T) {} }`  | ❌        | ❌   |
+| Static functions or properties in generic type                                       | ❌        | ❌   |
```
ktoso, line 65: This deserves some more docs about how we import it, since we drop the generic <T> from the signature we should explain this -- but for simple things I thikn we'll be able to pull off doing a generic on the Java side tbh right?

Tests/JExtractSwiftTests/JNI/JNIGenericTypeTests.swift
sidepelican: I see. I don't have a strong preference, so I think selfPointer is fine as well.

The variable name `self` originates from this location:
https://github.com/swiftlang/swift-java/blob/da7d0a5982909cb40db5e19cdcb132aab7886d15/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator%2BNativeTranslation.swift#L55

Changing this to `selfPointer` would require updating a large number of existing tests.
Would you still like me to proceed with the change?

ktoso: Yeah let's bite the bullet, we can do it in a separate PR.

This is looking good, happy to land it after some nitpicks

ktoso (approved):
I'm happy to merge this as a first step towards generics support :) Just the few nitpicks please

ktoso:
I just fixed the android CI, a new run should pass

SwiftKitCore/src/main/java/org/swift/swiftkit/core/_OutSwiftGenericInstance.java
sidepelican: `OutSwiftGenericInstance` is a temporary container used for receiving multiple return values.

Both `selfPointer` and `selfTypePointer` are updated from the Swift side using JNI's `SetLongField`.
While JNI can technically bypass the `final` modifier to modify fields, I'm concerned that JIT optimizations might lead to unexpected behavior if the compiler makes incorrect assumptions about these values being constant.

Should I add the `final` modifier ?

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
sidepelican: I wrote documentation to make it clearer what gets exported and what doesn't.

https://github.com/swiftlang/swift-java/pull/572/changes/016766b1e74b4dc33c2f4c4918e78aebce06099f

SwiftKitCore/src/main/java/org/swift/swiftkit/core/_OutSwiftGenericInstance.java
ktoso: Hmm, I see. Maybe let’s make this some _ type or otherwise mark it somehow that’s clear that users should not be touching it?

im ok with the non final fields to make synthesis and JNI setting easier

sidepelican: I agree. I'll go with adding the `_` prefix.

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -354,3 +357,47 @@ This is a mode for legacy platforms, where `CompletableFuture` is not available,
 In this mode `async` functions in Swift are extracted as Java methods returning a `java.util.concurrent.Future`.
 To enable this mode pass the `--async-func-mode future` command line option, 
 or set the `asyncFuncMode` configuration value in `swift-java.config`
+
+### Generic types
+
+> Note: Generic types are currently only supported in JNI mode. 
+
+Support for generic types is limited.
```
ktoso:
```suggestion
Support for generic types is still work-in-progress and limited.
```

ktoso:
Let's get this first step in, thank you

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
sidepelican: Apologies, I missed the latter part of your comment. I also believe that simple generics are feasible, though I haven't done a detailed investigation yet.

PR #573, opened 2026-02-25
jextract/jni: Skip discriminator generation if enum has no cases

sidepelican:
Currently, when the Swift enum has no cases, the generated Java code results in a compilation error.

```
error: sealed class must have subclasses
  public sealed interface Case {}
                ^
```

In Swift, it is a common idiom to use enums with no cases as namespaces.

I have updated the generator to skip the generation of any case-related code when an enum contains no cases.


sidepelican:
Looks good, having fewer differences might be preferable.
I'll reconsider the implementation in that direction.

ktoso:
Thank you!

sidepelican:
New PR: https://github.com/swiftlang/swift-java/pull/577

ktoso:
https://github.com/swiftlang/swift-java/pull/577 faced some complications, so let's accept this one actually, thank you!

PR #574, opened 2026-02-25
ci: stop building for Android in a container

finagolfin:
This will fix the recent break because I updated the official SDK workflow script, as we only run that one on the GitHub runner directly, and stop downloading the NDK on every run.

I've also greatly broadened out the testing matrix, as all these other options should work too. We can trim them back down, if wanted, after trying a run.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -203,7 +203,7 @@ jobs:
           apt-get -q update && apt-get -yq install curl
           cd Samples/SwiftJavaExtractJNISampleApp
           curl -L -O --retry 3 https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/install-and-build-with-sdk.sh
-          perl -pi -e "s#(download_and_extract_toolchain \".ANDROID_SDK_TAG\"\))#\1\n        PATH=\\\$(\\\$SWIFT_EXECUTABLE_FOR_ANDROID_SDK):\\\$PATH\n        echo \"path is now \\\$PATH\"#" install-and-build-with-sdk.sh
+          perl -pi -e "s#(download_and_extract_toolchain \".ANDROID_SDK_TAG\"\))#\1\n        PATH=\\\$(dirname \\\$SWIFT_EXECUTABLE_FOR_ANDROID_SDK):\\\$PATH\n        echo \"path is now \\\$PATH\"#" install-and-build-with-sdk.sh
           chmod 500 install-and-build-with-sdk.sh
           ./install-and-build-with-sdk.sh --android --build-command="swift build" --android-sdk-triple="${{ matrix.sdk_triple }}" --android-ndk-version="${{ matrix.ndk_version }}" ${{ matrix.swift_version }}
 


finagolfin:
@madsodgaard, the latest Feb. 24 trunk snapshot SDK is failing, seemingly in the Java build step, whereas the new 6.3 snapshot builds I'm adding in this pull all passed. Any idea what's going on?

finagolfin:
Maybe [this line](https://github.com/swiftlang/swift-java/actions/runs/22403048228/job/64854881232#step:4:11605)?
```
Error executing swiftc -print-target-info
```
I see it in multiple jobs with trunk, but not 6.3. 🤨

ktoso:
I'll see what I can find out here but cannot today sadly. Thank you for the PR!

finagolfin:
OK, rolling back to the container got the Android build working again, so I think I know what the problem is. Using a container like this works but is wasteful, as you're installing a container with the latest trunk snapshot toolchain pre-installed, then downloading the NDK and latest trunk snapshot toolchain _again_ on every CI run. I think the reason that works though is that the pre-installed toolchain is added to the `PATH`, whereas the one downloaded by the upstream script is not, but invoking `swiftc` directly from the `PATH` is apparently required by your plugin.

I'll try modifying the upstream script after downloading it in the CI here, so that it also adds the toolchain it downloads to the `PATH`: that should finally fix running in the GitHub runner with no container.

.github/workflows/pull_request.yml
```diff
@@ -184,33 +184,28 @@ jobs:
...
-          JAVA_HOME_17: ${{ env.JAVA_HOME_17 }}
-          JAVA_HOME_25: ${{ env.JAVA_HOME_25 }}
-          ANDROID_NDK_HOME: ${{ env.ANDROID_NDK_HOME }}
         run: |
           apt-get -q update && apt-get -yq install curl
           cd Samples/SwiftJavaExtractJNISampleApp
-          curl -s --retry 3 https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/install-and-build-with-sdk.sh | \
-          bash -s -- --android --build-command="swift build" --android-sdk-triple="${{ matrix.sdk_triple }}" --android-ndk-version="${{ matrix.ndk_version }}" ${{ matrix.swift_version }}
+          curl -L -O --retry 3 https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/install-and-build-with-sdk.sh
+          perl -pi -e "s#(download_and_extract_toolchain \".ANDROID_SDK_TAG\"\))#\1\n        PATH=\\\$(dirname \\\$SWIFT_EXECUTABLE_FOR_ANDROID_SDK):\\\$PATH\n        echo \"path is now \\\$PATH\"#" install-and-build-with-sdk.sh
```
finagolfin, line 206: Alright, finally got this working in a GitHub runner without needing a Swift container image, by modifying the official workflow script with this regex that produces this diff:
```
--- install-and-build-with-sdk.sh                         +++ new.sh    
@@ -616,6 +616,8 @@
         log "Installing Swift toolchain to match Android Swift SDK snapshot: $ANDROID_SDK_TAG"
         initialize_os_info
         SWIFT_EXECUTABLE_FOR_ANDROID_SDK=$(download_and_extract_toolchain "$ANDROID_SDK_TAG")
+        PATH=$(dirname $SWIFT_EXECUTABLE_FOR_ANDROID_SDK):$PATH
+        echo "path is now $PATH"
         if [[ $? -eq $EXIT_TOOLCHAIN_NOT_FOUND ]]; then
             # Don't fail the workflow if we can't find the right toolchain
             exit 0
```
`SWIFT_EXECUTABLE_FOR_ANDROID_SDK` is the full path to the Swift toolchain that the official script uses, so this regex merely adds it to the `PATH` also, as this swift-java plugin requires.

I don't know why the new 6.3 snapshot testing I added in this pull never failed, @madsodgaard, does 6.3 not have the swift-java plugin enabled or something, ie some code in this repo checks that the SDK is from trunk and only runs then?

finagolfin:
This is ready to go: I will make any final changes wanted by reviewers, such as changing the new Android testing matrix for the config you want, and squash the commits, before we can get it in.

.github/workflows/pull_request.yml
ktoso: Ah I see, thank you

ktoso: 6.3 shouldn't be any different tbh, so that's a bit weird...

ktoso (approved):
Thank you, solution looks ok enough to me, thank you!

PR #575, opened 2026-02-26
Escape generated JNI for `@JavaImplementation` methods

colemancda:
Fix generated JNI for `@JavaImplementation` methods by escaping underscores and other characters correctly.

### DIFF
--- a/Tests/SwiftJavaMacrosTests/JavaImplementationMacroTests.swift
+++ b/Tests/SwiftJavaMacrosTests/JavaImplementationMacroTests.swift
@@ -1,0 +1,190 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftJavaMacros
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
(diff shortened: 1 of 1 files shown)


Sources/SwiftJavaMacros/ImplementsJavaMacro.swift
```diff
@@ -16,10 +16,39 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
 
-enum JavaImplementationMacro {}
+package enum JavaImplementationMacro {}
+
+// JNI identifier escaping per the JNI specification:
+// https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#resolving_native_method_names
+private extension String {
```
ktoso: Can you move the private on the method rather than the extension? That'll unbreak the formatting ci check

ktoso:
```suggestion
extension String {
```

Sources/SwiftJavaMacros/ImplementsJavaMacro.swift
```diff
@@ -16,10 +16,39 @@ import SwiftSyntax
...
+  /// Returns the string with characters escaped according to JNI symbol naming rules.
+  /// - `_` → `_1`
+  /// - `.` and `/` → `_` (package/class separator)
+  /// - `;` → `_2`
+  /// - `[` → `_3`
+  /// - Non-ASCII → `_0XXXX` (UTF-16 hex)
+  var escapedJNIIdentifier: String {
```
ktoso, line 30:
```suggestion
  private var escapedJNIIdentifier: String {
```

Sources/SwiftJavaMacros/ImplementsJavaMacro.swift
ktoso: This is to resolve the CI failure, we put access control on members, not extensions

ktoso:
Thank you!

PR #576, opened 2026-02-27
jextract: Support fully qualified type names

sidepelican:
Currently, types containing a module prefix are not recognized correctly.
This is because it searches for the type including that module name.

```swift
public struct MyValue {}

// Not exported
public func fullQualifiedType() -> MyModule.MyValue
```

By utilizing the symbol table, I will change it so that if a known imported type name appears first, it can be searched by ignoring that part.

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
@@ -675,7 +675,7 @@ extension FFMSwift2JavaGenerator {
           throw JavaTranslationError.unhandledType(swiftType)
         }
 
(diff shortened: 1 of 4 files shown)


Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -675,7 +675,7 @@ extension FFMSwift2JavaGenerator {
           throw JavaTranslationError.unhandledType(swiftType)
         }
 
-        let javaType: JavaType = .class(package: nil, name: swiftNominalType.nominalTypeDecl.name)
+        let javaType: JavaType = .class(package: nil, name: swiftNominalType.nominalTypeDecl.qualifiedName)
```
sidepelican, line 678: This is a secondary bug found incidentally🤣

PR #577, opened 2026-02-27
jextract/jni: Support enums that have no cases

sidepelican:
Currently, when the Swift enum has no cases, the generated Java code results in a compilation error.

```
error: sealed class must have subclasses
  public sealed interface Case {}
                ^
```

In Swift, it is a common idiom to use enums with no cases as namespaces.
In cases where no enum cases exist, I will remove `sealed` to ensure that no compilation errors occur.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
@@ -296,7 +296,7 @@ extension JNISwift2JavaGenerator {
     ) { printer in
       let selfPointer = self.printSelfJLongToUnsafeMutablePointer(
         &printer,
-        swiftParentName: type.swiftNominal.name,
+        swiftParentName: type.swiftNominal.qualifiedName,
         selfPointerParam
       )
       printer.printBraceBlock("switch (\(selfPointer).pointee)") { printer in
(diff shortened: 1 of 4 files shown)


ktoso (approved):

ktoso:
Thanks!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -296,7 +296,7 @@ extension JNISwift2JavaGenerator {
     ) { printer in
       let selfPointer = self.printSelfJLongToUnsafeMutablePointer(
         &printer,
-        swiftParentName: type.swiftNominal.name,
+        swiftParentName: type.swiftNominal.qualifiedName,
```
sidepelican, line 299: I have found another problem again

```
error: cannot find type 'Nested' in scope
544 |   assert(selfPointer != 0, "selfPointer memory address was null")
545 |   let selfBits$ = Int(Int64(fromJNI: selfPointer, in: env$))
546 |   guard let self$ = UnsafeMutablePointer<Nested>(bitPattern: selfBits$) else {
    |                                          `- error: cannot find type 'Nested' in scope
547 |     fatalError("self memory address was null in call to \(#function)!")
548 |   }
```

ktoso: Hah ouch

ktoso:
You know what, let's accept your other PR, it's simple enough. I don't think we need to over think this that much, not emitting the sealed interface will be fine

PR #578, opened 2026-02-27
add `AndroidCoreLibraryDesugaring` trait

madsodgaard:
The API 23 runtime checks we were doing for desugaring class name happens at compile time, where I thought they were at runtime. So, that means even on API 24 (where `Optional` was added`). The core library desugaring still uses the new class name.

So instead we introduce a trait `AndroidCoreLibraryDesugaring`, which will then enable this behaviour at compile time instead.


ktoso (changes requested):
Needs to be documented please

PR #580, opened 2026-03-02
Rename `self` parameter name to `selfPointer`

sidepelican:
This is a follow-up to the renaming discussion at:
https://github.com/swiftlang/swift-java/pull/572#discussion_r2850870776

I have changed the parameters described as `self: jlong` in various locations to `selfPointer`.

During the process, I felt that `SwiftSelfParameter.instance` having its own parameter name property was a source of confusion.
In Swift, unlike Go, function receiver names cannot be changed, so the parameter name property is redundant information. Therefore, I have also included a change to remove it.


ktoso:
Thank you!

PR #582, opened 2026-03-04
jni: Unify common methods using generic implementations

sidepelican:
## Motivation

As mentioned in the 'Others' section of PR #572, we currently have branching logic in the generator for common functions like `destroy`, depending on whether the target type is generic or not.

This branching is undesirable as it increases maintenance costs.
While PR #567 partially addressed this by introducing `.synthesizedFunction`, I have found a better approach, I propose it and solve the problem.

## Solution

Instead of generating type-specific code, providing a single generic implementation within the `SwiftJava` package for functions like `toString` and `destroy`.

Since `JNISwiftInstance` always including type metadata, I realized that we can utilize the technique proposed in [this comment](https://github.com/swiftlang/swift-java/issues/533#issuecomment-3894556886) to call a generic implementation.

```swift
@_cdecl("Java_org_swift_swiftkit_core_SwiftObjects_toString__JJ")
public func Java_org_swift_swiftkit_core_SwiftObjects_toString__JJ(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, selfPointer: jlong, selfTypePointer: jlong) -> jstring? {
  let selfTypeBits$ = Int(Int64(fromJNI: selfTypePointer, in: environment))
  guard let selfType$ = UnsafeRawPointer(bitPattern: selfTypeBits$) else {
    fatalError("selfType metadata address was null")
  }
  let typeMetadata = unsafeBitCast(selfType$, to: Any.Type.self)

  func perform<T>(as type: T.Type) -> jstring? {
    guard let self$ = UnsafeMutablePointer<T>(bitPattern: selfPointer) else {
      fatalError("self memory address was null")
    }
    return String(describing: self$.pointee).getJNIValue(in: environment)
  }
  return perform(as: typeMetadata)
}
```

### Pros:

* Significantly simplifies the code generator implementation.
* Greatly reduces the amount of generated code, leading to smaller binary sizes.
* Aligns the implementation more closely with the FFM side.

### Cons:

* Increase in runtime overhead:
    * Add calling to runtime functions (SIL's `open_existential_metatype`).
    * Loss of certain optimizations (e.g., inlining that might skip VWT lookups when individual implementations are generated).

Regarding the downsides, I believe the overhead is negligible compared to the inherent cost of JNI calls.

## Changes

Following this new direction, I have replaced the generated code for the following functions with generic implementations:

* `toString`
* `toDebugString`
* `destroy`
* `getDiscriminator`

Additionally, I have removed `.synthesizedFunction` as it is no longer needed.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -152,7 +152,7 @@ let package = Package(
     .library(
       name: "SwiftJava",
       type: .dynamic,
-      targets: ["SwiftJava"]
+      targets: ["SwiftJava", "SwiftJavaRuntimeSupport"]
     ),
 
     .library(
@@ -224,12 +224,6 @@ let package = Package(
       targets: [
         "SwiftJavaPlugin"
       ]
-    ),
-
-    // Support library written in Swift for SwiftKit "Java"
-    .library(
-      name: "SwiftJavaRuntimeSupport",
-      targets: ["SwiftJavaRuntimeSupport"]
     ),
 
     .library(
--- a/Samples/SwiftJavaExtractJNISampleApp/Package.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Package.swift
@@ -110,7 +110,6 @@ let package = Package(
       dependencies: [
         .product(name: "SwiftJava", package: "swift-java"),
         .product(name: "CSwiftJavaJNI", package: "swift-java"),
-        .product(name: "SwiftJavaRuntimeSupport", package: "swift-java"),
       ],
       exclude: [
         "swift-java.config"


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -717,42 +723,23 @@ extension JNISwift2JavaGenerator {
...
       printer.print("@Override")
       printer.printBraceBlock("public long $typeMetadataAddress()") { printer in
-        printer.print(
-          """
-          long self$ = this.$memoryAddress();
-          if (CallTraces.TRACE_DOWNCALLS) {
-            CallTraces.traceDowncall("\(type.swiftNominal.name).\(funcName)",
-                "this", this,
-                "self", self$);
-          }
```
sidepelican, line 731: `traceDowncall` internally calls `toString`, which in turn calls `$typeMetadataAddress`, creating an infinite loop.
Therefore, I'll omit `traceDowncall`.

ktoso: Thanks for the comment, yeah those can be annoying to catch sometimes 😓 Maybe worth putting a comment in source about this, so we know why it's not there?

Sources/SwiftJava/SynthesizedAPIs.swift
```diff
@@ -0,0 +1,105 @@
...
+  var _rawDiscriminator: Int32 { get }
+}
+
+@_cdecl("Java_org_swift_swiftkit_core_SwiftObjects_getRawDiscriminator__JJ")
+public func Java_org_swift_swiftkit_core_SwiftObjects_getRawDiscriminator__JJ(
+  environment: UnsafeMutablePointer<JNIEnv?>!,
+  thisClass: jclass,
+  selfPointer: jlong,
+  selfTypePointer: jlong
+) -> jint {
```
ktoso: We should be able to get away with implementing them as 

```swift
@JavaImplementation("org.swift.swiftkit.core.JavaObjects")
struct SwiftObjects { 
  @JavaStaticMethod
  func getRawDiscriminator(selfPointer: Long, selfTypePointer: Long) -> Int { ... }
}
```

since SwiftJava depends on the SwiftJavaMacros... I'd like to avoid not-using our own tools in this project, this way it's a good feedback mechanism if something is off or missing, and we avoid the typical mistakes with the cdecls etc.

I forget the exact syntax so sorry if the snippet isn't exactly right above here :)

Package.swift
```diff
@@ -311,6 +311,7 @@ let package = Package(
       exclude: ["swift-java.config"],
       swiftSettings: [
         .swiftLanguageMode(.v5),
+        .enableUpcomingFeature("ImplicitOpenExistentials"),
```
ktoso, line 308: 👍

Tests/JExtractSwiftTests/JNI/JNIToStringTests.swift
```diff
@@ -69,28 +49,8 @@ struct JNIToStringTests {
...
-      .swift,
-      detectChunkByInitialLines: 1,
-      expectedChunks: [
-        """
-        @_cdecl("Java_com_example_swift_MyType__00024toDebugString__J")
-        public func Java_com_example_swift_MyType__00024toDebugString__J(environment: UnsafeMutablePointer<JNIEnv?>!, thisClass: jclass, selfPointer: jlong) -> jstring? {
-          ...
-          return String(reflecting: selfPointer$.pointee).getJNIValue(in: environment)
+        public String toDebugString() {
+          return SwiftObjects.toDebugString(this.$memoryAddress(), this.$typeMetadataAddress());
```
ktoso, line 53: Overall I really like this shape, it's true as you say we take some performance hit here for the opening but perhaps it is worth it anyway.

Would you be able to do a quick benchmark before/after? We have benchmarks set up in the repo, should be quick to add and try.

Overall though I'm supportive of this change, it's a nice pattern we'll likely follow in many more places!

Sources/SwiftJava/SynthesizedAPIs.swift
sidepelican: I didn't realize that we can use SwiftJavaMacros here! This would be better.

sidepelican:
@ktoso  After rewriting the implementation to use SwiftJavaMacros, I added and ran some benchmarks.

The results showed a performance degradation than I had expected.
Curious about this, I investigated the cause and found that the current implementation of SwiftJavaMacros introduces significant overhead.
When I temporarily reverted the implementation and re-measured, the results returned to the expected range.


Before (main branch), sha: 0b01d82af7329e155f797843f7a96f52f0fba8a4

```
Benchmark                                          (dataSize)  (taskCount)  Mode  Cnt     Score     Error  Units
GenericBenchmark.basicStruct_basicGetter                  N/A          N/A  avgt   30     9.189 ±   0.028  ns/op
GenericBenchmark.basicStruct_toString                     N/A          N/A  avgt   30  1837.659 ±  15.291  ns/op
GenericBenchmark.genericStruct_basicGetter                N/A          N/A  avgt   30    40.452 ±   0.409  ns/op
GenericBenchmark.genericStruct_toString                   N/A          N/A  avgt   30  1949.575 ±  13.998  ns/op
```

After (Using SwiftJavaMacros), sha: be833a470928903f677ff14c70499b51421101dd

```
Benchmark                                          (dataSize)  (taskCount)  Mode  Cnt     Score      Error  Units
GenericBenchmark.basicStruct_basicGetter                  N/A          N/A  avgt   30     9.365 ±    0.151  ns/op
GenericBenchmark.basicStruct_toString                     N/A          N/A  avgt   30  2599.917 ±   31.407  ns/op
GenericBenchmark.genericStruct_basicGetter                N/A          N/A  avgt   30    42.743 ±    1.196  ns/op
GenericBenchmark.genericStruct_toString                   N/A          N/A  avgt   30  2687.621 ±  192.211  ns/op
```

After (without SwiftJavaMacros), sha: c87e3904513822ed1a59f440d780ee9cd44820a4

```
Benchmark                                          (dataSize)  (taskCount)  Mode  Cnt     Score     Error  Units
GenericBenchmark.basicStruct_basicGetter                  N/A          N/A  avgt   30     9.303 ±   0.048  ns/op
GenericBenchmark.basicStruct_toString                     N/A          N/A  avgt   30  1911.742 ±  19.403  ns/op
GenericBenchmark.genericStruct_basicGetter                N/A          N/A  avgt   30    44.299 ±   2.601  ns/op
GenericBenchmark.genericStruct_toString                   N/A          N/A  avgt   30  1920.005 ±  17.951  ns/op
```

~~While we might be able to resolve this by improving the SwiftJavaMacros implementation, I have reverted the changes for now to prioritize performance. I'm open to keeping the SwiftJavaMacros version and improving it later, but what are your thoughts?~~

sidepelican:
Sorry, I realized I was misusing SwiftJavaMacros.
I mistakenly implemented a static function as an instance function.
After fixing that, it performed at an ideal speed.

After (with SwiftJavaMacro, rev2), sha: ff4e8e766c144fd5d25df3b4f7945a5239e65d30

```
Benchmark                                          (dataSize)  (taskCount)  Mode  Cnt     Score     Error  Units
GenericBenchmark.basicStruct_basicGetter                  N/A          N/A  avgt   30     9.321 ±   0.034  ns/op
GenericBenchmark.basicStruct_toString                     N/A          N/A  avgt   30  1909.003 ±  10.223  ns/op
GenericBenchmark.genericStruct_basicGetter                N/A          N/A  avgt   30    43.091 ±   1.101  ns/op
GenericBenchmark.genericStruct_toString                   N/A          N/A  avgt   30  1924.357 ±   9.108  ns/op
```

ktoso:
Sweet, yes they should perform as well as manually writing these for the most part. Thank you! Giving it a final look then

ktoso:
Seems we have a real runtime crash sadly

sidepelican:
Currently, many warnings like the following are being displayed when running tests:

> objc[75138]: Class _TtC9SwiftJava6Writer is implemented in both /Users/iceman/github/swift-java/.build/arm64-apple-macosx/debug/libSwiftJava.dylib (0x105901658) and /Users/iceman/github/swift-java/Samples/SwiftJavaExtractJNISampleApp/.build/arm64-apple-macosx/debug/libMySwiftLibrary.dylib (0x125040ba8). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.

I believe this is caused by the coexistence of the dynamic `SwiftJava` package and `MySwiftLibrary`, which has a static dependency (via `SwiftJavaRuntimeSupport`) on the same `SwiftJava` package.

As a result, `EnumWithValueCases` might be failing to cast because it conforms to the **static** version of `_RawDiscriminatorRepresentable` but not the **dynamic** one.

ktoso:
Hm, worth a try as a quick fix. We can follow up with refactoring the modules perhaps...

sidepelican:
As a simple fix, I merged `SwiftJavaRuntimeSupport` into `libSwiftJava` .
I considered several other options, but they didn't really work out for various reasons.

ktoso:
Hm, I'm not sure about this... @madsodgaard wdyt? The runtime funcs split was to be able to depend only on those I guess, idk if we truly do so anywhere

madsodgaard:
I don't think we are using the separate product for anything. I think its a legacy thing from when we were experiementing with dynamic products, but now that `SwiftJava` is dynamic, I think its fine to move the runtime support product into `SwiftJava`.

Though it is a breaking change, so should probably highlight it.

ktoso:
We're not guaranteeing stability yet so I'd rather do changes like this aggressively right now rather than after we tag and release.

Let's go with this then, thanks for the great work here @sidepelican !

PR #583, opened 2026-03-05
Split off swift-java-jni-core

ktoso:
Replaces https://github.com/swiftlang/swift-java/pull/384 and makes this repo depend on github.com/swiftlang/swift-java-jni-core.


ktoso:
Let's land this step, and I think we'll be doing some more reorganization in the near term

Issue #586, opened 2026-03-05
Carry over Java `@RequiresApi` and `@Deprecated` annotations to generated Swift JNI code


colemancda:
When generating Swift JNI wrappers from Java source, annotations from
`android.support.annotation` (and `androidx.annotation`) that carry API
level requirements or deprecation status should be translated to their
Swift equivalents rather than silently dropped.

## Motivation

Currently, semantic annotations on Java methods are lost during Swift
codegen. This means callers on the Swift side lose compile-time
enforcement of minimum API levels and deprecation warnings — safety
guarantees that exist on the Java side simply vanish.

## Example

**Java source:**

```java
@RequiresApi(api = Build.VERSION_CODES.R)  // API 30
void myFunction() { ... }

@Deprecated
void legacyFunction() { ... }
```

**Current generated Swift (annotations dropped):**

```swift
func myFunction() { ... }
func legacyFunction() { ... }
```

**Expected generated Swift:**

```swift
@available(Android 30, *)
func myFunction() { ... }

@available(*, deprecated)
func legacyFunction() { ... }
```

## Requested Annotation Mappings

|Java Annotation                            |Generated Swift                              |
|-------------------------------------------|---------------------------------------------|
|`@RequiresApi(api = N)`                    |`@available(Android N, *)`                   |
|`@RequiresApi(api = Build.VERSION_CODES.X)`|Resolve constant → `@available(Android N, *)`|
|`@Deprecated`                              |`@available(*, deprecated)`                  |
|`@Deprecated` + Javadoc `@deprecated <msg>`|`@available(*, deprecated, message: "<msg>")`|

## Notes

- `Build.VERSION_CODES` constants should be resolved to their integer
  values during codegen (e.g. `R` → `30`, `S` → `31`, `TIRAMISU` → `33`)
- If the Javadoc `@deprecated` tag is present alongside `@Deprecated`,
  its description should populate the `message:` parameter of `@available`
- Both `android.support.annotation.RequiresApi` and
  `androidx.annotation.RequiresApi` should be handled

ktoso:
Very neat idea, thanks for the issue

ktoso:
We can't handle javadoc btw

> If the Javadoc @deprecated tag is present alongside @Deprecated,


at last not yet, it'd be a big new feature to ingest javadocs, since we're working on classes/bytecode. But the other parts of the idea we can do

ktoso:
[...]

PR #587, opened 2026-03-06
wrap-java: Import deprecated and requires-api annotations

ktoso:
We now handle Java Deprecated and also Android's RequiresApi annotations and import them as availability annotations on Swift side.

The Android availability requires at least Swift 6.3

resolves https://github.com/swiftlang/swift-java/issues/586

Claude assisted boilerplate and tests; refactored a bit how we do ThreadSafe as well, there's a new enum for "known" java annotations

Methods end up as

```
@available(*, deprecated)
  #if compiler(>=6.3)
  @available(Android 28 /* Pie */, *)
  #endif
  @JavaMethod
  open func oldApi28Method()
```

need to do the compiler check since Android availability is only since 6.3.

@colemancda Would love to see https://github.com/PureSwift/Android/ re-generated with these :-)

### DIFF
--- a/Sources/SwiftJavaToolLib/JavaClassTranslator.swift
+++ b/Sources/SwiftJavaToolLib/JavaClassTranslator.swift
@@ -645,20 +645,24 @@ extension JavaClassTranslator {
         let apiLevel = try? annotation.as(JavaObject.self)
           .dynamicJavaMethodCall(methodName: "api", resultType: Int32.self)
         if let apiLevel, apiLevel > 1 {
-          result.attributes.append(SwiftAttribute(
-            value: "@available(Android \(apiLevel)\(apiLevelComment(apiLevel)), *)",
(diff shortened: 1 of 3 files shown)


Sources/SwiftJavaToolLib/JavaClassTranslator.swift
```diff
@@ -590,18 +591,98 @@ extension JavaClassTranslator {
...
+    var attributes: [SwiftAttribute] = []
+
+    func render() -> String {
+      if attributes.isEmpty {
+        return ""
+      }
+      var lines: [String] = []
+      for attr in attributes {
+        if let (major, minor) = attr.minimumCompilerVersion {
+          lines.append("#if compiler(>=\(major).\(minor))")
```
ktoso, line 615: I think that's the right check we need

PR #589, opened 2026-03-06
wrap-java: Basic classfile parser to get CLASS retained annotations 

ktoso:
Implement basic class file parsing, only enough to get annotations

Thankfully the classfile format is well known and documented.

We avoid the usual "Java way" of doing this work which would be to pull
in the asm.jar dependency, but instead choose to do the parsing
ourselfes.

The format is well specified and stable, so we can be pretty confident
in it. I based it on the JDK22 revision of the spec: https://docs.oracle.com/javase/specs/jvms/se22/html/jvms-4.html

We only forus on the "RuntimeInvisibleAannotations" attribute because
runtime visible ones we're able to get from plain reflection calls
(which the Deprecated test cases showcase already in previous PR): https://docs.oracle.com/javase/specs/jvms/se22/html/jvms-4.html#jvms-4.7.17

This is necessary to support Android's RequiredApi annotations which are
CLASS retained (or how kotlin confusingly calls it BINARY which mislead
me and I thouhgt they're retained but yeah makes sense).

This way we're able to emit Android availability annotations for the
whole Android SDK.

### DIFF
--- a/Sources/SwiftJavaToolLib/JavaClassTranslator.swift
+++ b/Sources/SwiftJavaToolLib/JavaClassTranslator.swift
@@ -654,70 +654,42 @@ extension JavaClassTranslator {
 
   /// Build Swift `@available` attributes from Java annotations on a reflective element.
   private func swiftAvailableAttributes(
-    from annotations: [Annotation],
+    from runtimeAnnotations: [Annotation],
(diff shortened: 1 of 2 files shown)


ktoso:
resolves https://github.com/swiftlang/swift-java/issues/586

Issue #592, opened 2026-03-06
Use Android's api-versions.xml to inform availability of Android APIs


ktoso:
We'd need a special feature in wrap-java to provide the `--android-api-versions platforms/android-34/data/api-versions.xml` and then parse that, and use that as information for the levels.

Turns out `@RequiresApi` is a lie and does not exist at all in any APIs in android.jar huh

PR #593, opened 2026-03-08
Add --android-api-version-file support for availability

ktoso:
We parse the xml file and derive availability annotations for Android using it. Turns out the @RequiresApi annotations are a lie and are not present in classes in android.jar, even though it appears as if they are when browsing docs etc; instead, we had to rely on the version xml which is included in the android SDK.

resolves https://github.com/swiftlang/swift-java/issues/592

### DIFF
--- a/Sources/SwiftJavaDocumentation/Documentation.docc/Android.md
+++ b/Sources/SwiftJavaDocumentation/Documentation.docc/Android.md
@@ -21,4 +21,24 @@ let package = Package(
   ]
 )
 ```
+
+### Android SDK Availability
+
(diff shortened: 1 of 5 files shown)


ktoso:
Still need to do more cleanups here but this generally should work, I'll give it a go with an Android SDK as well before we get this in.

ktoso:
I also confirmed this actually works when doing a wrap-java on android.jar :-)

```
ktoso@cherudim [16:48:04] [~/code/swift-java-android-sdk]
-> % ack '@available\(Android' Sources/ | wc -l
   68752
```

Issue #599, opened 2026-03-10
Discrepancy between files in generated/ directories and wrap-java output


sidepelican:
I’ve noticed that the files in the `generated/` directories are currently out of sync with the output of the `wrap-java` command.
It appears that manual edits were made where the generator is lacking features, but this discrepancy makes it difficult to verify work on other issues like #169.

I believe the generated code should be fully reproducible by the tool to maintain a healthy codebase.

ktoso:
Yes, we should perhaps even CI validate this, thank you

ktoso:
Fantastic work here @sidepelican !

PR #600, opened 2026-03-10
Update JavaIterator implementation

sidepelican:
This is a first, very small step toward the improvements in #599.
I have updated the implementation of `JavaIterator.swift` using the wrap-java command and fixed the parts that required attention.


ktoso (approved):

ktoso:
Thanks! agreed that we should be able to always wrap-java generate and keep those stable basically

PR #601, opened 2026-03-10
javakit-macros: Support wrapping generic arguments

sidepelican:

Previously, generic method arguments (such as `T`) had to be manually replaced with `JavaObject`.
I have updated the `@JavaMethod` macro to handle this automatically.
The macro now uses a heuristic search to determine if a type is generic and inserts the necessary conversion to `JavaObject`.

Additionally, I have updated the implementation of `JavaOptional.swift` and (partially) `List.swift` using `wrap-java`.

### DIFF
--- a/Sources/SwiftJavaMacros/JavaMethodMacro.swift
+++ b/Sources/SwiftJavaMacros/JavaMethodMacro.swift
@@ -71,7 +71,7 @@ extension JavaMethodMacro: BodyMacro {
     var paramNames: [String] = []
     for param in params {
       guard let name = param.parameterName else {
-        throw MacroErrors.parameterMustHaveName
+        throw MacroErrors.parameterMustHaveName(method: funcName, paramSyntax: param.trimmedDescription)
       }
       if isJNIGenericParameter(param.type, funcDecl: funcDecl, in: context) {
         let erasedName: TokenSyntax = "\(name)$erased"
@@ -183,8 +183,9 @@ extension JavaMethodMacro: BodyMacro {
     return resultStatements
   }
 
-  /// Determines whether an argument is generic.
+  /// Determines whether an argument is generic in heuristic way.
   /// Since Optional does not appear in JNI signatures, it is removed before checking.
+  /// FIXME: It might be preferable to explicitly specify the type from JavaClass, similar to `typeErasedResult`.
   private static func isJNIGenericParameter(
     _ type: TypeSyntax,
     funcDecl: FunctionDeclSyntax,
--- a/Sources/SwiftJavaMacros/MacroErrors.swift
+++ b/Sources/SwiftJavaMacros/MacroErrors.swift
@@ -21,5 +21,5 @@ enum MacroErrors: Error {
   case methodNotOnFunction
   case missingEnvironment
   case macroOutOfContext(String)
-  case parameterMustHaveName
+  case parameterMustHaveName(method: String, paramSyntax: String)
 }


Tests/SwiftJavaTests/BasicRuntimeTests.swift
sidepelican, line 107: Currently, this `add` function uses a generic implementation.

Sources/SwiftJavaMacros/JavaMethodMacro.swift
```diff
@@ -137,23 +160,59 @@ extension JavaMethodMacro: BodyMacro {
...
+      if genericParams.contains(where: { $0.name.text == typeName }) {
+        return true
+      }
+    }
+
+    for contextNode in context.lexicalContext {
+      if let decl = contextNode.asProtocol(WithGenericParametersSyntax.self) {
+        if decl.genericParameterClause?.parameters.contains(where: {
+          $0.name.text == typeName
+        }) == true {
```
ktoso, line 210: no need to `== true`
```suggestion
        }) {
```

Sources/SwiftJavaMacros/JavaMethodMacro.swift
```diff
@@ -137,23 +160,59 @@ extension JavaMethodMacro: BodyMacro {
...
-    // no return type conversions
-    return [
-      "return \(resultSyntax)"
-    ]
+    return resultStatements
+  }
+
+  /// Determines whether an argument is generic.
+  /// Since Optional does not appear in JNI signatures, it is removed before checking.
+  private static func isJNIGenericParameter(
```
ktoso, line 189: Let's document that this is a heuristic please, since it may get things wrong. It's good to mark this as source of potential mixups

Sources/SwiftJavaMacros/JavaMethodMacro.swift
```diff
@@ -65,7 +67,28 @@ extension JavaMethodMacro: BodyMacro {
 
     let isStatic = node.attributeName.trimmedDescription == "JavaStaticMethod"
     let params = funcDecl.signature.parameterClause.parameters
-    let paramNames = params.map { param in param.parameterName?.text ?? "" }.joined(separator: ", ")
+
+    var paramNames: [String] = []
+    for param in params {
+      guard let name = param.parameterName else {
+        throw MacroErrors.parameterMustHaveName
```
ktoso: Can this include the method and just whole "\(param)" we're failing on? better diagnostics in macro failures are always good

ktoso:
Minor requests here and there, looks good, thank you!

Sources/SwiftJavaMacros/JavaMethodMacro.swift
sidepelican: `genericParameterClause?` produces optional bool here.

Sources/SwiftJavaMacros/JavaMethodMacro.swift
sidepelican: Upon further review, I noticed that `typeErasedResult` references JavaClass information, which I believe is more preferable. I've added a comment regarding this point.

ktoso: nice, thank you

Sources/SwiftJavaMacros/JavaMethodMacro.swift
ktoso: ah i see

ktoso (approved):

Issue #596, opened 2026-03-09
jextract/jni: tuple support


ktoso:
We should be able to handle `-> (A, B)` as well as accepting tuples as parameters and fields etc.

PR #602, opened 2026-03-10
jextract/jni: Basic accept/return support for Swift tuples

ktoso:
Support for returning and accepting tuples in jextract/jni.

We implement this by passing single element arrays as indirect results into which the thunks initialize the values.

I somehow super messed up my local gradle config so didn't verify the sample :| I'm working on figuring what happened...

resolves https://github.com/swiftlang/swift-java/issues/596

Disclosure: boilerplate and general structure was generated using claude code. I have fully reviewed and polished up the output to a standard I deem acceptable for the project. The way we handle the out parameters is good as well, the array trick is good for these one off parameters of potentially different types. We could optimize this if we notice the tuple is of the same type and just use a single array then etc.


ktoso:
I'm sure this has edge cases I missed, e.g. arrays inside tuples, but let's take this incremental feature 👍

PR #613, opened 2026-03-12
wrap-java: Fix redundant method generation

sidepelican:

When generating code for `PrintWriter`, the following compilation error occurs:

```
.../PrintWriter.swift:136:13: error: overriding declaration requires an 'override' keyword
134 |   /// ```
135 |   @JavaMethod
136 |   open func append(_ arg0: CharSequence?) throws -> Appendable!
    |             `- error: overriding declaration requires an 'override' keyword
137 | 
138 |   /// Java method `append`.
```

This function does not exist for `PrintWriter` in Java. An overridden function `PrintWriter append(char c)` exists instead.

`MethodVariance` currently only evaluates subclass hierarchies, so it doesn't realize that a method in `PrintWriter` is actually a refinement of a method from a parent interface like `Appendable`.

I have updated the logic that determines method variance to consider general subtype relationships rather than just subclassing.

### DIFF
--- a/Tests/SwiftJavaToolLibTests/Java2SwiftTests.swift
+++ b/Tests/SwiftJavaToolLibTests/Java2SwiftTests.swift
@@ -468,10 +468,10 @@ class Java2SwiftTests: XCTestCase {
   }
 
   func testSuperTypeHasCovariantOverride() throws {
-    /// `Appendable` has `Appendable append(char c)`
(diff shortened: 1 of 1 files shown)


ktoso (approved):
Very nice work, thank you

sidepelican:
github is unstable today 😢

ktoso:
Yeah huh :)

PR #615, opened 2026-03-12
wrap-java: Fix incorrect generic parameter propagation

sidepelican:

When generating code for types like `ArrayDeque`, the following compilation error occurs:

```
.../ArrayDeque.swift:6:42: error: cannot specialize non-generic type 'JavaObject'
  4 | 
  5 | @JavaClass("java.util.ArrayDeque")
  6 | open class ArrayDeque<E: AnyJavaObject>: JavaObject</* MISSING MAPPING FOR */ E> {
    |                                          `- error: cannot specialize non-generic type 'JavaObject'
  7 |   @JavaMethod
  8 |   @_nonoverride public convenience init(_ arg0: Int32, environment: JNIEnvironment? = nil)
```

`ArrayDeque<E>` inherits from `AbstractCollection<E>`. In this context, `AbstractCollection` is not being scanned, so the generator skips it and falls back to the next ancestor, which is `Object`.

However, due to an oversight in JavaClassTranslator.swift, the generic parameter `E` from the skipped parent was incorrectly applied to the fallback `JavaObject`.
Since `JavaObject` is a non-generic type, this results the error shown above.

https://github.com/swiftlang/swift-java/blob/512672268028df1dffeb02e46814dd23545ffcf8/Sources/SwiftJavaToolLib/JavaClassTranslator.swift#L166

### DIFF
--- a/Sources/JavaStdlib/JavaUtil/generated/ArrayDeque.swift
+++ b/Sources/JavaStdlib/JavaUtil/generated/ArrayDeque.swift
@@ -10,111 +10,273 @@ open class ArrayDeque<E: AnyJavaObject>: JavaObject {
   @JavaMethod
   @_nonoverride public convenience init(environment: JNIEnvironment? = nil)
 
-  @JavaMethod
-  @_nonoverride public convenience init(_ arg0: JavaCollection<JavaObject>?, environment: JNIEnvironment? = nil)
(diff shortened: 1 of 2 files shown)


ktoso (commented):
Thank you, I remember hitting that and forgot to make an issuue about it i guess :)

PR #616, opened 2026-03-12
Cleanup how we handle local references in JNI interactions

madsodgaard:
We were hitting "JNI ERROR (app bug): local reference table overflow (max=512)". 

In order to improve this, I've added some `DeleteLocalRefs` and explicit `Push/Pop` of local frames to SwiftJava where I saw fit.

These changes made our app stopping crashing with that error.

This also modifies the "implement Swift protocols in Java" generated wrappers, such that all calls are wrapped in a local frame to ensure that resources are cleaned up, as this might not be called from a JVM thread and be auto-cleaned.


Sources/SwiftJava/JavaObject+MethodCalls.swift
```diff
@@ -153,7 +155,11 @@ extension AnyJavaObject {
   ) throws -> Result {
     // Retrieve the method that performs this call, then package the values and
     // call it.
+    // Size the frame to 1 ref per argument (strings/objects each need 1;
+    // primitives need 0, so this slightly over-estimates) plus 1 for the result ref.
     let jniMethod = Result.jniMethodCall(in: environment)
+    environment.interface.PushLocalFrame(environment, Int32(countArgs(repeat each args) + 2))
+    defer { environment.interface.PopLocalFrame(environment, nil) }
```
ktoso: I see, hope this'll prove helpful enough

ktoso:
I'll give this a closer look actually tomorrow

Sources/SwiftJava/JavaObject+MethodCalls.swift
ktoso: The nil is scary actually, lets think more about it

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -171,6 +171,19 @@ extension JNISwift2JavaGenerator {
...
+      // Push a local JNI frame so refs created during this upcall are freed on exit.
+      // When called from a Swift async context (e.g. cooperative thread pool) there is
+      // no enclosing JNI frame, so refs would otherwise accumulate indefinitely. When
+      // called from a Java-initiated native call there is already a frame, but pushing
+      // a sub-frame still frees refs earlier and prevents overflow within a single call.
+      let paramCount = function.originalFunctionSignature.parameters.count
+      printer.print(
+        """
+        let environment$ = try! JavaVirtualMachine.shared().environment()
+        environment$.interface.PushLocalFrame(environment$, \(paramCount * 2 + 4))
```
ktoso: we should be checking the result of the push, we may be OOMing the JVM here, and we'd pop the wrong frame because we didn't push one (!)

> jint PushLocalFrame(JNIEnv *env, jint capacity);
> 
> Creates a new local reference frame, in which at least a given number of local references can be created. **Returns 0 on success, a negative number and a pending OutOfMemoryError on failure.**

`    if pushed != JNI_OK `

and need to guard the pop the same...

We may suggest doing a withFrame pattern tbh: https://github.com/swiftlang/swift-java-jni-core/pull/11

Sources/SwiftJava/JavaObject+MethodCalls.swift
```diff
@@ -219,7 +225,10 @@ extension AnyJavaObject {
   ) throws {
     // Retrieve the method that performs this call, then package the arguments
     // and call it.
+    // Size the frame to 1 ref per argument; no result ref needed for void calls.
     let jniMethod = environment.interface.CallVoidMethodA!
+    environment.interface.PushLocalFrame(environment, Int32(countArgs(repeat each args) + 1))
```
ktoso: Again we need to check the result type, so I think we need a with... pattern here - proposing one here https://github.com/swiftlang/swift-java-jni-core/pull/11

Sources/SwiftJava/JavaObject+MethodCalls.swift
ktoso: Okey tne nil is correct but commented elsewhere about frame management

madsodgaard:
@ktoso I updated to use the new frame helpers when we generate protocol wrappers. On top of that, I actually removed the frame helpers from the method calls. Those should be handled by the other `DeleteLocalRef`s, and were probably the ones I was most doubting the correctness of, so let's leave them out for now.

I've tested these changes on our Android devices, and it no longer crashes 👍

ktoso:
Renamed the PR for future changelog, sorry :)

PR #617, opened 2026-03-13
Remove duplicate Java wrapper classes

sidepelican:
## Motivation

I noticed that `java.lang.reflect.Type` and `java.lang.reflect.ParameterizedType` were generated in both the `SwiftJava` and `JavaLangReflect` packages.
To avoid confusion, I want to ensure that these are generated into a single package.

## Changes

I have removed these classes from `SwiftJava`.
All classes under the `java.lang.reflect` namespace are moved into `JavaLangReflect` package.


Sources/JavaStdlib/JavaLangReflect/generated/Array.swift
```diff
@@ -0,0 +1,198 @@
+// Auto-generated by Java-to-Swift wrapper generator.
+import SwiftJava
+import SwiftJavaJNICore
+
+@JavaClass("java.lang.reflect.Array")
+open class Array: JavaObject {
```
ktoso, line 6: I wonder if we should call it JavaArray hmmm

Sources/SwiftJava/generated/JavaClass.swift
```diff
@@ -1,159 +1,466 @@
 // Auto-generated by Java-to-Swift wrapper generator.
 import SwiftJavaJNICore
 
-@JavaClass("java.lang.Class", implements: JavaReflectType.self)
+@JavaClass("java.lang.Class")
```
ktoso, line 4: It's not great that we lost this implements here

Sources/SwiftJava/generated/JavaClass.swift
```diff
@@ -1,159 +1,466 @@
...
+  /// public boolean java.lang.Class.isSynthetic()
+  /// ```
   @JavaMethod
   open func isSynthetic() -> Bool
 
-  @JavaMethod
-  open func getGenericSuperclass() -> JavaReflectType!
-
-  @JavaMethod
-  open func getGenericInterfaces() -> [JavaReflectType?]
```
ktoso, line 115: seems we lost this?

Sources/SwiftJavaToolLib/JavaClassTranslator.swift
```diff
@@ -141,13 +141,13 @@ struct JavaClassTranslator {
...
       var javaSuperclass = javaClass.getSuperclass()
-      var javaGenericSuperclass: JavaReflectType? = javaClass.getGenericSuperclass()
+      var javaGenericSuperclass: Type? = javaClass.getGenericSuperclass()
       var swiftSuperclassName: String? = nil
       var swiftSuperclassTypeArgs: [String] = []
       while let javaSuperclassNonOpt = javaSuperclass {
         do {
           swiftSuperclassName = try translator.getSwiftTypeName(javaSuperclassNonOpt, preferValueTypes: false).swiftName
-          if let javaGenericSuperclass = javaGenericSuperclass?.as(JavaReflectParameterizedType.self) {
+          if let javaGenericSuperclass = javaGenericSuperclass?.as(ParameterizedType.self) {
```
ktoso, line 150: This one comes from where now?

sidepelican: `ParameterizedType` was already present in the `JavaLangReflect` package.

Sources/SwiftJava/generated/JavaClass.swift
sidepelican: These were manually moved to `JavaClass+Reflection.swift`.

Sources/SwiftJavaToolLib/JavaClassTranslator.swift
ktoso: Gotcha thanks, the types inside SwiftJava are generally there so we could bootstrap the tool and generate the other JavaStdlib... perhaps we can get away from that actually... I'll think about it some more but PR looks ok

Sources/SwiftJava/generated/JavaClass.swift
sidepelican: `GenericDeclaration` wasn't originally included here, so this took the same form.
I understand this isn't ideal.

PR #618, opened 2026-03-13
wrap-java: Update subtyping logic for generic method overrides

sidepelican:
This PR is likely the final part of the improvements for #599.

In Java,

* `Executable` defines: `Class<?> getDeclaringClass()`
* `Constructor` overrides it as: `Class<T> getDeclaringClass()`

Currently, `wrap-java` identifies these as the same method signature and adds the `override` keyword in Swift.
However, since Swift does not recognize a subtype relationship about them(`JavaClass<JavaObject>` vs `JavaClass<T>`), it results in the following error:

```
.../Constructor.swift:113:22: error: method does not override any method from its superclass
111 |   /// ```
112 | @JavaMethod
113 |   open override func getDeclaringClass() -> JavaClass<T>!
    |                      `- error: method does not override any method from its superclass
114 | 
115 |   /// Java method `getAnnotation`.
```

To resolve this, I have updated the type-checking logic to be more strict:

* In `Type.isEqualTo(_:)`, resolved an issue where `Class<?>` and `Class<T>` were incorrectly treated as identical types.
    * Removed the call to `adjustToJavaBounds`
* In `Type.isEqualToOrSubtypeOf(_:)`, changed the logic so that types with different generic arguments are no longer considered to have a subtype relationship.

### DIFF
--- a/Sources/SwiftJavaToolLib/JavaType+Equality.swift
+++ b/Sources/SwiftJavaToolLib/JavaType+Equality.swift
@@ -121,6 +121,22 @@ extension Type {
       return selfClass.isSubclass(of: otherClass)
     }
 
+    if let selfType = self.as(ParameterizedType.self),
+      let otherType = other.as(ParameterizedType.self)
+    {
+      if !selfType.getRawType().isEqualToOrSubtypeOf(otherType.getRawType()) {
+        return false
+      }
+
+      if selfType.getActualTypeArguments().allTypesEqual(otherType.getActualTypeArguments()) {
+        return true
+      }
+
+      // Even where Java allows subtyping (e.g., Box<?> from Box<T>),
+      // this cannot be a subtype relationship in Swift.
+      return false
+    }
+
     // Anything object-like is a subclass of java.lang.Object
     if let otherClass = other.as(JavaClass<JavaObject>.self),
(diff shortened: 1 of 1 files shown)


Sources/SwiftJavaToolLib/JavaGenericsSupport.swift
```diff
@@ -41,7 +41,7 @@ func getGenericJavaTypeOriginInfo(_ type: Type?, from method: Method) -> [Generi
     return [] // it's not a generic type, no "origin" of the use to detect
   }
 
-  var methodTypeVars = method.getTypeParameters()
+  var methodTypeVars = method.getTypeParameters() as [TypeVariable<JavaLangReflect.Method>?]
```
sidepelican, line 44: Unfortunately, The override being treated as overload, explicit type annotations are now required. I believe this is unavoidable.

ktoso: There's probably no great way to avoid this... Unless we suffix the methods in the subclass? It might be an inherent incompatibility of how swift and java do overrides though

Open to sneaky ideas, maybe @DougGregor might have some?

ktoso:
Let's take this and we can keep thinking about overrides

PR #620, opened 2026-03-14
jextract: fix conditional imports

madsodgaard:
Fixes a bug where the generator would generate:
```
#if canImport(FoundationEssentials)
import FoundationEssentials
#else
import Foundation
#endif
#if canImport(Foundation)
import Foundation
#else
import FoundationEssentials
#endif
```


ktoso (approved):
Oh good catch!

Issue #423, opened 2025-11-05
JavaStdlib/JavaNet.URLClassLoader must inherit ClassLoader


ktoso:
We're missing methods otherwise, like `loadClass` that we need to use during certain wrap-java tasks

PR #621, opened 2026-03-16
Resolve manual workarounds for URLClassLoader generation

sidepelican:
Resolves: https://github.com/swiftlang/swift-java/issues/423

This PR reorganizes the target assignments for several classes to enable the code generation of `URLClassLoader` using only the wrap-java command.

Unlike Java, which allows bidirectional imports between packages, Swift targets cannot have circular dependencies.
To avoid dependency loops, I’ve had to consolidate these core classes into SwiftJava target.

Relocations: 

| java class | before Swift target | after |
|----|----|----|
| `java.net.URL` | JavaNet |  SwiftJava |
| `java.util.Enumeration` | JavaUtil |  SwiftJava |


ktoso (approved):
LGTM

ktoso:
Thanks for the ongoing work here!

Issue #612, opened 2026-03-11
jextract/jni: Support native Swift Set use from Java


ktoso:
Support Set types using a wrapper type and downcalls

PR #622, opened 2026-03-16
Native Swift Set support for jextract/jni

ktoso:
This provides `SwiftSet` java wrapper which refers to the Swift Set in native memory, unless explicitly copied. This implements the Set interface in Java.

Resolves https://github.com/swiftlang/swift-java/issues/612

Disclosure: this was AI assisted and based on the work on Dictionary, since it is very similar the results are also very similar for Set as they are for Dictionary. I've thoroughly reviewed the code and wrote parts of it.

### DIFF
--- a/Sources/SwiftJava/BridgedValues/SwiftSet+JNI.swift
+++ b/Sources/SwiftJava/BridgedValues/SwiftSet+JNI.swift
@@ -1,0 +1,59 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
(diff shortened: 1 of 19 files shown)


Sources/SwiftJava/BridgedValues/SwiftSet+JNI.swift
```diff
@@ -0,0 +1,59 @@
...
+@JavaImplementation("org.swift.swiftkit.core.collections.SwiftSet")
+extension SwiftSetJava {
+
+  private static func setBox(from pointer: Int64) -> AnySwiftSetBox {
+    let rawPointer = UnsafeRawPointer(bitPattern: Int(pointer))!
+    return Unmanaged<AnySwiftSetBox>.fromOpaque(rawPointer).takeUnretainedValue()
+  }
+
+  @JavaMethod("$size")
+  public static func _setSize(environment: UnsafeMutablePointer<JNIEnv?>!, pointer: Int64) -> Int32 {
```
ktoso, line 30: note that we must use _set here otherwise we'll nameclash with the dictionary one as those become global funcs... I'll add docs about this

PR #624, opened 2026-03-16
Introduce subcommand for jextract to prevent linking Foundation

### LINKED ISSUE #619: Adding `JExtractSwiftPlugin` links Full Foundation
When adding the `JExtractSwiftPlugin` to a package, it has the side-effect of making the binary link against full `Foundation`:
```
U static Foundation.Bundle.main.getter : Foundation.Bundle
U Foundation.Bundle.__allocating_init(path: Swift.String) -> Foundation.Bundle?
U type metadata accessor for Foundation.Bundle
```

I am guessing this is due to some SPM plugin file copying or something?

madsodgaard:
Resolves #619 

Specifying any non Swift files as build plugin outputs, will make SPM generate a bundle accessor file which imports Foundation unconditionally.

To solve this, we introduce a new subcommand that handles the callback part of the JExtract build plugin.


Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -139,7 +139,11 @@ struct JExtractSwiftBuildToolPlugin: SwiftJavaPluginProtocol, BuildToolPlugin {
       arguments += [
         "--generated-java-sources-list-file-output", javaSourcesListFileName,
       ]
-      jextractOutputFiles += [javaSourcesFile]
+      // NOTE: javaSourcesFile is intentionally NOT added to jextractOutputFiles.
+      // Adding a non-Swift file as a build command output causes SPM to bundle it
+      // as a module resource, which triggers resource_bundle_accessor.swift generation
```
ktoso, line 144: Ugh oh no... ;-(

Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -148,7 +152,7 @@ struct JExtractSwiftBuildToolPlugin: SwiftJavaPluginProtocol, BuildToolPlugin {
         executable: toolURL,
         arguments: arguments,
         inputFiles: [configFile] + swiftFiles,
-        outputFiles: jextractOutputFiles
+        outputFiles: jextractOutputFiles,
```
ktoso, line 155: pretty sure formatter will dislike this?

madsodgaard: it passed 🤷

ktoso: okey I'm honestly confused with swift-format sometimes... 😅

Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -188,85 +192,48 @@ struct JExtractSwiftBuildToolPlugin: SwiftJavaPluginProtocol, BuildToolPlugin {
...
+    // declares only a Swift file as its output. This avoids SPM treating any
+    // intermediate artifact (Gradle output directories, compiled .class files,
+    // swift-java.config) as module resources, which would trigger
+    // resource_bundle_accessor.swift generation and pull Foundation.Bundle into
+    // the target binary.
+    //
+    // inputFiles includes the Swift outputs from jextract so that SPM knows
+    // this command must run after jextract finishes.
+    var javaCallbacksArguments = [
+      "java-callbacks-build",
```
ktoso, line 214: Do we HAVE to do it all int one huge mega step? Can we not invoke multiple commands one by one? I'd like to make this hidden in command line somehow...

Sources/SwiftJavaTool/Commands/JavaCallbacksBuildCommand.swift
```diff
@@ -0,0 +1,211 @@
...
+import Foundation
+import Subprocess
+
+#if canImport(System)
+import System
+#else
+@preconcurrency import SystemPackage
+#endif
+
+/// Builds Swift-Java callbacks in a single command:
```
ktoso: Put a huge `WORKAROUND: rdar://172649681 if we invoke commands one by one with java outputs SwiftPM will link Foundation` here

ktoso: Actually, would be good if you filed an issue on swiftpm explaining this behavior, might be better to track that way coming from you directly as well

Sources/SwiftJavaTool/Commands/JavaCallbacksBuildCommand.swift
```diff
@@ -0,0 +1,211 @@
...
+/// 3. Running `swift-java configure` to produce a swift-java.config
+/// 4. Running `swift-java wrap-java` to generate Swift wrappers
+///
+/// This command is used by ``JExtractSwiftPlugin`` to consolidate all of the above
+/// into a single build command that declares only a Swift file as its output,
+/// avoiding SPM treating intermediate Java artifacts (compiled classes, config files,
+/// Gradle output directories) as module resources, which would trigger
+/// resource_bundle_accessor.swift generation and pull Foundation.Bundle into the binary.
+extension SwiftJava {
+  struct JavaCallbacksBuildCommand: AsyncParsableCommand {
```
ktoso, line 39: Man I really dislike this command... we'll have to copy around and keep it updated constantly :-\

Before we merge this, can we make sure we can't avoid it? How about just ignoring the java outputs, does swiftpm prevent things from working if we just pretend they don't exist but we write them anyway?

Sources/SwiftJavaTool/SwiftJava.swift
```diff
@@ -37,6 +37,7 @@ struct SwiftJava: AsyncParsableCommand {
       ResolveCommand.self,
       WrapJavaCommand.self,
       JExtractCommand.self,
+      JavaCallbacksBuildCommand.self,
```
ktoso, line 40: Can you check how we can hide this form `--help`? Alternatively, _ prefix it or do something to its name in argument parser etc.

Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
madsodgaard: Hm, what would we benefit from writing multiple commands? As this is only intended for internal use and a workaround for the SwiftPM limitation.

Sources/SwiftJavaTool/SwiftJava.swift
madsodgaard: That's what the `shouldDisplay: false` does:

```
OVERVIEW: Generate sources and configuration for Swift and Java interoperability.

USAGE: swift-java <subcommand>

OPTIONS:
  -h, --help              Show help information.

SUBCOMMANDS:
  configure               Configure and emit a swift-java.config file based on an input dependency or jar file
  resolve                 Resolve dependencies and write the resulting swift-java.classpath file
  wrap-java               Wrap Java classes with corresponding Swift bindings.
  jextract                Wrap Swift functions and types with Java bindings, making them available to be called from Java
```

Sources/SwiftJavaTool/Commands/JavaCallbacksBuildCommand.swift
madsodgaard: https://github.com/swiftlang/swift-package-manager/issues/9813

Sources/SwiftJavaTool/Commands/JavaCallbacksBuildCommand.swift
madsodgaard: The problem is I am guessing that SwiftPM uses the input/output to derive the dependencies between the build commands. So, I am afraid that it will run the builds in the wrong order. 

Alternatively, if its only using it for stale checks, that means that if we remove them from the outputs, we would retrigger those commands on each invocation of the build plugin. Maybe someone from the SwiftPM team can confirm the behaviour here?

Sources/SwiftJavaTool/SwiftJava.swift
ktoso: Thanks missed that :) At least we're hiding it then, thank you

Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
ktoso: I meant finding some way to keep the previous workflow going rather than introducing this command

Sources/SwiftJavaTool/Commands/JavaCallbacksBuildCommand.swift
ktoso: Thank you

Sources/SwiftJavaTool/Commands/JavaCallbacksBuildCommand.swift
ktoso: @dschaefer2  would you mind confirming we have no other way out here?

PR #626, opened 2026-03-17
wrap-java: Supporting return type erasure bounds

sidepelican:

Calling `getGenericDeclaration()` on a generated `java.lang.reflect.TypeVariable` currently results in a runtime crash:

> JavaLangReflect/TypeVariable.swift:14: Fatal error: Java call threw unhandled exception: java.lang.NoSuchMethodError: Lsun/reflect/generics/reflectiveObjects/TypeVariableImpl;.getGenericDeclaration()Ljava/lang/Object;

This is because the JNI method lookup is searching for a signature returning `Object`, whereas the actual erased signature in Java returns `GenericDeclaration` (the upper bound of the type parameter `D`).

Currently, the `@JavaMethod` macro expands using `JavaObject?.self` as the `resultType` for the erased generic return type.
Since the macro cannot infer the original Java bounds from the Swift method signature alone, it defaults to `Object`.

I have introduced a new macro argument, `typeErasedResultBound`, to explicitly provide the bound type.

With this change, we can specify the correct type for JNI method resolution:

```swift
@JavaMethod(typeErasedResult: "D!", typeErasedResultBound: GenericDeclaration?.self)
public func getGenericDeclaration() -> D!
```

```swift
  let result$ = {
    do {
      return try dynamicJavaMethodCall(methodName: "getGenericDeclaration", resultType: /*type-erased:D*/ GenericDeclaration?.self)
    } catch {
  ...
```

### DIFF
--- a/Sources/SwiftJava/Macros.swift
+++ b/Sources/SwiftJava/Macros.swift
@@ -141,7 +141,7 @@ public macro JavaStaticField(_ javaFieldName: String? = nil, isFinal: Bool = fal
 /// This allows the macro to form a call into the get() method, which at runtime, will have an `java.lang.Object`
 /// returning method signature, and then, convert the result to the expected `T` type on the Swift side.
 ///
-/// If the return type is a wildcard type, specify the bound type instead of `java.lang.Object`:
+/// If the return type is a bounded type parameter, specify the bound type instead of `java.lang.Object`:
 /// ```swift
 /// // Java: class Test<T extends Animal> { public T get(); }
 /// @JavaMethod(typeErasedResult: "T!", typeErasedResultBound: Animal?.self)


Sources/SwiftJavaToolLib/JavaClassTranslator.swift
```diff
@@ -988,9 +988,23 @@ extension JavaClassTranslator {
...
-        parameters.append("typeErasedResult: \"\(resultType)\"")
+        let returnType = javaMethod.getReturnType()!
+        var boundType = try translator.getSwiftTypeNameAsString(
+          method: javaMethod,
+          returnType.as(Type.self),
+          substitution: substitution,
+          preferValueTypes: true,
+          outerOptional: .nonoptional
+        )
+        // `getSwiftTypeNameAsString` does not includes generic parameters for non parameterized type
```
ktoso:
```suggestion
        // `getSwiftTypeNameAsString` does not include generic parameters for non parameterized type
```

Sources/SwiftJava/Macros.swift
```diff
@@ -146,6 +146,13 @@ public macro JavaMethod(
   typeErasedResult: String? = nil
 ) = #externalMacro(module: "SwiftJavaMacros", type: "JavaMethodMacro")
 
+@attached(body)
```
ktoso: Could need some docs why this one does require the erased result types and then do a `See:` to the "main one" for more docs.

Sources/SwiftJava/Macros.swift
```diff
@@ -146,6 +146,13 @@ public macro JavaMethod(
...
+public macro JavaMethod<Result: JavaValue>(
```
ktoso: Actually... how about we keep ONE macro, and diagnose from inside the macro if `typeErasedResultBound` was provided but the `typeErasedResult` wasn't? Since that's what you're trying to enforce here. I'd like to avoid having many overloads here in the sources so that docs are easy to find

sidepelican:
I was able to unify the macro definitions.
I have added an explanation for `typeErasedResultBound`. Let me know what you think.

Sources/SwiftJava/Macros.swift
```diff
@@ -134,16 +134,24 @@ public macro JavaStaticField(_ javaFieldName: String? = nil, isFinal: Bool = fal
...
 /// ```swift
-/// // Java: class Test<T> { public <T> get(); }
+/// // Java: class Test<T> { public T get(); }
 /// @JavaMethod(typeErasedResult: "T!")
 /// func get() -> T!
 /// ```
 /// This allows the macro to form a call into the get() method, which at runtime, will have an `java.lang.Object`
 /// returning method signature, and then, convert the result to the expected `T` type on the Swift side.
+///
+/// If the return type is a wildcard type, specify the bound type instead of `java.lang.Object`:
```
ktoso:
```suggestion
/// If the return type is a bounded type parameter, specify the bound type instead of `java.lang.Object`:
```

A wildcard is this guy `?`

ktoso:
Very nice, LGTM; only nitpickt that the docs were not showing a wildcard, a wildcard is `?`

PR #627, opened 2026-03-18
Regenerate all files using wrap-java

sidepelican:
Completes #599.

This PR regenerates all files within the `generated/` directory using the `wrap-java` command.
To streamline this process, I have also added a utility script: `scripts/wrap-java-generate.sh`.

Any manually implemented `extension` have been moved to separate files.

#### Notable Changes

* `JarEntry`, `JarFile`, `JarInputStream`, `JarOutputStream`, and `ZipEntry`:
Static fields originating from `ZipConstants` are no longer included, as `ZipConstants` is a private interface.
* `JavaCharacter.UnicodeScript`:
Removed because the generator does not yet support enum generation.
* `AbstractMap.SimpleEntry`, `AbstractMap.SimpleImmutableEntry`:
This was previously located in `TreeMap`, but it correctly belongs to `AbstractMap`. Since `AbstractMap` is not currently a generation target and has relatively niche utility, I have omitted it for now.

### DIFF
--- a/Samples/JavaKitSampleApp/Sources/JavaKitExample/JavaKitExample.swift
+++ b/Samples/JavaKitSampleApp/Sources/JavaKitExample/JavaKitExample.swift
@@ -46,7 +46,7 @@ extension HelloSwift: HelloSwiftNativeMethods {
     _ = self.sayHelloBack(42)
 
     let predicate: JavaPredicate<JavaInteger> = self.lessThanTen()!
-    let value = predicate.test(JavaInteger(3).as(JavaObject.self))
+    let value = predicate.test(JavaInteger(3))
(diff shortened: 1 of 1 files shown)


ktoso:
Very exciting! We should follow up with checking in CI that there’s no change unless it’s also committed… I’ll do that later.

some small nits incoming, this is great though!

scripts/wrap-java-generate.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
ktoso, line 1: That's a good script but we'll want to make it a bit more automatic...

I think I'm ok with merging this as is and I'll work on CI infra for validation of these things and update the script then as well 👍

PR #635, opened 2026-03-19
remove sil-verify-none workaround

ktoso:
this removes last use of unsafe flags


ktoso:
Let's give this a go again.

PR #648, opened 2026-03-26
jextract-jni: Add type parameters to generated java types

sidepelican:
The Java classes introduced in #572 were generated without their generic parameters.
This PR adds them.

To map Swift generic parameter values to their corresponding Java type names, I have added `translateGenericTypeParameter` to `JNISwift2JavaGenerator`.
(Adding this function ended up requiring an extra argument for many other functions 🥲)

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
+++ b/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
@@ -23,11 +23,11 @@ public class GenericTypeTest {
     @Test
     void genericTypeValueRoundtrip() {
         try (var arena = SwiftArena.ofConfined()) {
-            var stringId = MySwiftLibrary.makeStringID("Java", arena);
+            MyID<String> stringId = MySwiftLibrary.makeStringID("Java", arena);
             assertEquals("Java", stringId.getDescription());
             assertEquals("Java", MySwiftLibrary.takeStringValue(stringId));
 
-            var intId = MySwiftLibrary.makeIntID(42, arena);
+            MyID<Long> intId = MySwiftLibrary.makeIntID(42, arena);
             assertEquals("42", intId.getDescription());
             assertEquals(42, MySwiftLibrary.takeIntValue(intId));
         }
@@ -36,8 +36,8 @@ public class GenericTypeTest {
     @Test
     void genericTypeProperty() {
         try (var arena = SwiftArena.ofConfined()) {
-            var intId = MySwiftLibrary.makeIntID(42, arena);
-            var entity = MyEntity.init(intId, "name", arena);
+            MyID<Long> intId = MySwiftLibrary.makeIntID(42, arena);
+            MyEntity entity = MyEntity.init(intId, "name", arena);
             assertEquals("42", entity.getId(arena).getDescription());
         }
     }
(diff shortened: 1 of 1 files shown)


Sources/JExtractSwiftLib/JavaTypes/JavaType+SwiftKit.swift
```diff
@@ -44,4 +44,8 @@ extension JavaType {
     .class(package: "org.swift.swiftkit.core.collections", name: "SwiftSet", typeParameters: [E.boxedType])
   }
 
+  /// A container for receiving Swift generic instances.
+  static var _OutSwiftGenericInstance: JavaType {
```
sidepelican, line 48: This was mistakenly ported to jni-core, so I'll remove the definition from that side later.

Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
```diff
@@ -21,29 +21,31 @@
 
 public class GenericTypeTest {
     @Test
-    void returnsGenericType() {
+    void genericTypeValueRoundtrip() {
         try (var arena = SwiftArena.ofConfined()) {
-            MyID stringId = MySwiftLibrary.makeStringID("Java", arena);
+            var stringId = MySwiftLibrary.makeStringID("Java", arena);
```
ktoso: Can we spell those types out in source please maybe? :)

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -234,6 +234,12 @@ extension JNISwift2JavaGenerator {
         )
       }
       printer.println()
+      let genericClause =
+        if decl.swiftNominal.isGeneric {
+          "<\(decl.swiftNominal.genericParameters.map(\.name).joined(separator: ", "))>"
```
ktoso, line 239: 👍

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -404,11 +403,11 @@ public func makeIntID() -> MyID<Int> {
 will be exported as
 
 ```java
-public final class MyID implements JNISwiftInstance {
+public final class MyID<T> implements JNISwiftInstance {
```
ktoso, line 406: Very exciting :)

Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
sidepelican: I accidentally changed it to `var`, but from a testing perspective, that was actually better.

Sources/JExtractSwiftLib/JavaTypes/JavaType+SwiftKit.swift
ktoso: Thanks! I had a feel we'll move one or two too many huh

Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
ktoso: I kinda like if we had the explicit types here, we'd fail to compile i guess if we mess up gen 🤔 

Not a big issue

PR #649, opened 2026-03-27
jextract: [SwiftType] `optional`, `array`, `dictionary`, and `set` into `nominal` type

sidepelican:
## Problem

Currently, `SwiftType` defines `optional`, `array`, `dictionary`, and `set` as distinct cases.
However, these types are sometimes represented nominally (e.g., `Optional<Int>`).

This fragmentation forces us to implement logic for both `.nominal` and the specific cases (like `.optional`), leading to redundant code.
More importantly, it creates a risk of logic desynchronization; in fact, I've encountered issues where logic was updated for one case but overlooked for the other.

## Solution

This PR unifies these special types into the `.nominal` case.

To preserve the original syntactic representation (such as `Int?` or `[Int]`), I have introduced `SwiftNominalType.SugarName`.
This allows us to handle these types as nominal while still maintaining their original "sugared" form in the grammar.

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
@@ -795,7 +795,6 @@ extension FFMSwift2JavaGenerator {
                       )
                     )
                   ),
-                  // .constant("var  = new \(.placeholderForDowncallThunkName).."),
                   .placeholderForDowncall, // perform the downcall here
                 ],
                 extractResult: .property(.constant("_result_initialize"), propertyName: "result")


Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownTypeDecls.swift
```diff
@@ -14,6 +14,142 @@
 
 import SwiftSyntax
 
+enum SwiftKnownType: Equatable {
```
ktoso, line 17: This enum is a bit weird, we're 1:1 mapping things almost hmmm

ktoso:
Sorry I'l finish reviewing tomorrow

Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownTypeDecls.swift
sidepelican: These two types are essentially similar, the only difference is whether they represent the decl site or the use site.
I haven't quite found a better way to reduce duplication.

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -774,6 +751,57 @@ extension FFMSwift2JavaGenerator {
...
+                          // We need to refer to the nested class that is created for this function.
+                          // The class that contains all the related functional interfaces is called the same
+                          // as the downcall function, so we use the thunk name to find this class/
+                          .placeholderForSwiftThunkName, .constant("$_result_initialize.Function$Impl()"),
+                        ],
+                        separator: "."
+                      )
+                    )
+                  ),
+                  // .constant("var  = new \(.placeholderForDowncallThunkName).."),
```
ktoso: stray comment?
```suggestion
```

Sources/JExtractSwiftLib/Common/TypeAnnotations.swift
```diff
@@ -18,10 +18,12 @@ import SwiftJavaJNICore
...
 /// in Java sources when the corresponding Java type is rendered.
 func getTypeAnnotations(swiftType: SwiftType, config: Configuration) -> [JavaAnnotation] {
-  switch swiftType {
-  case .array(let wrapped) where wrapped.isUnsignedInteger:
+  if swiftType.isUnsignedInteger {
     return [JavaAnnotation.unsigned]
-  case _ where swiftType.isUnsignedInteger:
+  }
+
+  switch swiftType.asNominalType?.asKnownType {
```
ktoso, line 25: OKey yeah I think this is a nice improvement

Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownTypeDecls.swift
ktoso: Yeah, let's run with it thanks

Issue #337, opened 2025-07-31
[JExtract] Translate String to a imported type


rintaro:
`Swift.String` should be translated to an imported `String` type instead of `java.lang.String`, just like `Foundation.Data`.

In JExtract/FFM, `String` is currently translated to `java.lang.String` and lowered to C-string (`UnsafePointer<Int8>`).
That works, but is making passing `String` value between Swift and Java more expensive than it should be.

For example, using this interface,
```swift
public func getSomeString() -> String
public func receiveString(str: String)
```
In Java:
```java
var str = MyLib.getSomeString()
receiveString(str)
```
This receiving and passing string value should not cause copy.

rintaro:
Like `Data` hardcoded API is available here.
https://github.com/swiftlang/swift-java/blob/7f24d6276f59c34544a4c8f6771a8bc560aa5db9/Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownModules.swift#L84-L87

ktoso:
Very good idea, maybe we'll end up with a `SwiftString extends CharSequence` for this? 🤔  https://docs.oracle.com/javase/jp/8/docs/api/java/lang/CharSequence.html and a toString on it would cause a copy if we needed it.

PR #651, opened 2026-03-28
jextract/ffm: Support returning Strings

ktoso:
For now we do this by copying twice... copy swift string into c string, return pointer, then we copy that into Java string and free the C string.

We can do better in the future and also consider https://github.com/swiftlang/swift-java/issues/337 [JExtract] Translate String to a imported type #337

but for now I just wanted to get an API with Strings working in FFM


ktoso:
Just fixed the formatting

PR #652, opened 2026-03-28
[Android] Add support for emitting symbol list for better DCE, also add docs about binary sizes

madsodgaard:
Does a few things that allows us to accomplish better DCE for Android binaries
* The JNI mode of JExtract will emit a `.map` file which can be passed to the linker as `--version-script` meaning that it can use those as main entries for DCE.
* Annotate all `@_cdecls` with the new `@used` in 6.3, so they do not get DCE when using flags like `-internalize-at-link`
* Add docs on how developers can achieve smaller binary sizes.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
@@ -140,7 +140,8 @@ extension JNISwift2JavaGenerator {
       return
     }
 
-    let symbolLines = generatedCDeclSymbolNames
+    let symbolLines =
+      generatedCDeclSymbolNames
       .sorted()
       .map { "  \($0);" }
       .joined(separator: "\n")
--- a/Package.swift
+++ b/Package.swift
@@ -134,7 +134,7 @@ let package = Package(
       dependencies: [
         "SwiftJava",
         "SwiftJavaRuntimeSupport",
-        "SwiftRuntimeFunctions"
+        "SwiftRuntimeFunctions",
       ]
     ),
 
@@ -400,7 +400,7 @@ let package = Package(
     .testTarget(
       name: "SwiftRuntimeFunctionsTests",
       dependencies: [
-        "SwiftRuntimeFunctions",
+        "SwiftRuntimeFunctions"
       ],
       swiftSettings: [
         .swiftLanguageMode(.v5)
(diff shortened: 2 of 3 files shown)


Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -117,6 +117,17 @@ struct JExtractSwiftBuildToolPlugin: SwiftJavaPluginProtocol, BuildToolPlugin {
 
     let jextractOutputFiles = outputSwiftFiles
 
+    // In JNI mode, emit a linker version script so the linker can DCE unused Swift code.
```
ktoso, line 120: Do we have to restrict it to JNI mode? I think we have enough thunks in FFM as well that they should keep alive anything used hm

We can do it separately, can you file an issue please?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -119,6 +119,41 @@ extension JNISwift2JavaGenerator {
...
+    }
+    guard !generatedCDeclSymbolNames.isEmpty else {
+      return
+    }
+
+    let symbolLines = generatedCDeclSymbolNames
+      .sorted()
+      .map { "  \($0);" }
+      .joined(separator: "\n")
+    let contents = "{\nglobal:\n\(symbolLines)\nlocal: *;\n};\n"
```
ktoso: Would you mind doing this as multiline string rather than manually \n please?

Sources/SwiftJavaDocumentation/Documentation.docc/Android.md
```diff
@@ -41,4 +41,80 @@ open class OldVersionedClass: JavaObject {
...
+
+```bash
+swift build \
+  --swift-sdk aarch64-unknown-linux-android28 \
+  -c release \
+  --experimental-lto-mode=full \
+  -Xlinker --version-script=.build/plugins/outputs/MyLibrary/MySwiftLibrary/JExtractSwiftPlugin/swift-java-jni-exports.map
+```
+
+> Tip: Adjust the `--version-script` path to match your package name and target name. Run `find .build/plugins/outputs -name swift-java-jni-exports.map` after the first build if you are unsure of the exact path.
```
ktoso, line 120: Thanks for the docs!

Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
madsodgaard: #653

madsodgaard:
@ktoso Agreed with tests. Just added one explicit test instead.

PR #654, opened 2026-03-30
Lower minimum macOS version to v13

llsc12:
My swift package had a minimum deployment version of macOS v13, which conflicted with swift-java's v15.

However, swift-java has no issue when this is changed to v13, likely just unchanged from package initialisation!

Thank you.


ktoso:
Hmm, yeah I don't see why we can't lower it -- and we cna do so now while we're in 0.x timeline so that's good, thanks for the patch

PR #657, opened 2026-03-31
jextract/jni: Support generic types within Optional and Tuple

sidepelican:
Currently, code generation is incorrect when generic types are used in combination with other complex types.
This PR fixes this for Optional and Tuple types.

## Key Changes

Previously, when a native function used indirect, the downcall was printed in `JavaNativeConversionStep`.
This design required calling `.print(.placeholder)` during the initial Step of the conversion flow.

However, detecting whether a step is the "first" one is difficult during flow construction. To simplify the logic, I have moved the downcall printing to `JNISwift2JavaGenerator`.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/GenericType.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/GenericType.swift
@@ -59,11 +59,11 @@ public enum MyIDs {
     value.rawValue
   }
 
-  public static func makeOptionalIntID(_ value: Int?) -> MyID<Int?> {
+  public static func makeOptionalStringID(_ value: String?) -> MyID<String?> {
     MyID(value)
   }
 
-  public static func takeOptionalIntValue(from id: MyID<Int?>) -> Int? {
+  public static func takeOptionalStringValue(from id: MyID<String?>) -> String? {
     id.rawValue
   }
 }
--- a/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
+++ b/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/GenericTypeTest.java
@@ -40,17 +40,14 @@ public class GenericTypeTest {
             assertEquals("Java", MyIDs.takeValuesFromTuple(ids).$0);
             assertEquals(42, MyIDs.takeValuesFromTuple(ids).$1);
 
-            // MyID<Boolean>[] boolIds = MyIDs.makeBoolIDArray(true, 3, arena);
-            // assertEquals(List.of(true, true, true), MyIDs.takeBoolValuesFromArray(boolIds));
-            
             Optional<MyID<Double>> doubleIdOptional = MyIDs.makeDoubleIDOptional(42.195, arena);
             assertTrue(doubleIdOptional.isPresent());
             assertEquals(42.195, MyIDs.takeDoubleValueOptional(doubleIdOptional).getAsDouble());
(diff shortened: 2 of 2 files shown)


Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/GenericType.swift
```diff
@@ -22,20 +22,58 @@ public struct MyID<T> {
...
 
-public func takeStringValue(from value: MyID<String>) -> String {
-  value.rawValue
+  // public static func makeBoolIDArray(_ value: Bool, length: Int) -> [MyID<Bool>] {
+  //   Array(repeating: MyID(value), count: length)
+  // }
+
+  // public static func takeBoolValues(from ids: [MyID<Bool>]) -> [Bool] {
+  //   ids.map { $0.rawValue }
+  // }
```
ktoso: delete or enable these please

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
```diff
@@ -164,18 +164,31 @@ extension JavaType {
...
+    case .char: return "Character"
+    case .short: return "Short"
+    case .int: return "Integer"
+    case .long: return "Long"
+    case .float: return "Float"
+    case .double: return "Double"
+    case .void: return "Void"
+    case .javaLangString: return "String"
+    case .class(let package, let name, let typeParameters):
+      let packageClause: String =
```
ktoso: It's not really a "clause" it's a Part maybe
```suggestion
      let packagePart: String =
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1490,40 +1482,51 @@ extension JNISwift2JavaGenerator {
...
         let inner = inner.render(&printer, placeholder)
-        printer.print(
-          """
-          environment.interface.SetLongField(environment, \(outArgumentName), _JNIMethodIDCache._OutSwiftGenericInstance.selfPointer, \(inner))
-          let metadataPointer = unsafeBitCast(\(swiftFunctionResultType).self, to: UnsafeRawPointer.self)
-          let metadataPointerBits$ = Int64(Int(bitPattern: metadataPointer))
-          environment.interface.SetLongField(environment, \(outArgumentName), _JNIMethodIDCache._OutSwiftGenericInstance.selfTypePointer, metadataPointerBits$.getJNIValue(in: environment))
-          """
-        )
+        printer.printBraceBlock("do") { printer in
```
ktoso, line 1520: why do we need the `do {}`?

sidepelican: Because the variable name `metadataPointer` can cause conflicts in tuples.

ktoso:
Ok happy enough with it, once tests pass

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
ktoso: Okey gotcha thx

sidepelican:
It looks like there’s an issue with 32-bit Android.
It might take some effort to resolve, and since I don't have any more time today, so I’ll work on it tomorrow.

ktoso:
Thank you!

sidepelican:
I've confirmed that this is an existing issue related to `Int?` and not related to this PR.
Fixing it is outside the scope of this PR, so I'll update the sample code to use another pattern instead.

ktoso:
Sounds good, thank you!

Issue #34, opened 2024-10-01
jextract-swift support importing `throws` functions


ktoso:
Functions in swift that are throwing have a special calling convention so we'll need the help of thunks on the Swift side for this.

Once we have those, the imported methods should mirror the throwing behavior and be declared checked throwing in Java (?).

Typed throws would be step 2 of this.

madsodgaard:
We should rename this to only FFM, JNI mode supports throwing functions.

PR #658, opened 2026-03-31
jextract/ffm: Support throwing funcs in FFM mode

ktoso:
resolves long standing missing feature https://github.com/swiftlang/swift-java/issues/34

I introduce a new Java error type that we use to "type erase" the thrown errors when Java gets them. In future we can do typed throws.

We also jextract and commit the wrapper for it so we don't have bootstrap issues, i.e. we can build without having to run jextract first: `org.swift.swiftkit.ffm.generated.SwiftJavaErrorException` 

This is similar to the issue that wrap-java has with comitted generated wrappers in SwiftJava.

Added runtime and source gen tests

### DIFF
--- a/Sources/SwiftRuntimeFunctions/SwiftJavaError.swift
+++ b/Sources/SwiftRuntimeFunctions/SwiftJavaError.swift
@@ -31,4 +31,3 @@ public final class SwiftJavaError {
     unsafeBitCast(type(of: underlying), to: UnsafeRawPointer.self)
   }
 }
-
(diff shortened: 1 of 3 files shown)


Sources/SwiftRuntimeFunctions/SwiftJavaError.swift
```diff
@@ -0,0 +1,33 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Wraps a Swift `any Error` value so it can be passed across the FFI boundary
+/// as a reference-counted heap object
+public final class SwiftJavaError {
```
ktoso, line 17: This is to avoid guessing if the struct we're throwing needs to be retained or now, we just wrap them all into this

Issue #338, opened 2025-07-31
[JExtract] Support `typealias`


rintaro:
```swift
public struct Bytes {
  public typealias Element = Int8
  public func append(_ element: Element)
}
```
`Bytes.append(_:)` should be translated to `void append(java.lang.Byte element)`

To do that JExtract should support `typealias` declaration.

rintaro:
This should also fix this:
https://github.com/swiftlang/swift-java/blob/7f24d6276f59c34544a4c8f6771a8bc560aa5db9/Sources/JExtractSwiftLib/SwiftTypes/SwiftKnownModules.swift#L81-L82

timofeys1:
I am curious if this is the limitation of jextract relying heavily on Swift Syntax? Does this mean Swift Java or Swift Syntax itself would have to reimplement some form of type checking and inference? I imagine Swift Java can't just naively translate a typealias because Java doesn't have a notion of a typealias.

rintaro:
We do have a basic name-lookup implementation already 
https://github.com/swiftlang/swift-java/blob/8cf1829d8d57f66b15aea66c3a40f9db74aa57a0/Sources/JExtractSwiftLib/SwiftTypes/SwiftTypeLookupContext.swift#L31-L37
It's far from complete, but  it should be sufficient for initial `typealias` supporting work.

For anything that isn't handled by the current lookup logic, we can file separate issues individually.

tkrajacic:
I was just going to create a new issue, but I guess this is just as good a place.

Top level typealiases are pretty useful when dealing with properties that are not (yet) bridged to Java:
```swift
#if os(Android)
public typealias Amount = Double
#else
public typealias Amount = Decimal
#endif
```

See https://github.com/tkrajacic/swift-android-examples/commit/b42c1784febf3789d55a6cd4348bb4aa8a512fae

This doesn't work yet, but would be pretty convenient.

PR #664, opened 2026-04-01
jextract/jni: Specialization and constrained extension support

ktoso:
This allows specializing types e.g. Box<T> with a typealias `FishBox = Box<Fish>` becomes its own type in Java and gets all methods from Box but also any methods where the `extension Box where Element == Fish {}`.

This gives Java access to previously unavailable APIs in a safe way. Though yes it makes the type different in Java - so one may have to convert to a Box<> sometimes etc. We may need more helper methods for those.

For now I want to enable getting to those otherwise unavailable methods via specialization.

Specialization is available via a typealias trigger, or via configuration which is nice if you don't control the sources you're trying to specialize.

This partially addresses typealias support as well https://github.com/swiftlang/swift-java/issues/338

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/BoxSpecialization.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/BoxSpecialization.swift
@@ -1,0 +1,37 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024-2026 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct Box<Element> {
+  public var count: Int64
+
+  public init(count: Int64) {
+    self.count = count
+  }
+}
+
+public struct Fish {
+  public var name: String
+
+  public init(name: String) {
+    self.name = name
+  }
+}
+
+extension Box where Element == Fish {
+  public func describeFish() -> String {
+    "A box of \(count) fish"
+  }
+}
+
+public typealias FishBox = Box<Fish>
--- a/Sources/JExtractSwiftLib/ImportedDecls.swift
+++ b/Sources/JExtractSwiftLib/ImportedDecls.swift
@@ -29,8 +29,17 @@ package enum SwiftAPIKind: Equatable {
 
 /// Describes a Swift nominal type (e.g., a class, struct, enum) that has been
 /// imported and is being translated into Java.
+///
+/// When `base` is non-nil, this is a specialization of a generic type
+/// (e.g. `FishBox` specializing `Box<Element>` with `Element` = `Fish`).
+/// The specialization delegates its member collections to the base type
+/// so that extensions discovered later are visible through all specializations.
 package final class ImportedNominalType: ImportedDecl {
   let swiftNominal: SwiftNominalTypeDeclaration
+
+  /// If this type is a specialization (FishTank), then this points at the Tank base type of the specialization.
+  /// His allows simplified
+  package let specializationBaseType: ImportedNominalType?
 
   // The short path from module root to the file in which this nominal was originally declared.
   // E.g. for `Sources/Example/My/Types.swift` it would be `My/Types.swift`.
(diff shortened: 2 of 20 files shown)


Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/BoxSpecialization.swift
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+extension Box where Element == Fish {
+  public func describeFish() -> String {
+    "A box of \(count) fish"
+  }
+}
+
+public typealias FishBox = Box<Fish>
```
ktoso, line 37: This triggers specialization for a new FishBox "type" on the java side 🐟

Sources/SwiftJava/Macros.swift
```diff
@@ -192,3 +192,17 @@ public macro JavaStaticMethod(_ javaMethodName: String? = nil) =
...
+/// ```swift
+/// @JavaExport public typealias FishTank = Tank<Fish>
+/// ```
+/// This tells jextract to generate a concrete `FishTank` Java class for `Tank<Fish>`.
+///
+/// When applied to a struct, class, or enum, it force-includes the type for export
+/// even if it would otherwise be excluded by `swiftFilterExclude` configuration.
+@attached(peer)
+public macro JavaExport() =
+  #externalMacro(module: "SwiftJavaMacros", type: "JavaExportMacro")
```
ktoso, line 206: Did not end up using in this PR, but this is a way to exclude "everything" and then include a specific specialization

Sources/JExtractSwiftLib/ImportedDecls.swift
```diff
@@ -38,29 +38,204 @@ package final class ImportedNominalType: ImportedDecl {
...
+}
+
+struct SpecializationError: Error {
+  let message: String
+}
+
+/// A specialized view of an imported generic type for a concrete type argument substitution.
+/// E.g. `SpecializedImportedNominalType` wrapping `Tank<T>` with specializedTypeName="FishTank"
+/// and genericArguments=["Element": "Fish"].
+package final class SpecializedImportedNominalType: ImportedNominalType {
```
ktoso: Yeah I have mixed feelings about the subclassing here... Let me actually rework this maybe

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
```diff
@@ -63,6 +63,7 @@ SwiftJava's `swift-java jextract` tool automates generating Java bindings from S
 | Dictionaries: `[String: Int]`, `[K:V]`                                               | ❌        | ✅   |
 | Generic type: `struct S<T>`                                                          | ❌        | ✅   |
 | Functions or properties using generic type param: `struct S<T> { func f(_: T) {} }`  | ❌        | ❌   |
+| Generic type specialization and conditional extensions: `struct S<T>{} extension S where T == Value {}`  | ✅        | ❌   |
```
sidepelican, line 66: I think these are swapped.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -1185,11 +1195,44 @@ extension JNISwift2JavaGenerator {
       )
     }
 
+    /// Translate a single element type for tuple results on the Java side.
+    private func translateTupleElementResult(
```
sidepelican, line 1199: I deleted this function in my previous https://github.com/swiftlang/swift-java/pull/657 . It seems like it might have slipped back in during a merge?

ktoso: Ugh seems so, I had a few long running branches -- sorry, let me remove that

Sources/SwiftJavaDocumentation/Documentation.docc/SupportedFeatures.md
ktoso, line 66: Huh indeed, thank you.

I wonder if inverse order would be preferable tbh... seems most people use jni mode nowadays 🤔

sidepelican: I think FFM has a better design, but since I need Android support, I have to stick with the older approach🤣

ktoso: Yeah, that's the life we're stuck with on Android 😅

ktoso:
Thanks for review @sidepelican , added cleanups to https://github.com/swiftlang/swift-java/pull/666

PR #668, opened 2026-04-01
jextract/jni: handle UnsafeRawBufferPointer params

ktoso:
This allows passing a Java byte array to an `UnsafeRawBufferPointer` accepting API in Swift. This conversion needs to copy a bunch but makes pretty much sense. This is in JNI mode specifically.

In FFM mode we could do better by allocating and passing a MemorySegment but this is to unblock APIs in JNI mode specifically.

~~We do allow returning an URBP to Java as well, but this is very dubious and scary, so we don't make anything sneaky here and literarily return this as an `SwiftUnsafeRawBufferPointer` that you may be able to do "something" with in java...~~ I actually ended up deciding that's a horrible idea, if anyone really needs it they can talk to us and we'll figure it out, otherwise I'm skipping this for now.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1164,7 +1164,7 @@ extension JNISwift2JavaGenerator {
     /// `SwiftType(inner)`
     indirect case labelessInitializer(NativeSwiftConversionStep, swiftType: SwiftType)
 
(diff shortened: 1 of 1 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1154,6 +1164,9 @@ extension JNISwift2JavaGenerator {
     /// `SwiftType(inner)`
     indirect case labelessInitializer(NativeSwiftConversionStep, swiftType: SwiftType)
 
+    /// Converts a jbyteArray to UnsafeRawBufferPointer via GetByteArrayElements + defer
```
```suggestion
    /// Converts a jbyteArray to UnsafeRawBufferPointer via GetByteArrayElements
```

PR #669, opened 2026-04-01
jextract/jni: support generic params : DataProtocol 

ktoso:
Functions with `<D: DataProtocol>` generic constraints failed to translate because `.foundationDataProtocol` and `.essentialsDataProtocol` fell through to the default case in `translateGenericTypeParameter`.

This uses the automatic existential opening in the thunks!

The test thunks outputs are a bit huge but I found it useful to have them visually tbh... Worst case if much changes we can change them to be less checking

### DIFF
--- a/Sources/JExtractSwiftLib/Convenience/SwiftSyntax+Extensions.swift
+++ b/Sources/JExtractSwiftLib/Convenience/SwiftSyntax+Extensions.swift
@@ -131,7 +131,7 @@ extension AttributeListSyntax.Element {
   /// Whether this node has `SwiftJava` wrapping attributes (types that wrap Java classes).
   /// These are skipped during jextract because they represent Java->Swift wrappers.
   /// Note: `@JavaExport` is NOT included here — it forces export of Swift types to Java.
-  var isJava: Bool {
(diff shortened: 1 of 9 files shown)


Tests/JExtractSwiftTests/DataImportTests.swift
```diff
@@ -578,4 +578,161 @@ final class DataImportTests {
...
+          #else
+          func secondpointer$DoLoad<Ty>(_ ty: Ty.Type) -> any (DataProtocol) {
+            secondpointer$RawPointer$.load(as: ty) as! any (DataProtocol)
+          }
+          let secondpointer$Existential$ = _openExistential(secondpointer$DynamicType$, do: secondpointer$DoLoad)
+          #endif
+          secondswiftObject$ = secondpointer$Existential$
+          return SwiftModule.verify(first: firstswiftObject$, second: secondswiftObject$).getJNILocalRefValue(in: environment)
+        }
+        """
```
ktoso, line 673: On the other hand... man that's a lot of code for the chunk... I'll just take the imp[important bits maybe

PR #670, opened 2026-04-02
jextract: LabeledTuple support!

ktoso:
We now support passing and returing labelled tuples like (a: Int, b: Stirng) etc.

```java
        var result = MySwiftLibrary.labeledTuple();
        // Access via named accessors
        assertEquals(10, result.x());
        assertEquals(20, result.y());
        // Positional access still works (inherited from Tuple2)
        assertEquals(10, result.$0);
        assertEquals(20, result.$1);

        // The labelled tuple is a subclass of Tuple2
        assertInstanceOf(Tuple2.class, result);
        // And the generic types match positionally as well
        @SuppressWarnings("unused")
        Tuple2<Integer, Integer> check = result;
```

This is important for being able to quickly port over code from swift to Java using the same libraries as we don't have to special case the labelled tuples and move them over to positional use.

These ad hoc tuple types are printed per method.

They would conflict if you use the same shape twice in a method right now... I'm not seeing much of that use so ignored it for now.

The new tuples inherit from TupleN so Java code can just use them as positional when necessary, also for passing them along to other methods.

The change is large because I also cleaned up type printing, I think printing full qualified type names is good here, better nor risk clashes.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+LabeledTuples.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+LabeledTuples.swift
@@ -26,7 +26,7 @@ extension JNISwift2JavaGenerator {
   /// public static class LabeledTuple_fn_x_y<T0, T1>
   ///     extends org.swift.swiftkit.core.tuple.Tuple2<T0, T1> {
   ///
-  ///   public LabeledTuple_fn_x_y(T0 $0, T1 $1) { super($0, $1); }
+  ///   public LabeledTuple_fn_x_y(T0 param0, T1 param1) { super(param0, param1); }
   ///   public T0 x() { return $0; }
   ///   public T1 y() { return $1; }
   /// }
@@ -61,9 +61,12 @@ extension JNISwift2JavaGenerator {
     let typeParamsClause = "<\(typeParams.joined(separator: ", "))>"
     let baseTupleClass = "org.swift.swiftkit.core.tuple.Tuple\(arity)"
 
-    // Constructor parameters: T0 $0, T1 $1, ...
-    let ctorParams = typeParams.enumerated().map { "\($1) $\($0)" }.joined(separator: ", ")
-    let superArgs = (0..<arity).map { "$\($0)" }.joined(separator: ", ")
+    // Constructor parameters: T0 param0, T1 param1, ...
+    // Use paramN names (not $0, $1) because `$N` is invalid as a Swift parameter name,
+    // and the wrap-java generator copies parameter names verbatim into Swift wrappers
+    let paramNames = (0..<arity).map { "param\($0)" }
+    let ctorParams = zip(typeParams, paramNames).map { "\($0) \($1)" }.joined(separator: ", ")
+    let superArgs = paramNames.joined(separator: ", ")
 
     printer.printBraceBlock("public static final class \(rawClassName)\(typeParamsClause) extends \(baseTupleClass)\(typeParamsClause)") { printer in
       // Constructor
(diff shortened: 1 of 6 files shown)


Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -388,10 +388,10 @@ extension FFMSwift2JavaGenerator {
     // If a Swift function is 'throws' we throw a checked error for the Java side
     // TODO: When we support typed throws on Swift side we'll want to throw the right type here instead
     if translatedSignature.isThrowing {
-      throwsClauses.append(JavaType.swiftJavaErrorException.simpleClassName)
+      throwsClauses.append(JavaType.swiftJavaErrorException.className!)
```
ktoso, line 391: simpleClassName was crashing on misuse anyway

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -689,7 +689,8 @@ extension FFMSwift2JavaGenerator {
     /// Translate a Swift API result to the user-facing Java API result.
     func translateResult(
       swiftResult: SwiftResult,
-      loweredResult: LoweredResult
+      loweredResult: LoweredResult,
+      methodName: String
```
ktoso, line 693: we need the method name to form the adhoc types per method

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -1285,6 +1321,7 @@ extension JNISwift2JavaGenerator {
 
         let wrappedValueResult = try translate(
           swiftResult: SwiftResult(convention: .direct, type: swiftType),
+          methodName: "",
```
ktoso, line 1327: Double check this one

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+LabeledTuples.swift
```diff
@@ -0,0 +1,80 @@
...
+    guard case .class(_, let rawClassName, let genericArgs) = labeledTupleType else {
+      return
+    }
+
+    let arity = genericArgs.count
+    let elementNames: [String]
+
+    // Element names are embedded in the class name after "LabeledTuple_<baseName>_"
+    // We need to extract the last `arity` underscore-separated components
+    let parts = rawClassName.split(separator: "_")
```
ktoso, line 50: TODO: this is pretty horrible, we're getting the field names from the type name...

To make this nicer we need to intreoduce SwiftJavaType that wraps a JavaType since JavaType is just "pure" java types without additional context like this. That's quite a big refactor so we may want to do this but better to do separately...

Sources/JExtractSwiftLib/JNI/TranslatedFunctionDecl+NamedTuples.swift
```diff
@@ -0,0 +1,44 @@
...
+    if type.isSwiftJavaLabeledTuple {
+      result.append(type)
+    }
+    for ty in typeParameters {
+      collectLabeledTuples(from: ty, into: &result)
+    }
+  case .array(let element):
+    collectLabeledTuples(from: element, into: &result)
+  default:
+    break
```
ktoso, line 42: potentially more uses here...

Issue #633, opened 2026-03-19
jextract: Evaluate `IfConfigDecl` for simple patterns


sidepelican:
When generating `.swiftinterface` files using the `-emit-module-interface` flag, unexpected `#if` blocks are often generated. For example:

- Original `.swift`

```swift
public struct AppStore: Sendable, Codable {
  public var storeURL: URL?
  public init(storeURL: URL?) {
    self.storeURL = storeURL
  }
}
```

- Generated `.swiftinterface`

```swift
public struct AppStore : Swift.Sendable, Swift.Codable {
  public var storeURL: Foundation.URL?
  #if compiler(>=5.3) && $NonescapableTypes
  public init(storeURL: Foundation.URL?)
  #endif
  public func encode(to encoder: any Swift.Encoder) throws
  public init(from decoder: any Swift.Decoder) throws
}
```

### The Problem

The exact reason for this behavior is unclear, but expressions like `#if compiler(>=5.3) && $NonescapableTypes` appear in various places within generated interfaces.

Currently, `Swift2JavaVisitor` ignores `IfConfigDecl` entirely.
As a result, any method or property signatures wrapped in these `#if` blocks are lost.

### Proposed Solution

We could potentially use [SwiftIfConfig](https://github.com/swiftlang/swift-syntax/blob/491b82f2c03b6aafbfa4e6d80ebdf4d346a3217b/Sources/SwiftIfConfig/SwiftIfConfig.docc/SwiftIfConfig.md) to evaluate these blocks.

Although providing a full [BuildConfiguration](https://github.com/swiftlang/swift-syntax/blob/33e65f8d5bc8823a1402ef76332970822824b36a/Sources/SwiftIfConfig/BuildConfiguration.swift) might be complex, it would be highly beneficial to at least support simple, common patterns.

ktoso:
IfConfig is probably the way, wdyt @DougGregor since you worked on that could you chime in if it's appropriate to extend parsing here to handle code in those blocks?

DougGregor:
SwiftIfConfig is good for handling `#if`s, but it's not easy to know what build configuration to use with it. Recent compilers (Swift 6.3+) have an option to spit out a static build configuration for a particular target triple, but I don't think jextract really knows the target triple. I suppose you could bake a stripped-down version of the static build configuration into jextract and use that consistently. `#if os(...)` wouldn't be evaluated, but `hasFeature` could work with a fixed set of features.

ktoso:
[...]

PR #671, opened 2026-04-02
jextract: Evaluate IfConfigDecl and add --static-build-config option

sidepelican:
- #633

It seems difficult for the `jextract` plugin to reliably execute `-print-static-build-config`.
Instead, this PR embeds a fixed `JExtractDefaultBuildConfiguration`.

The policy for this fixed configuration is to resolve the widest possible range of `#if` conditions. 

This approach introduces a risk where `jextract` might generate code for functions that are not actually present in the loaded Swift library, which would result in a runtime error. 
However, I believe providing access to these functions by default is more useful than having them be completely unavailable.

To support cases where a specific configuration is required, I have added a `--static-build-config <filepath>` option to `jextract`.
This allows users to provide their own generated JSON file to customize the `BuildConfiguration`.


Sources/JExtractSwiftLib/JExtractDefaultBuildConfiguration.swift
```diff
@@ -0,0 +1,101 @@
...
+private let shared = JExtractDefaultBuildConfiguration()
+
+extension BuildConfiguration where Self == JExtractDefaultBuildConfiguration {
+  static var jextractDefault: JExtractDefaultBuildConfiguration {
+    shared
+  }
+}
+
+// $ swift frontend -print-static-build-config -target aarch64-unknown-linux-gnu
+private let printStaticBuildConfigOutput = Data(
```
sidepelican: Based on my investigation, since the `attributes` and `features` are identical across all targets, I have embedded the results from a single representative environment.

Sources/JExtractSwiftLib/Swift2Java.swift
```diff
@@ -34,7 +34,6 @@ public struct SwiftToJava {
     }
 
     let translator = Swift2JavaTranslator(config: config)
-    translator.log.logLevel = config.logLevel ?? .info
```
sidepelican, line 37: It was overwriting with the same value, so the operation was redundant.

Tests/JExtractSwiftTests/Asserts/TextAssertions.swift
```diff
@@ -87,9 +87,17 @@ func assertOutput(
 
   let sourceLocation = SourceLocation(fileID: fileID, filePath: filePath, line: line, column: column)
   for notExpectedChunk in notExpectedChunks {
+    let outputNotContainsNotExpectedChunk = !output.contains(notExpectedChunk)
```
sidepelican, line 90: Same approach with https://github.com/swiftlang/swift-java/pull/552

Plugins/JExtractSwiftPlugin/JExtractSwiftPlugin.swift
```diff
@@ -71,6 +71,13 @@ struct JExtractSwiftBuildToolPlugin: SwiftJavaPluginProtocol, BuildToolPlugin {
       //       We'll have to make up some caching inside the tool so we don't re-parse files which have not changed etc.
     ]
 
+    if let staticBuildConfig = configuration?.staticBuildConfigurationFile {
+      guard let resolvedURL = URL(string: staticBuildConfig, relativeTo: configFile) else {
+        fatalError("Could not resolve 'staticBuildConfigurationFile' url")
```
ktoso:
```suggestion
        fatalError("Could not resolve 'staticBuildConfigurationFile' url: \(staticBuildConfig)")
```

Sources/JExtractSwiftLib/JExtractDefaultBuildConfiguration.swift
```diff
@@ -0,0 +1,101 @@
...
+  var languageVersion: VersionTuple {
+    base.languageVersion
+  }
+
+  var compilerVersion: VersionTuple {
+    base.compilerVersion
+  }
+}
+
+private let shared = JExtractDefaultBuildConfiguration()
```
ktoso: I'd rather have it as a static on the `JExtractDefaultBuildConfiguration` type

Sources/JExtractSwiftLib/JExtractDefaultBuildConfiguration.swift
ktoso: I think we do have to shell out to compute this... otherwise we'll have trouble with new features; I don't want to manually maintian this list here

Tests/JExtractSwiftTests/IfConfigTests.swift
```diff
@@ -0,0 +1,171 @@
...
+        }
+        #endif
+        """,
+      .ffm,
+      .java,
+      detectChunkByInitialLines: 1,
+      expectedChunks: [
+        "public static void androidFunc()",
+        "public final class CannotImport",
+        "public final class IsNotDebug",
```
ktoso, line 67: nice, that's what I would expect here

ktoso (commented):
Looks great, minor concerns about the big features json... I'd rather we don't have to commit that. We may have to NOT have a .shared at all and just compute this during generation instead

sidepelican:
@ktoso  The new BuildToolsPlugin now allows running `-print-static-build-config` during builds.

It uses a global `swift` command so the results may not always match the exact build-time swift version.

sidepelican:
Is the Android CI having some trouble at the moment?
Also, since 6.3 has been officially released, I thought it would be a good idea to add a CI job pinned specifically to 6.3.

ktoso:
Yeah we should do a 6.3 job, I'll follow up on it

madsodgaard:
Great addition! Could we perhaps add some docs about this feature?

ktoso:
Docs would be good about this, it's worth understanding how this works.

I'm offline this evening already, can merge in the morning

ktoso:
Nightlies have some issues with this, I'll have a look later as well.

I'll also add a 6.3 CI

ktoso:
Nightly is having issues, this is not this PR fault

PR #673, opened 2026-04-06
Narrow down `java.library.path` to prevent loading stale dylibs

sidepelican:
`java.library.path` is currently too broad, causing the sample app to load a stale `libSwiftJava.dylib` from the repository root's `.build` directory instead of the latest local build.
This made testing changes extremely confusing and difficult to debug.

Narrowed the search path to prevent unintended binary shadowing and ensure the app consistently loads the correct, newly built library.

### DIFF
--- a/Samples/SwiftJavaExtractFFMSampleApp/build.gradle.kts
+++ b/Samples/SwiftJavaExtractFFMSampleApp/build.gradle.kts
@@ -80,7 +80,7 @@ application {
     applicationDefaultJvmArgs = listOf(
         "--enable-native-access=ALL-UNNAMED",
         // Include the library paths where our dylibs are that we want to load and call
-        "-Djava.library.path=" + javaLibraryPaths(project.projectDir).joinToString(":"),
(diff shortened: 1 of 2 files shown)


Samples/SwiftJavaExtractFFMSampleApp/build.gradle.kts
```diff
@@ -80,7 +80,7 @@ application {
     applicationDefaultJvmArgs = listOf(
         "--enable-native-access=ALL-UNNAMED",
         // Include the library paths where our dylibs are that we want to load and call
-        "-Djava.library.path=" + (javaLibraryPaths(rootDir) + javaLibraryPaths(project.projectDir)).joinToString(":"),
+        "-Djava.library.path=" + javaLibraryPaths(project.projectDir).joinToString(":"),
```
ktoso: Let's cleanup the : while we're here and use File.pathSeparator?
```suggestion
        "-Djava.library.path=" + javaLibraryPaths(project.projectDir).joinToString(File.pathSeparator),
```

PR #674, opened 2026-04-06
Add type name to macro-generated function names

sidepelican:
Functions generated by the `@JavaImplementation` macro currently use `context.makeUniqueName`.
While this ensures uniqueness within the local macro expansion context, it is not globally unique.
As a result, using the same function name across multiple classes leads to symbol collisions.

This PR adds the class name as a prefix to the generated function names to ensure global uniqueness.


ktoso:
Lgtm thanks! Nightly ci is broken right now, this is not on us

PR #696, opened 2026-04-10
jextract: Use fully qualified `java.lang.String` to avoid shadowing

sidepelican:
When generating Java code for next Swift type a nested Java record named `String` is created.

```swift
public enum MyEnum {
    ...
    case string(String)
}
```

```java
public final class JSON implements JNISwiftInstance {
  public record String(java.lang.String arg0) implements Case {
    record _NativeParameters(java.lang.String arg0) {}
  }
```

This causes the standard `java.lang.String` to be shadowed within that scope, leading to uncompilable source code where the generator expects the standard JDK `String` but finds the newly generated one instead.

To prevent this, the generator now uses the fully qualified name.

## Note

This applies to other types as well (such as `AtomicBoolean`), but I am only addressing String in this PR for now.


ktoso:
Lgtm!

PR #697, opened 2026-04-13
jextract: Implement identifier escaping for functions and enum cases

sidepelican:
Implements escaping for identifiers to prevent compilation errors when Swift function names use backticks or conflict with Java reserved keywords.

- Automatically removes backticks from escaped Swift identifiers (e.g., \`default\` → default).
- Appends an underscore as a suffix if an identifier conflicts with a Java reserved keyword (e.g., `default` → `default_`).

In this PR, these are applied to function names and enum cases.
I have skipped class names for now because the implementation is more complex, and such naming conflicts are considered rare at the class level.


ktoso (approved):

ktoso:
Nice one, lgtm

PR #699, opened 2026-04-13
Fix Java package mapping for dependent modules

iDevid:
The main goal of this PR is to preserve the mapping between dependent Swift modules and their Java packages so JNI generation can resolve cross-module Java references correctly.

While doing that, I also replaced tuple-based dependent-config plumbing with named structures to make the data flow clearer and less error-prone.

 A concrete example of this setup can be found in TriForce: https://github.com/iDevid/TriForce
  
In that sample project, the Android app depends on **SharedNetworkLayer**, which in turn depends on **SharedModels**.

### DIFF
--- a/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
+++ b/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
@@ -15,6 +15,19 @@
 import SwiftJavaJNICore
 
 extension JavaType {
+  var rawClassReference: String {
+    switch self {
+    case .class(let package, let name, _):
+      if let package, !package.isEmpty {
+        return "\(package).\(name)"
+      } else {
+        return name
+      }
+    default:
+      return description
+    }
+  }
+
   var jniTypeSignature: String {
     switch self {
     case .boolean: return "Z"
--- a/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
+++ b/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
@@ -16,19 +16,6 @@ import SwiftJavaJNICore
 
 extension JavaType {
 
-  var fullyQualifiedName: String {
-    switch self {
-    case .class(let package, let name, _):
-      if let package, !package.isEmpty {
-        return "\(package).\(name)"
-      } else {
-        return name
-      }
-    default:
-      return description
-    }
-  }
-
   var jniTypeSignature: String {
     switch self {
     case .boolean: return "Z"
--- a/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
+++ b/Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
@@ -15,7 +15,8 @@
 import SwiftJavaJNICore
 
 extension JavaType {
-  var rawClassReference: String {
+
+  var fullyQualifiedName: String {
     switch self {
     case .class(let package, let name, _):
       if let package, !package.isEmpty {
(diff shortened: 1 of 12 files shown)


Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
```diff
@@ -15,6 +15,19 @@
 import SwiftJavaJNICore
 
 extension JavaType {
+  var rawClassReference: String {
```
ktoso: I don't really like this one, this is just a qualified name, don't we have this already somewhere? Also why "raw"? Do you mean without generic bits?

ktoso: I see this is just used in 

```            
arguments: [.constant("\(javaType.rawClassReference)[]::new")]
```

maybe we just need `fullyQualifiedName` but I'm surprised we don't have it yet? This should not be called "raw" "reference" though -- maybe just rename it, thanks!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
```diff
@@ -32,6 +32,7 @@ package class JNISwift2JavaGenerator: Swift2JavaGenerator {
   let lookupContext: SwiftTypeLookupContext
 
   let javaClassLookupTable: JavaClassLookupTable
+  let dependentJavaPackages: [String: String]
```
ktoso: This needs docs, seems this is `swiftModuleName` to `javaPackage`? I'd maybe typealias those 

```
typealias SwiftModuleName = String
typealias JavaPackage = String
let dependentJavaPackages: [String: JavaPackage]
```

ktoso:
Some requests inline, thanks!

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
iDevid: Done 👍

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
iDevid: Declared a ModuleJavaPackages, following the style of JavaClassLookupTable, wdyt?

iDevid:
Thanks for the feedbacks @ktoso 
I've applied your suggestions, let me know if more changes are needed! 🙇

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
ktoso: That's good enough, thanks :)

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
```diff
@@ -15,6 +15,20 @@
 import SwiftJavaJNICore
 
 extension JavaType {
+
+  var fullyQualifiedName: String {
```
sidepelican: We have already defined this here:
https://github.com/swiftlang/swift-java-jni-core/blob/0c915b910175cf64a58fed99cde6494a2ef3d47a/Sources/SwiftJavaJNICore/JavaType%2BJavaSource.swift#L84

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
sidepelican: `className` is no longer in use, so is it generating a warning?

iDevid: Yes, it was, I removed the declaration

Sources/JExtractSwiftLib/Convenience/JavaType+Extensions.swift
iDevid: Thanks for the hint, I replaced `fullyQualifiedName` with the already available one

ktoso (approved):
Looks good now, thank you.

I've fixed the formatting issue and pushed

ktoso:
Thank you!

PR #700, opened 2026-04-13
Fix JNI generation for sending closures

iDevid:
This updates generated native translation code so async completion paths no longer capture non-Sendable JNI references directly inside `Task` closures. It also makes generated Java references fully qualified where needed to avoid ambiguity in generated calls.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1650,6 +1650,21 @@ extension JNISwift2JavaGenerator {
           """
         )
 
+        printer.print("struct _SwiftJavaUncheckedSendableBox<T>: @unchecked Sendable { let value: T }")
+        for globalRef in globalRefs {
+          printer.print("let \(globalRef)_sendable$ = _SwiftJavaUncheckedSendableBox(value: \(globalRef))")
+        }
+        if let selfParameter = nativeFunctionSignature.selfParameter {
+          for parameter in selfParameter.parameters {
+            printer.print("let \(parameter.name)$sendable$ = _SwiftJavaUncheckedSendableBox(value: \(parameter.name)$)")
+          }
+        }
+        if let selfTypeParameter = nativeFunctionSignature.selfTypeParameter {
+          for parameter in selfTypeParameter.parameters {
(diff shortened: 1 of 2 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1722,8 +1750,8 @@ extension JNISwift2JavaGenerator {
         printer.printHashIfBlock("swift(>=6.2)") { printer in
           printer.printBraceBlock("if #available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, *)") { printer in
             printer.printBraceBlock("task = Task.immediate") { printer in
-              // Immediate runs on the caller thread, so we don't need to attach the environment again.
-              printer.print("var environment = environment!") // this is to ensure we always use the same environment name, even though we are rebinding it.
+              // Even immediate tasks are a sending closure in Swift 6.2+, so reattach instead of capturing the caller's environment directly.
+              printer.print("var environment = try! JavaVirtualMachine.shared().environment()")
```
ktoso, line 1747: That's likely a good idea anyway, thanks

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1650,6 +1650,21 @@ extension JNISwift2JavaGenerator {
           """
         )
 
+        printer.print("struct _SwiftJavaUncheckedSendableBox<T>: @unchecked Sendable { let value: T }")
```
ktoso: I don't think we should silence the warning this way, with an ad hoc new type per every method

ktoso: Instead, the problem is that the jobject is not sendable but here we trust them to be fine... I suppose since they are global references.

So we should change the above code maybe to:

```
"nonisolated(unsafe) let \(outParameter.name) = environment.interface.NewGlobalRef(environment, \(outParameter.name))"
```

and re-bind the globalFuture as well perhaps, would that work?

ktoso (changes requested):

ktoso:
Thank you for the contribution! I think we need to adjust the approach a little bit though

iDevid:
Hi @ktoso, I managed to use your suggestion and make it work, let me know wdyt

ktoso:
Thank you, I think this is right here... seems like region analysis wouldn't trust us accessing these but we know they're safe

Issue #339, opened 2025-07-31
[JExtract] Support module qualified types


rintaro:
Currently `SwiftSymbolTable` doesn't support module qualified lookups. For example
```swift
public func receive(ptr: Swift.UnsafeRawPointer)
```
Importing this API fails because it doesn't resolve `Swift.UsnafeRawPointer`.

Supporting module qualified lookups is crucial for importing `.swiftinterface`.

timofeys1:
I stumbled onto this while trying to import a `.swiftinterface` from the `Samples/SwiftJavaExtractFFMSampleApp`. Building it with:
```
xcodebuild clean build -workspace Samples/SwiftJavaExtractFFMSampleApp/ -scheme SwiftJavaExtractFFMSampleApp -destination generic/platform=macOS -derivedDataPath dd SWIFT_EMIT_MODULE_INTERFACE=YES ARCHS=arm64
```
and then generating the java sources with:
```
./.build/arm64-apple-macosx/debug/swift-java jextract --swift-module MySwiftLibrary --input-swift dd/Build/Products/Debug/MySwiftLibrary.swiftmodule/arm64-apple-macos.swiftinterface --java-package com.example.swift --output-swift S1 --output-java J1
```
I see methods like:
```
public class MySwiftClass {
  ...
  public static func factory(len: Swift.Int, cap: Swift.Int) -> MySwiftLibrary.MySwiftClass
```
missing because of the qualified `MySwiftLibrary.MySwiftClass` return type

PR #704, opened 2026-04-14
Initial module qualified lookups

ktoso:
Pretty simple approach, if we recognize the name as a module, we try to make the lookup inside it.

@rintaro this is a simple approach but I wonder if may be good enough, if we recognize a module name, we try to lookup in it basically.

More complex follow up would be cross modules where the generated java does a fully qualified name for how the other module was imported as java package etc. 

Part of https://github.com/swiftlang/swift-java/issues/339

### DIFF
--- a/Sources/JExtractSwiftLib/SwiftTypes/SwiftSymbolTable.swift
+++ b/Sources/JExtractSwiftLib/SwiftTypes/SwiftSymbolTable.swift
@@ -153,9 +153,15 @@ extension SwiftSymbolTable: SwiftSymbolTableProtocol {
       }
     }
 
-    // FIXME: Implement module qualified name lookups. E.g. 'Swift.String'
-
     return nil
+  }
+
+  /// Look for a top-level nominal type in a specific module by name
+  package func lookupTopLevelNominalType(_ name: String, inModule moduleName: String) -> SwiftNominalTypeDeclaration? {
+    if moduleName == self.moduleName {
+      return parsedModule.lookupTopLevelNominalType(name)
+    }
+    return importedModules[moduleName]?.lookupTopLevelNominalType(name)
   }
 
   // Look for a nested type with the given name.
--- a/Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
+++ b/Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
@@ -353,19 +353,21 @@ extension SwiftType {
       self = knownTypes.optionalSugar(try SwiftType(optionalType.wrappedType, lookupContext: lookupContext))
 
     case .memberType(let memberType):
-      // If the parent type isn't a known module, translate it.
-      // FIXME: Need a more reasonable notion of which names are module names
-      // for this to work. What can we query for this information?
+      // If the parent type is a known module name, perform a module-qualified
+      // lookup instead of treating the module as a parent type
       let parentType: SwiftType?
+      let moduleName: String?
       if let base = memberType.baseType.as(IdentifierTypeSyntax.self),
         lookupContext.symbolTable.isModuleName(base.name.trimmedDescription)
       {
         parentType = nil
+        moduleName = base.name.trimmedDescription
       } else {
         parentType = try SwiftType(memberType.baseType, lookupContext: lookupContext)
+        moduleName = nil
       }
 
-      // Translate the generic arguments.
+      // Translate the generic arguments
       let genericArgs = try memberType.genericArgumentClause.map { genericArgumentClause in
         try genericArgumentClause.arguments.map { argument in
           switch argument.argument {
@@ -382,7 +384,8 @@ extension SwiftType {
         parent: parentType,
         name: memberType.name,
         genericArguments: genericArgs,
-        lookupContext: lookupContext
+        lookupContext: lookupContext,
+        module: moduleName
       )
 
     case .metatypeType(let metatypeType):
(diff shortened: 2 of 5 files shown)


Sources/JExtractSwiftLib/SwiftTypes/SwiftSymbolTable.swift
```diff
@@ -153,11 +153,17 @@ extension SwiftSymbolTable: SwiftSymbolTableProtocol {
       }
     }
 
-    // FIXME: Implement module qualified name lookups. E.g. 'Swift.String'
-
     return nil
   }
 
+  /// Look for a top-level nominal type in a specific module by name
+  package func lookupTopLevelNominalType(_ name: String, inModule moduleName: String) -> SwiftNominalTypeDeclaration? {
```
ktoso, line 160: just like the one above but with a module name

```
  package func lookupTopLevelNominalType(_ name: String) -> SwiftNominalTypeDeclaration? {


```

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
```diff
@@ -353,19 +353,21 @@ extension SwiftType {
...
       {
         parentType = nil
+        moduleName = base.name.trimmedDescription
       } else {
         parentType = try SwiftType(memberType.baseType, lookupContext: lookupContext)
+        moduleName = nil
       }
 
-      // Translate the generic arguments.
+      // Translate the generic arguments
```
rintaro: Accidental change?

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
```diff
@@ -431,7 +434,8 @@ extension SwiftType {
     parent: SwiftType?,
     name: TokenSyntax,
     genericArguments: [SwiftType]?,
-    lookupContext: SwiftTypeLookupContext
+    lookupContext: SwiftTypeLookupContext,
+    module: String? = nil
```
rintaro, line 438: At first glance, parent and module are mutually exclusive so I'd make an enum for "type parent". But on second thought, for module selector types e.g. `Foo.Module::Bar`, both are required, so accepting both is fine.

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
rintaro, line 349: It's highly likely `.swiftinterface` will adopt module selectors everywhere. So handling `IdentifierTypeSyntax.moduleSelector` will be needed soon.

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
```diff
@@ -353,19 +353,21 @@ extension SwiftType {
...
       let parentType: SwiftType?
+      let moduleName: String?
       if let base = memberType.baseType.as(IdentifierTypeSyntax.self),
         lookupContext.symbolTable.isModuleName(base.name.trimmedDescription)
       {
         parentType = nil
+        moduleName = base.name.trimmedDescription
       } else {
         parentType = try SwiftType(memberType.baseType, lookupContext: lookupContext)
+        moduleName = nil
```
rintaro, line 367: Similarly, handling `MemberTypeSyntax.moduleSelector` will be needed _in the near future_

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
```diff
@@ -440,6 +444,8 @@ extension SwiftType {
         throw TypeTranslationError.unknown(originalType)
       }
       typeDecl = lookupContext.symbolTable.lookupNestedType(name.text, parent: parentDecl)
```
rintaro, line 446: So `lookupNestedType` will also need `inModule:` variation (in the future)

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
```diff
@@ -440,6 +444,8 @@ extension SwiftType {
...
+    } else if let module {
+      typeDecl = lookupContext.qualifiedLookup(name: name.text, inModule: module)
```
rintaro: "qualified lookup" might be confusing with regular nested type lookups. Maybe `moduleQualifiedLookup(name:in:)` sounds better, but it's your call.

rintaro (commented):
Looks reasonable overall, thanks!

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
ktoso: Yeah I think both can co exist in the future like that

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
ktoso: Gotcha, thank you!

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
ktoso: I was actually iffy on the name, thank you! Qualified lookup would mean qualified inside types, not inside modules right? Thanks

Sources/JExtractSwiftLib/SwiftTypes/SwiftType.swift
ktoso: Undoing :)

PR #707, opened 2026-04-17
jextract: Fix duplicate `rawValue` generation for enums

sidepelican:
Currently, providing an enum that explicitly defines `rawValue` causes the generator to produce duplicate `rawValue` implementations.

```swift
public enum DeliveryStage : Swift.String, Swift.Sendable & Swift.Codable {
  case schedule
  case delivered
  public init?(rawValue: Swift.String)
  public typealias RawValue = Swift.String
  public var rawValue: Swift.String {
    get
  }
}
```

```
error: invalid redeclaration of 'Java_com_example_swift_DeliveryStage__00024getRawValue__J(environment:thisClass:selfPointer:)'
```

There was a simple logic error in the generator.


ktoso (approved):
Thank you!

PR #709, opened 2026-04-17
jextract/jni: Move enum case classes inside the `Case` interface

sidepelican:
Currently, generating Java code for a Swift enum where a case name matches a nested type name (as shown below) results in a compilation error.

```swift
public enum MyResponse: Codable {
  case success(Success)
  public struct Success {
    public init(message: String) {
      self.message = message
    }
    public var message: String
  }
  case failure(String)
}
```

This happens because `jextract` generates two Java classes named `Success` at the same nesting level—one for the enum case and one for the Swift struct.

To avoid this, I have moved the generated case classes inside the `interface Case`.

### Note

While a collision could still theoretically occur if the Swift enum defines its own `struct Case`, the probability of encountering such a conflict is significantly lower with this new structure.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/VehicleEnumTest.java
+++ b/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/VehicleEnumTest.java
@@ -92,7 +92,7 @@ public class VehicleEnumTest {
     void getAsBicycle() {
         try (var arena = SwiftArena.ofConfined()) {
             Vehicle vehicle = Vehicle.bicycle(arena);
-            Vehicle.Bicycle bicycle = vehicle.getAsBicycle().orElseThrow();
+            Vehicle.Case.Bicycle bicycle = vehicle.getAsBicycle().orElseThrow();
             assertNotNull(bicycle);
         }
     }
@@ -101,7 +101,7 @@ public class VehicleEnumTest {
     void getAsCar() {
         try (var arena = SwiftArena.ofConfined()) {
             Vehicle vehicle = Vehicle.car("BMW", Optional.empty(), arena);
-            Vehicle.Car car = vehicle.getAsCar().orElseThrow();
+            Vehicle.Case.Car car = vehicle.getAsCar().orElseThrow();
             assertEquals("BMW", car.arg0());
 
             vehicle = Vehicle.car("BMW", Optional.of("Long trailer"), arena);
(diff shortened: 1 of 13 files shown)


Tests/JExtractSwiftTests/JNI/JNIEnumTests.swift
```diff
@@ -163,7 +163,17 @@ struct JNIEnumTests {
       detectChunkByInitialLines: 1,
       expectedChunks: [
         """
-        public sealed interface Case {}
+        public sealed interface Case {
+          record First() implements Case {
```
sidepelican, line 167: Note: `public` is unnecessary within an interface.

Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/VehicleEnumTest.java
```diff
@@ -129,7 +129,7 @@ void getAsMotorbike() {
     void getAsTransformer() {
         try (var arena = SwiftArena.ofConfined()) {
             Vehicle vehicle = Vehicle.transformer(Vehicle.bicycle(arena), Vehicle.car("BMW", Optional.empty(), arena), arena);
-            Vehicle.Transformer transformer = vehicle.getAsTransformer(arena).orElseThrow();
+            Vehicle.Case.Transformer transformer = vehicle.getAsTransformer(arena).orElseThrow();
```
ktoso, line 132: The alternative would be to `Vehicle.TransformerCase` or `Vehicle.$Transformer` hm... The $ in the name is not ideal perhaps, we usually tell folks that things with $ in the name are "not for them" but the Case sufficient could work without the additional nesting...

WDYT, any opinions @madsodgaard ?

Sources/JExtractSwiftLib/JavaTypes/JavaType+JDK.swift
```diff
@@ -45,6 +45,11 @@ extension JavaType {
     .class(package: "java.lang", name: "Object")
   }
 
+  /// The description of the type java.util.Optional.
+  static func optional(_ T: JavaType) -> JavaType {
+    .class(package: "java.util", name: "Optional", typeParameters: [T])
+  }
```
ktoso, line 51: nice

ktoso (commented):
I don't dislike this tbh... let's give @madsodgaard  a chance to reply as well

madsodgaard:
Yeah, I've faced this as well. Moving it into `Case` seems fine, I don't see any issues with it.

ktoso:
Okey, I'm happy with the .Case., it's fine tbh :) Thank you !

ktoso (approved):

PR #712, opened 2026-04-19
Allow swift-collections up to next major

tkrajacic:
Not sure why this was so restrictive


ktoso (approved):
I can't remember a good reason for that, let's relax it, thank you

PR #717, opened 2026-04-22
jextract/jni: Use switch expressions in `getCase()`

sidepelican:
Using a switch expression for the `Case getCase()` implementation makes the code cleaner.
Since switch expressions are exhaustive, it eliminates the need to throw a `RuntimeException`.


ktoso (approved):

ktoso:
Oh wait... which java source version does this require...

sidepelican:
It became available in Java 14, and given the current use of sealed classes, I believe this is perfectly acceptable.

ktoso:
Sweet, I was just checking now -- 14 is perfectly fine :-) 
Thank you

PR #719, opened 2026-04-22
jextract: Suppress InternalImportsByDefault related warnings

sidepelican:
After https://github.com/swiftlang/swift-java/pull/710 , I started seeing numerous warnings.
Until `InternalImportsByDefault` is properly handled, I want to prevent these warnings from appearing in normal usage.


ktoso (approved):
LGTM, good idea

PR #720, opened 2026-04-23
jextract/jni: Support optional tuple return values

sidepelican:
Currently, code generation is skipped when an optional tuple is used as a return value.
This PR fixes this.

Support for optional tuples as parameters is currently out of scope.
Given the difficulty, I've decided to postpone this to a subsequent PR.

### Architectural Concerns & Refactoring

During the implementation, I encountered a bottleneck where `JNISwift2JavaGenerator+JavaTranslation.swift` needed to reference results from `JNISwift2JavaGenerator+NativeTranslation.swift`.

The current architecture handles Java-side and Swift-side glue code generation independently.
This separation makes it increasingly difficult to keep the generation strategies in sync and ensure consistency between the two sides.

### Future Work & Current Workaround

I believe unifying these generation logics will be essential in the future. 
However, given the current scale of the implementation, such a major refactoring is extremely difficult at this stage.

As a result, this PR addresses the immediate issue using a workaround to bypass these structural limitations for now, while acknowledging the underlying architectural debt.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Optionals.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Optionals.swift
@@ -91,3 +91,11 @@ public func multipleOptionals(
 ) -> Int64? {
   1
 }
+
+public func optionalTuple() -> (Int64, String)? {
+  (42, "hello")
+}
+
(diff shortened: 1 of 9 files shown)


sidepelican:
I understand the hesitation to merge this PR due to the workaround included.
I'll refine it further to make it more incrementally improved.

PR #722, opened 2026-04-24
jextract/jni: Add `nativeJavaType` to `TranslatedResult` for explicit type handling

sidepelican:
This PR is a refactoring effort to address the architectural concerns proposed in #720.

Previously, the return types of native functions were handled implicitly within `JavaTranslation`.
This PR changes that to an explicit exchange using the newly added `nativeJavaType` in `TranslatedResult`.

This change resolves existing `// FIXME:` comments in the tuple generation logic.

To ensure that `JavaTranslation` and `NativeJavaTranslation` generate identical signatures for native functions, I have added an assertion to validate the correctness of `nativeJavaType`.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
@@ -117,13 +117,20 @@ extension JNISwift2JavaGenerator {
 
       let conversions = try enumCase.parameters.enumerated().map { idx, parameter in
         let resultName = parameter.name ?? "arg\(idx)"
-        let result = SwiftResult(convention: .direct, type: parameter.type)
-        var translatedResult = try self.translate(swiftResult: result, methodName: methodName, resultName: resultName)
+        var translatedResult = try self.translateResult(
+          swiftType: parameter.type,
+          methodName: methodName,
+          resultName: resultName
+        )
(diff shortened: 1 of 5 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -1003,14 +1013,15 @@ extension JNISwift2JavaGenerator {
         } else {
           return TranslatedResult(
             javaType: javaType,
+            nativeJavaType: .long,
```
sidepelican, line 1016: This isn't a new concern; it just makes an existing assumption explicit.
For example, using `.wrapMemoryAddressUnsafe` already implies we know the native return type is `long`.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -323,7 +331,7 @@ extension JNISwift2JavaGenerator {
         )
       }
 
-      let translatedResult = try translate(swiftResult: SwiftResult(convention: .direct, type: swiftType.resultType), methodName: name)
+      let translatedResult = try translateResult(swiftType: swiftType.resultType, methodName: name)
```
ktoso, line 334: Ok change thx

ktoso (approved):
Looks good I think, thanks

PR #723, opened 2026-04-24
jextract/jni: Support optional tuple return values 2

sidepelican:
This is an improved version of #720.

Currently, code generation is skipped when an optional tuple is used as a return value.
This PR addresses this limitation.

Support for optional tuples as parameters remains out of scope for this PR.

Thanks to the architectural improvements in #722, I'm now able to handle these tuples properly without the need for any workarounds.


ktoso (approved):
Phone review but looks good, thank you!

PR #724, opened 2026-04-27
jextract/jni: Fix code generation for optional `String` in tuples

sidepelican:
Previously, there was a special handling logic for `String` within tuples, but this logic was not applied to `String?`, leading to incorrect output.

By leveraging the `nativeJavaType` introduced in #722, I was able to remove this special-case logic entirely.


ktoso (approved):
Love removing edge cases :-)

PR #725, opened 2026-04-27
jextract/jni: Handle single-element tuples correctly

sidepelican:
Code generation fails when a single-element tuple is used. In Swift, `(T)` is semantically identical to `T`, so attempting to access it via `.0` causes a compiler error.

```swift
1375 |   let element_0_jni$ = tupleResult$.0.getJNILocalRefValue(in: environment)
     |                                     `- error: value of type 'String' has no member '0'
```

I've updated the generator to treat tuples with a single element as a single unwrapped value instead of a tuple.


ktoso (approved):
Sounds good, happy to drop the Tuple1

PR #727, opened 2026-04-30
jextract/jni: Overhaul enum associated value retrieval

sidepelican:
This PR overhauls the implementation of enum associated value retrieval.
By leveraging the recent improvements in tuple handling, I can now support complex patterns (such as arrays and generic types).

## The Problem

The current `getAsCase` implementation relies on specialized, hard-coded logic. 

https://github.com/swiftlang/swift-java/blob/6e97fc0c5b571bf2103a5132f8abfe7bc0d90c28/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator%2BJavaTranslation.swift#L161-L221

It only works for very simple patterns.
Adding support for complex types (tuples, arrays, etc.) within the current structure is not sustainable and would lead to code duplication.

## The Solution

Instead of a specialized path, I have unified the logic with our existing tuple infrastructure. 

In Swift, associated values are extracted as an optional tuple (e.g., `(Foo, String)?`).
In Java, the glue code wraps it into the appropriate `Case` class.

### Example

Original:

```swift
enum MyEnum {
  case foo(Foo, String)
}
```

Generated:

```swift
extension MyEnum {
  fileprivate func getAsFoo() -> (Foo, String)? {
    if case let .foo(_0, _1) = self {
      return (_0, _1)
    }
    return nil
  }
}

// native $getAsFoo implementation
@_cdecl(...)
public func Java_com_example_swift_MyEnum__00024getAsFoo...
```

```java
public Optional<Case.Foo> getAsFoo() {
  MyEnum.$getAsFoo(...);
  ...
  Optional<Tuple2<Foo, String>> associatedValues$ = ...; // get results as optional tuple
  // Wrap it in a Case class
  return associatedValues$.map((t) -> {
    return new Case.Foo(t.$0, t.$1);
  }
}
private static native void $getAsFoo(long selfPointer, byte[] result_discriminator$, ...);
```

## Benchmarks

I have tested via `./gradlew jmh -PjmhIncludes='.EnumBenchmark*'` in `Samples/SwiftJavaExtractJNISampleApp/`.

- before

```
Benchmark                          Mode  Cnt    Score   Error  Units
EnumBenchmark.getAssociatedValues  avgt   30  974.888 ± 8.462  ns/op
```

- after

```
Benchmark                          Mode  Cnt    Score   Error  Units
EnumBenchmark.getAssociatedValues  avgt   30  750.577 ± 8.338  ns/op
```

The performance improvement is likely a direct result of reducing the number of JNI calls.

## Note

~~This PR is currently a draft as it depends on #725. I will mark it as ready for review once the dependency is merged.~~


ktoso (approved):
Neat improvement, lgtm!

PR #728, opened 2026-05-01
jextract/jni: Use boolean instead of byte for Optional discriminator

sidepelican:
This PR changes the discriminator type used to distinguish between `some` and `none` in Optional values from `byte` to `boolean`.

Using a `boolean` is more intuitive for representing a binary state. 
This change also simplifies the generated code by reducing the reliance on ternary operators.


ktoso:
The idea here was to pack multiple marker bits into the number rather than pass multiple booleans when we have many more optionals. So I'm not sure we want to take this change

sidepelican:
I see.
Keeping it as a byte makes sense for future optimization.
I’ll close this PR as it's not a high-priority update. Thanks for the explanation!

PR #729, opened 2026-05-01
wrap-java: Fix escaped method calls and handle $ prefix

sidepelican:
This PR fixes two issues related to Java-to-Swift method name mapping in wrap-java.

1. Fix for Escaped Method Calls

Currently, Java methods that are Swift keywords (e.g., `init`) are correctly escaped in the Swift interface. 
However, the macro was using these escaped names to call the Java methods, leading to failures.

expanded macro

This PR ensures the original name is passed to the macro arguments when a name is escaped.

2. Handling of `$` Prefix

Methods with a `$` prefix were previously ignored.
To support future use cases like `SwiftInstance.$memoryAddress()`, I have enabled importing these methods by mapping the `$` prefix to an underscore.

### DIFF
--- a/Sources/SwiftJavaToolLib/StringExtras.swift
+++ b/Sources/SwiftJavaToolLib/StringExtras.swift
@@ -27,17 +27,19 @@ extension String {
   }
 
   /// Escape a name with backticks if it's a Swift keyword.
-  var escapedSwiftName: String {
-    if isValidSwiftIdentifier(for: .variableName) {
-      return self
+  var escapedSwiftName: (swiftName: String, escaped: Bool) {
+    var escaped = false
+    var copy = self
+    if starts(with: "$") {
+      copy = "_\(self.dropFirst())"
+      escaped = true
     }
 
-    return "`\(self)`"
-  }
+    if copy.isValidSwiftIdentifier(for: .variableName) {
+      return (copy, escaped)
+    }
 
-  /// Returns whether this is a valid Swift function name
-  var isValidSwiftFunctionName: Bool {
-    !self.starts(with: "$")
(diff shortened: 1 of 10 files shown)


Sources/SwiftJavaToolLib/StringExtras.swift
```diff
@@ -27,17 +27,19 @@ extension String {
...
 
   /// Escape a name with backticks if it's a Swift keyword.
-  var escapedSwiftName: String {
-    if isValidSwiftIdentifier(for: .variableName) {
-      return self
+  var escapedSwiftName: (swiftName: String, escaped: Bool) {
+    var escaped = false
+    var copy = self
+    if starts(with: "$") {
+      copy = "_\(self.dropFirst())"
```
sidepelican, line 34: Swift does not allow `$` even when wrapped in backticks.
Therefore, always perform replacing.

```swift
❯ echo 'func `$foo`() {}' | swiftc -
<stdin>:1:6: error: cannot declare entity named '$foo'; the '$' prefix is reserved for implicitly-synthesized declarations
1 | func `$foo`() {}
  |      `- error: cannot declare entity named '$foo'; the '$' prefix is reserved for implicitly-synthesized declarations
2 | 
```

ktoso (approved):
Hm yeah that’s reasonable I guess… otherwise we could do some _dollar$ or something like that but I don’t love those…

Sources/SwiftJavaToolLib/StringExtras.swift
ktoso: Sounds good, thanks for checking the ‘’

PR #730, opened 2026-05-02
Fix async JNI strict concurrency

nerzh:
Problem found:

Swift Java generated JNI wrappers for `async` Swift functions no longer compile with newer Swift strict concurrency checks. The generated `Task` closures capture JNI values and temporary Swift conversion values that are not `Sendable`, so Swift reports data-race risk errors during compilation.

How it was fixed:

I added an unchecked `Sendable` wrapper for values that must cross into the generated async `Task` body. The generator now tracks, structurally, which temporary values are created during JNI-to-Swift parameter conversion and wraps only those values before creating the task. Inside the task, the original values are restored from the wrapper and used as before.

This avoids fragile string parsing of generated Swift code and keeps the fix limited to async JNI generation. Existing async JNI tests were updated to match the new generated output.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -232,7 +232,8 @@ extension JNISwift2JavaGenerator {
           ],
           conversion: .pointee(.extractSwiftValue(.placeholder, swiftType: type)),
           indirectConversion: nil,
-          conversionCheck: nil
+          conversionCheck: nil,
+          asyncTaskCaptureNames: [NativeSwiftConversionStep.extractedSwiftValueName(for: parameterName)]
         )
 
       case .tuple([]):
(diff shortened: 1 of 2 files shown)


ktoso:
Hi thanks for the PR, I’m on holiday so unable to review deeply here this week.

im worried about unsafe wrappers like this, and would like to explore other options

nerzh:
@ktoso yes, I agree. I used an unsafe wrapper to avoid changing the approach you were using, because previously there was already an unsafe constant there 🫣

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1072,6 +1079,24 @@ extension JNISwift2JavaGenerator {
 
     /// Represents check operations executed in if/guard conditional block for check during conversion
     let conversionCheck: NativeSwiftConversionCheck?
+
+    /// Temporary values created by this parameter conversion that must be moved
+    /// into async task bodies through an unchecked Sendable wrapper.
+    let asyncTaskCaptureNames: [String]
```
sidepelican: `NativeParameter` represents a single parameter for original swift functions.
It seems unnatural to include the names of multiple values you want to send.

nerzh:
@sidepelican I updated this PR, and if I understood you correctly, I tried to move the async capture information out of `NativeParameter` and into the conversion steps

sidepelican:
@nerzh It seems my explanation wasn't clear enough.
I meant to say that the additional property should be `let asyncTaskCaptureName: String`.

It seems multiple `CaptureNames` are required in the tuple, but this can be properly avoided by temporarily placing the tuple into a variable.

I'd like you to first define an async function taking a tuple argument in `Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Async.swift`, then verify whether `swift build --disable-sandbox` works correctly.

PR #731, opened 2026-05-07
jextract: Support nested typealiases

sidepelican:
This PR adds support for nested `typealias` declarations.
Previously, typealiases located within nested namespaces were not supported.


ktoso (approved):
Looks good to me!

PR #734, opened 2026-05-11
jextract: Fix compilation errors in constrained generic extensions

sidepelican:
This PR addresses compilation errors occurring in the generated "opener" code for certain constrained generic extensions. 

1. Protocol-constrained extensions

```swift
extension Foo where Element: SomeProtocol {
    public func something() {} 
}
```

When an extension has a protocol requirement, the functions were being exported into "opener" code that does not satisfy the protocol constraint, leading to a build failure.

2. Same-type constraints with type parameters on the RHS

```swift
public struct Box<Element> {}
public struct Bait {}

extension Box where Bait == Element {
  public func swappedObserveTheBait() {}
}
```

In cases like `where Bait == Element`, the generator was incorrectly exporting functions.
Similar to the protocol case, the opener fails to satisfy the same-type requirement when a type parameter is involved on the right-hand side.

I have updated the `Swift2JavaVisitor` implementation to correctly identify these problematic constraints.

### DIFF
--- a/Sources/JExtractSwiftLib/Swift2JavaVisitor.swift
+++ b/Sources/JExtractSwiftLib/Swift2JavaVisitor.swift
@@ -31,12 +31,12 @@ final class Swift2JavaVisitor {
   var log: Logger { translator.log }
 
   /// Constrained extensions deferred until specializations are applied
-  private var deferredConstrainedExtensions:
-    [(
-      node: ExtensionDeclSyntax,
-      sourceFilePath: String,
-      sameConstraint: [(String, String)]
-    )] = []
+  private struct DeferredConstrainedExtension {
+    var node: ExtensionDeclSyntax
+    var sourceFilePath: String
+    var constraints: [(String, String)]
+  }
+  private var deferredConstrainedExtensions: [DeferredConstrainedExtension] = []
 
   func visit(inputFile: SwiftJavaInputFile) {
     let node = inputFile.syntax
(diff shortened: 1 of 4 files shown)


Sources/JExtractSwiftLib/Swift2JavaVisitor.swift
```diff
@@ -31,7 +31,12 @@ final class Swift2JavaVisitor {
   var log: Logger { translator.log }
 
   /// Constrained extensions deferred until specializations are applied
-  private var deferredConstrainedExtensions: [(ImportedNominalType, ExtensionDeclSyntax, String)] = []
+  private var deferredConstrainedExtensions:
+    [(
+      node: ExtensionDeclSyntax,
+      sourceFilePath: String,
+      sameConstraint: [(String, String)]
+    )] = []
```
ktoso: I'd rather have a type than a big tuple like this

ktoso (approved):

Sources/JExtractSwiftLib/Swift2JavaVisitor.swift
```diff
@@ -623,18 +650,18 @@ final class Swift2JavaVisitor {
...
+        return true
+      }
+      if specialized.genericArguments[rhs] == lhs {
+        return true
       }
-      // If the param isn't in the mapping, we allow it (might be a secondary constraint)
     }
-    return true
+    return false
   }
```
ktoso, line 665: Hmmm this seems suspicious, one sec

ktoso: Yeah this is wrong because we must match _all_ same type requirements, I'll push a fix.

PR #735, opened 2026-05-11
jextract/ffm: Treat underscore labels as empty suffixes in function overloads

sidepelican:
Currently, when a function overload uses an underscore (`_`) as an argument label, the generator uses the underscore literally as a suffix for the Java method name.

```swift
func foo(a: Int)    // in Java, void fooA(long a)
func foo(_ b: Int)  // in Java, void foo_(long b)
```

In Swift, `_` is not an identifier; it simply denotes the omission of a label. Therefore, IMO, using it as a literal suffix in Java feels unnatural.

This PR updates the naming logic for function overloads to treat the `_` label as an empty suffix.

## Notes

This change can lead to name collisions in specific cases. For example:

```swift
public func takeValue(_ a: Int, b: Int) -> Int { 0 }
public func takeValue(b: Int, _ c: Int) -> Int { 0 }
```

With this PR, both methods would map to `takeValueB`.
However, even if we kept the underscore (the previous behavior), collisions would still be unavoidable if a function like `func takeValue_(b: Int)` existed.
To approach such corner-case collisions, we should eventually provide a way for users to explicitly specify the export name. (like `@JavaExport("export_name")`).

### DIFF
--- a/Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftLibrary.swift
+++ b/Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftLibrary.swift
@@ -165,6 +165,10 @@ public func globalOverloaded(b: Int) {
   p("globalOverloaded(b: \(b))")
 }
 
+public func globalOverloaded(_ c: Int) {
+  p("globalOverloaded(c: \(c))")
+}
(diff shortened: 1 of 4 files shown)


ktoso:
This sounds good to me, long term solution as well, thank you! 

Maybe worth a new issue about the longer term fix

PR #736, opened 2026-05-12
Fix extension constraint matching for multiple constraints

ktoso:
All same type constraints must match, not just one of them. This is a follow up after https://github.com/swiftlang/swift-java/pull/734#discussion_r3223126771

https://github.com/swiftlang/swift-java/pull/734 FYI @sidepelican


sidepelican:
You're right, that was indeed a mistake. Thank you for the thorough check and follow-up!

ktoso:
No problem, that's what reviews are for :)

PR #738, opened 2026-05-12
jextract/jni: Fix compilation of nested generic types

sidepelican:
Currently, nested generic types cause compilation errors because the generator fails to include necessary namespaces and generic parameters.

```swift
// MARK: original declaration
public enum NamespaceEnum {
  public struct GenericType<T> {
    public var value: Int
  }
}

// MARK: generated code

//        ↓ missing namespace
extension GenericType: _MySwiftLibrary_GenericType_opener {
  static func _get_value(...) -> jlong {
    ...
    let selfPointer$ = UnsafeMutablePointer<NamespaceEnum.GenericType>(bitPattern: selfPointerBits$)
    //                                    missing type parameter <T> ↑
    ...
  }
```

Based on my experiments, Swift handles type inference differently for nested generics compared to top-level generics.
While top-level generic parameters can often be inferred within an extension, nested generic types require explicit type parameters when referenced inside the extension body.

```swift
struct TopLevelGeneric<T> {}

extension TopLevelGeneric {
    static func foo() -> String {
        "\(TopLevelGeneric.self)" // OK
    }
}

enum Nested {
    struct Generic<T> {}
}

extension Nested.Generic {
    static func foo() -> String {
        "\(Nested.Generic.self)" // error: generic parameter 'T' could not be inferred
    }
}
```

To resolve this, I have updated the generator to:

1.	Use the qualified name for extension declarations.
2.	Use the full type name including generic parameters within function bodies.
    The root cause was that `SwiftType` in `SwiftFunctionSignature` was missing its generic parameters.
    It appears that `ImportedNominalType.swiftType` was incorrectly stripping these parameters during the type conversion process.
    This PR fixes that logic to ensure generic information is preserved.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/NestedTypes.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/NestedTypes.swift
@@ -51,7 +51,7 @@ public enum NamespaceEnum {
     public static func something() {}
   }
 
-  public struct GenericType<T> {
+  public struct NestedGenericType<T> {
     public var value: Int
   }
 }
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/NestedTypes.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/NestedTypes.swift
@@ -50,4 +50,8 @@ public enum NamespaceEnum {
   public enum Nested {
     public static func something() {}
   }
+
+  public struct GenericType<T> {
+    public var value: Int
+  }
 }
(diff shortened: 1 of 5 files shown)


Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/NestedTypes.swift
```diff
@@ -50,4 +50,8 @@ public enum NamespaceEnum {
   public enum Nested {
     public static func something() {}
   }
+
+  public struct GenericType<T> {
```
ktoso:
```suggestion
  public struct NestedGenericType<T> {
```

to make it easier to spot what this is

ktoso (approved):

PR #739, opened 2026-05-12
Make SwiftNominalType.genericArguments non-optional

sidepelican:
This is a minor refactoring to change `SwiftNominalType.genericArguments` from an optional array to a non-optional one.

- In practice, we don't need to distinguish between `nil` and an empty array `[]`.
- Improves consistency since `JavaType` already maintains non-optional type parameters.


ktoso (approved):
Always happy to take such cleanups, sounds good!

ktoso:
PS: I'm back from holiday so, as you may have noticed, more responsive again :-) Thanks for the continued PRs!

PR #744, opened 2026-05-14
Revert "more workarounds for nested swift-builds"

ktoso:
This reverts commit 06947f18f77827835a4a446f843d4a04a386e7bf.

This PR is to see if the issue is resolved in more recent nightlies.


marcprux:
We had the same issue with Skip's invocation of Android builds.

We addressed this by [filtering _everything_](https://github.com/skiptools/skipstone/blob/7624efafd3bde0a92788cfa304bc015a31e3e415/Sources/SkipBuild/Commands/AndroidCommand.swift#L765) out of the environment except for a pre-approved-list of variables when running from Xcode (as determined by checking the `XCODE_VERSION_MAJOR` environment variable).

In addition to filtering out `SDKROOT`, this can also be important for ensuring that incremental builds work, since some of Xcode's ever-changing environment variables (like `LLBUILD_TASK_ID`) will cause Swift to think that a complete rebuild is necessary every time. I'm not sure if this is still true with swiftbuild, but it was leading to very slow rebuilds for us.

ktoso:
Hah, interesting, thanks for the info Marc! I wonder then if we just need t keep this...

Issue #747, opened 2026-05-14
Trim the CI jobs a bit -- too many variants


ktoso:
We have too many CI jobs now due to the fact that we check 6.1, 6.2, 6.3, 6.4, linux, mac, android... 

We likely can skip 6.2 in samples, and focus on the others.

I'd keep 6.1 on linux as double check; and then keep macos on latest only maybe 6.3 and snapshot.

swift-test would be on 6.1 and 6.3 as well

PR #749, opened 2026-05-14
Lessen the number of CI runs

ktoso:
Reduce 64 builds down to 49 + soundness, which still is a lot but we don't need ALL versions everywhere testing ALL samples.

We don't test 6.1 on macos anymore, just current 6.3 and next 6.4-snapshot because those runners are precious.

Benchmark doesn't do much, combine jmh and swift.


ktoso:
resolves https://github.com/swiftlang/swift-java/issues/747

PR #751, opened 2026-05-14
fix optional Int JNI conversion

nerzh:
I noticed that Int? / UInt? JNI bindings did not use the same 32-bit-safe conversion path as plain Int / UInt. I fixed it by routing optional values through the existing indirect conversion and overflow checks 🫣

```
/oberton-wallet-core/.build/plugins/outputs/cust-wallet-core/CustWalletCore/destination/JExtractSwiftPlugin/Sources/AbiTypes+SwiftJava.swift:3755:28: error: cannot assign value of type 'Int.JNIType' (aka 'Int32') to type 'jlong' (aka 'Int64')
3753 |   let result$: jlong
3754 |   if let innerResult$ = selfPointer$.pointee.balance {
3755 |     result$ = innerResult$.getJNIValue(in: environment)
     |                            `- error: cannot assign value of type 'Int.JNIType' (aka 'Int32') to type 'jlong' (aka 'Int64')
3756 |     var flag$ = Int8(1)
3757 |     environment.interface.SetByteArrayRegion(environment, result_discriminator$, 0, 1, &flag$)

/oberton-wallet-core/ObertonWalletCore/destination/JExtractSwiftPlugin/Sources/AbiTypes+SwiftJava.swift:3778:77: error: cannot convert value of type 'jlong' (aka 'Int64') to expected argument type 'Int.JNIType' (aka 'Int32')
3776 |     fatalError("selfPointer memory address was null in call to \(#function)!")
3777 |   }
3778 |   selfPointer$.pointee.balance = newValue_discriminator == 1 ? Int(fromJNI: newValue_value, in: environment) : nil
     |                                                                             `- error: cannot convert value of type 'jlong' (aka 'Int64') to expected argument type 'Int.JNIType' (aka 'Int32')
3779 | } // printCDecl(_:javaMethodName:parentName:parameters:resultType:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+SwiftThunkPrinting.swift:818
3780 | 

```

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -485,18 +485,30 @@ extension JNISwift2JavaGenerator {
               throw JavaTranslationError.unsupportedSwiftType(swiftType)
             }
 
+            let indirectStepType = JNIJavaTypeTranslator.indirectConversionStepSwiftType(
+              for: knownType,
+              from: knownTypes
+            )
+            let indirectCheck = JNIJavaTypeTranslator.checkStep(for: knownType, from: knownTypes)
+            let valueConversion: NativeSwiftConversionStep =
+              if indirectStepType != nil {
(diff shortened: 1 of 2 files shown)


ktoso (approved):
Looks great, thank you for catching this and the fix!

PR #754, opened 2026-05-15
jextract/jni: Support non-JavaBoxable types in Dictionary

sidepelican:
## Problem

Currently, `Dictionary` only supports types that conform to `JavaBoxable`. This limitation prevents the use of dictionaries with custom or generated types, such as `[Int: Fish]`.

## Solution

I have introduced the `JobjectBridge` protocol, which allows us to define custom conversion logic for types when interacting with Java via `jobject`.

- Existing `JavaBoxable` types are handled through a generic `JavaBoxableBridge<T: JavaBoxable>`.
- For types generated by jextract, the generator now automatically creates a dedicated Bridge type form them.

### 1. Type-specific native functions
Using dedicated native functions for each type to avoid `jobject` would be ideal, as it could leverage the existing `translateResult` mechanism and potentially reduce JNI overhead.
However, I couldn't find a viable way to implement this.

### 2. Direct protocol conformance
I considered having types conform directly to `JobjectBridge`.
However, this is problematic for classes with inheritance, as Swift does not allow a type to conform to the same protocol multiple times (which would be necessary at different levels of a class hierarchy). Introducing separate Bridge types effectively bypasses this limitation.


sidepelican:
Note:: Tuples aren't supported in this PR yet.

Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/CollectionBoxable.swift
```diff
@@ -0,0 +1,97 @@
...
+#else
+import Foundation
+#endif
+
+public func makeIntToFishDictionary() -> [Int: Fish] {
+  [
+    1: Fish(name: "salmon"),
+    2: Fish(name: "clownfish"),
+  ]
+}
```
ktoso, line 26: Fantastic, thank you :)

Sources/SwiftJavaRuntimeSupport/JextractedTypeBridge.swift
```diff
@@ -0,0 +1,83 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftJava
+
+public protocol JextractedTypeBridge: JobjectBridge {
```
ktoso, line 21: I like the idea, but since it's pretty tricky, i'd like some docs how this is used here please

Sources/SwiftJavaRuntimeSupport/JextractedTypeBridge.swift
```diff
@@ -0,0 +1,83 @@
...
+      environment,
+      Self.javaClass,
+      Self.wrapMemoryAddressUnsafe,
+      &args
+    )
+  }
+
+  public static func fromJavaObject(_ obj: jobject?, in environment: JNIEnvironment) -> SwiftType {
+    guard let obj else {
+      fatalError("fromJavaObject received a null Java object")
```
ktoso: Maybe include the jclass as we crash here, so we at least know what type it was?

Sources/SwiftJava/BridgedValues/JavaValue+Set.swift
```diff
@@ -17,20 +17,27 @@ import SwiftJavaJNICore
...
     return jlong(Int(bitPattern: rawPointer))
   }
 
   /// Reconstruct a Swift set from a JNI jlong pointer to a SwiftSetBox.
-  public init(fromJNI value: jlong, in environment: JNIEnvironment) {
+  public init<ElementBridge: JobjectBridge>(
+    fromJNI value: jlong,
+    in environment: JNIEnvironment,
+    elementBridge: ElementBridge.Type
+  ) where ElementBridge.SwiftType == Element {
```
ktoso, line 38: I see, this is neat

ktoso:
I'll review this some more later, gotta head offline soon but I like the overall approach, this looks pretty good already 👍

PR #755, opened 2026-05-15
jextract/jni: Fix enums with unconvertible cases

sidepelican:
Currently, defining a generic enum like the one below leads to a Java compilation error:

```swift
public enum GenericEnumWithValue<T> {
  case some(T)
  case none
}
```

```
GenericEnumWithValue.java:86: error: the switch expression does not cover all possible input values
    return switch (this.getDiscriminator()) {
           ^
```

This occurs because `some(T)` contains a unsupported generic parameter and it is skipped.
The Java switch expression becomes non-exhaustive, leading to the build failure.

In cases where a `getCase` method cannot be fully generated due to unsupported generic parameters.
Instead of producing uncompilable code, the generator now provides a dummy implementation instead.


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -504,25 +504,40 @@ extension JNISwift2JavaGenerator {
...
+    // Ensure all cases can be generated to avoid inconsistency with the discriminator.
+    let allCasesCanBeTranslated = decl.cases.allSatisfy({
+      translatedEnumCase(for: $0) != nil
+    })
+    if !allCasesCanBeTranslated {
+      printer.print(
+        """
+        // This is unavailable because it contains unsupported cases.
+        private Case getCase() {
+          return null;
```
ktoso: We should probably  throw an exception instead or just a stray null  like that

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -504,25 +504,40 @@ extension JNISwift2JavaGenerator {
...
-          } else {
-            let arenaArgument = translatedCase.requiresSwiftArena ? "swiftArena" : ""
-            printer.print(
-              "case \(enumCase.name.uppercased()) -> this.getAs\(translatedCase.name)(\(arenaArgument)).orElseThrow();"
-            )
+    // Ensure all cases can be generated to avoid inconsistency with the discriminator.
+    let allCasesCanBeTranslated = decl.cases.allSatisfy({
+      translatedEnumCase(for: $0) != nil
+    })
+    if !allCasesCanBeTranslated {
```
ktoso: I'm not sure I understand -- wouldn't we want to just fail if we're in the not imported case in thew switch below?

sidepelican: Either way works for me. 

I implemented it this way because I thought it would be more user-friendly to make the function unavailable from the start, rather than causing a runtime error.
This is because users can still use alternatives like `getAsNone` instead of `getCase` to work around it.

I'm happy to change it to whichever you prefer.
Which would you prefer: throwing an exception for unsupported cases in the switch, or omitting the function entirely?

ktoso: Hm, if I read this correctly we'd make the entire getCase unavailable if any of the cases is not supported; but perhaps someone wants to use an enum with only a bunch of cases that _were_ supported. So I think the dynamic may be ok... though we should emit warnings about the missed cases and maybe even some comments in source?

ktoso: So yeah, the throw variant: and we'd throw with some message that the case was not extracted because e.g. ... "type ... was not extracted"?

ktoso (approved):

PR #757, opened 2026-05-18
jextract/jni: Support static function calls for specialized types

sidepelican:
For types that were specialized via `typealias`, these static members were not being exported.

```swift
public struct Box<Element>: Hashable {
  ...
  // not exported for FishBox 
  public static func describeElement() -> String {
    String(describing: Element.self)
  }
}

public typealias FishBox = Box<Fish>
```

When all generic parameters are fully specialized, it should be possible to invoke these static members because `Self` can be successfully resolved from the Java side. 
This PR updates the generator to correctly support and export these static function calls.

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
@@ -366,20 +366,23 @@ extension JNISwift2JavaGenerator {
       guard let selfParameter else {
         return nil
       }
-
-      let isGeneric = selfParameter.selfType.asNominalTypeDeclaration?.isGeneric == true
-      if isGeneric {
-        return try self.translateParameter(
-          swiftType: .metatype(selfParameter.selfType),
-          parameterName: "selfTypePointer",
-          methodName: methodName,
-          parentName: parentName,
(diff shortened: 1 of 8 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -366,21 +366,23 @@ extension JNISwift2JavaGenerator {
...
+      switch selfParameter {
+      case .instance:
+        return TranslatedParameter(
+          parameter: JavaParameter(name: "selfTypePointer", type: .long),
+          conversion: .typeMetadataAddress(.placeholder),
+        )
+      case .staticMethod, .initializer:
+        return TranslatedParameter(
+          parameter: JavaParameter(name: "selfTypePointer", type: .long),
+          conversion: .constant("$typeMetadataAddressDowncall()"),
```
sidepelican, line 383: Note: Use `$typeMetadataAddressDowncall()` instead of `this.$typeMetadataAddress()` because static methods cannot use `this`.

ktoso: Yup, thx!

ktoso (approved):

PR #759, opened 2026-05-19
jextract/ffm: Fix TranslatedParameter to hold a single JavaParameter

sidepelican:
This is a minor refactoring to fix the property type of `javaParameters` in the FFM-side `TranslatedParameter`.

Currently, it is defined as an array. However, I believe this is incorrect.
This type represents the Java function interface that the user interacts with, meaning it is not yet "lowered". Therefore, the parameter count should always match the number of Swift arguments.

In practice, there wasn't a place in the codebase where multiple parameters were actually being used.

This PR changes the property to hold a single `JavaParameter`.
This aligns the implementation with the JNI side.

### DIFF
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
@@ -42,7 +42,7 @@ extension FFMSwift2JavaGenerator {
 
   /// Represent a Swift API parameter translated to Java.
   struct TranslatedParameter {
-    /// Java parameter(s) mapped to the Swift parameter.
+    /// Java parameter mapped to the Swift parameter.
     var parameter: JavaParameter
 
     /// Describes how to convert the Java parameter to the lowered arguments for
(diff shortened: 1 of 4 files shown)


ktoso (approved):

Tests/JExtractSwiftTests/DataImportTests.swift
```diff
@@ -471,7 +471,7 @@ final class DataImportTests {
          * public func receiveDataProtocol<T: DataProtocol>(dat: some DataProtocol, dat2: T?)
          * }
          */
-        public static void receiveDataProtocol(Data dat, Optional<Data> dat2) {
+        public static void receiveDataProtocol(Data dat, java.util.Optional<Data> dat2) {
```
ktoso, line 474: nice side effect, FQN are better

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -42,10 +42,8 @@ extension FFMSwift2JavaGenerator {
 
   /// Represent a Swift API parameter translated to Java.
   struct TranslatedParameter {
-    /// Java parameter(s) mapped to the Swift parameter.
-    ///
-    /// Array because one Swift parameter can be mapped to multiple parameters.
-    var javaParameters: [JavaParameter]
+    /// Java parameter mapped to the Swift parameter.
+    var parameter: JavaParameter
```
ktoso, line 46: 👍 Yeah I think that's right, lowering would explode it into multiple if necessary

PR #760, opened 2026-05-20
Detect double destruction and generalize $createDestroyFunction

sidepelican:
This PR introduces two improvements to the destruction process of Swift instances.

### 1. Detect double destruction

Currently, if a double destruction occurs due to an implementation error, the process simply crashes without any diagnostics.
To throw an error, this PR passes the `statusDestroyedFlag` directly to the cleanup instance instead of `markAsDestroyed`.

### 2. Generalize `$createDestroyFunction`

Previously, a unique `destroy` function was generated for each individual class.
However, we have transitioned to a shared invocation of `SwiftObjects.destroy` (in #582 ).
Since there is no longer a reason to generate a separate `$createDestroyFunction` implementation for every single class, this PR provides a default implementation to minimize the footprint of the generated code.

Note: Unlike the FFM side, `$createDestroyFunction` is still required here because certain classes still implement their own custom `destroy` logic.

### DIFF
--- a/SwiftKitFFM/src/main/java/org/swift/swiftkit/ffm/FFMSwiftInstanceCleanup.java
+++ b/SwiftKitFFM/src/main/java/org/swift/swiftkit/ffm/FFMSwiftInstanceCleanup.java
@@ -19,26 +19,29 @@ import org.swift.swiftkit.core.SwiftInstanceCleanup;
 import static org.swift.swiftkit.ffm.SwiftJavaLogGroup.LIFECYCLE;
 
 import java.lang.foreign.MemorySegment;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class FFMSwiftInstanceCleanup implements SwiftInstanceCleanup {
     private final MemorySegment memoryAddress;
     private final SwiftAnyType type;
-    private final Runnable markAsDestroyed;
+    private final AtomicBoolean statusDestroyedFlag;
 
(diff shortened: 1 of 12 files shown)


sidepelican:
Gradle seems unhealthy again today 🤒

ktoso:
Yeah that's what we get for running 60 jobs... I'll work on getting less jobs to run soon; but we'll have to cut coverage of swift versions etc. At least less flaky fails then

SwiftKitFFM/src/main/java/org/swift/swiftkit/ffm/FFMSwiftInstanceCleanup.java
```diff
@@ -19,26 +19,29 @@
 import static org.swift.swiftkit.ffm.SwiftJavaLogGroup.LIFECYCLE;
 
 import java.lang.foreign.MemorySegment;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class FFMSwiftInstanceCleanup implements SwiftInstanceCleanup {
     private final MemorySegment memoryAddress;
     private final SwiftAnyType type;
-    private final Runnable markAsDestroyed;
+    private final AtomicBoolean statusDestroyedFlag;
```
ktoso, line 27: We could move to atomic field updater here, but I'm happy to do this separately

ktoso (approved):

PR #761, opened 2026-05-20
Rework the cleanup a little bit so that we avoid the atomic boolean 

ktoso:
The atomic boolean was one extra heap allocation which we now avoid --
we just create one cleanup and inside it have an int, rather than
creating the Bool separately and passing it to the cleanup. AFAICS this
will avoid one alloc, which isn't much, but can add up since we do it
for every wrapped swift object.

This keeps semantics added by @sidepelican in https://github.com/swiftlang/swift-java/pull/760 just avoids the one heap alloc per SwiftInstance

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
@@ -27,7 +27,6 @@ extension JNISwift2JavaGenerator {
     "org.swift.swiftkit.core.util.*",
     "org.swift.swiftkit.core.collections.*",
     "java.util.*",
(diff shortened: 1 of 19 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -317,16 +317,16 @@ extension JNISwift2JavaGenerator {
         /** Pointer to the "self". */
         private final long selfPointer;
 
-        /** Used to track additional state of the underlying object, e.g. if it was explicitly destroyed. */
-        private final AtomicBoolean $state$destroyed = new AtomicBoolean(false);
```
ktoso, line 321: The alloc we're removing; instead we have the Cleanup as we did before; alloc it early, but then inside the Cleanup we have no other heap objects -- just the int.

PR #765, opened 2026-05-23
jextract/jni: Propagate generic parameters to Case types of generic enums

sidepelican:
This PR is the first step toward supporting the extraction of generic associated values from generic enum.

Currently, the generated `Case` types for generic enums do not carry generic parameters.
While this isn't causing immediate issues, these parameters will be necessary in the future when we want to extract generic associated values from the `Case`.

```java
public sealed interface Case {
  // `Wrapped` is not usable from the static context
  record Some(Wrapped arg0) implements Case {}
  record None() implements Case {}
}
```

This PR focuses solely on propagating the generic parameters of the enum type down to its `Case` type. 

During this work, I encountered and fixed an issue in wrap-java where translation was failing when a nested type's parent type was generic.

---

## Note

I have loved contributing to `swift-java` for a half year, but unfortunately, I have run out of time to continue doing so.
While I will keep following the development of `swift-java`, I won't be able to actively contribute as I did before. 

Thank you so much for all your valuable code reviews.

### DIFF
--- a/Tests/SwiftJavaToolLibTests/JavaTranslatorTests.swift
+++ b/Tests/SwiftJavaToolLibTests/JavaTranslatorTests.swift
@@ -62,11 +62,42 @@ class JavaTranslatorTests: XCTestCase {
       """
       package com.example;
 
+      class MyString {}
+      class MyInt {}
+
+      class Outer<T> {
+        class Inner<U> {
+        }
+
+        Outer<MyString>.Inner<MyInt> foo() { return null; }
+        Outer<MyInt>.Inner<T> bar() { return null; }
+      }
+      """
+    )
+
+    try assertWrapJavaOutput(
+      javaClassNames: [
+        "com.example.MyString",
+        "com.example.MyInt",
+        "com.example.Outer$Inner",
+        "com.example.Outer",
+      ],
+      classpath: [classpathURL],
+      expectedChunks: [
+        "func foo() -> Outer<MyString>.Inner<MyInt>!",
+        "func bar() -> Outer<MyInt>.Inner<T>!",
+      ]
+    )
+  }
+
+  func testTranslateNestedParameterizedStaticTypes() async throws {
+    let classpathURL = try await compileJava(
+      """
+      package com.example;
+
       class MyOptional<Wrapped> {
         interface Case<Wrapped> {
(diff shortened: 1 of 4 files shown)


Tests/SwiftJavaToolLibTests/JavaTranslatorTests.swift
```diff
@@ -56,4 +56,40 @@ class JavaTranslatorTests: XCTestCase {
...
+        "com.example.MyOptional",
+      ],
+      classNameMappings: [
+        "com.example.MyOptional$Case": "MyOptional.Case",
+        "com.example.MyOptional$Case$None": "MyOptional.Case.None",
+      ],
+      classpath: [classpathURL],
+      expectedChunks: [
+        "func getCase() -> MyOptional.Case<Wrapped>!",
+        "func getAsNone() -> MyOptional.Case<Wrapped>.None<Wrapped>!",
```
sidepelican: On second thought, this might be incorrect. 
The generated Java name is `com.example.MyOptional$Case$None<Wrapped>`, meaning there are no generic parameters in between.
It feels like it should actually be `MyOptional<JavaObject>.Case<JavaObject>.None<Wrapped>` instead.

ktoso:
Oh too bad to hear you won’t have more time contributing … maybe in some hobby time :-)


Thanks for all the contributions, it’s all accounted for very useful improvements!

sidepelican:
It looked like the tests were hanging, but the root cause was that I had modified the name of the test below with good intentions. 
This test didn't start with `test~`, meaning it wasn't being picked up by XCTest.

https://github.com/swiftlang/swift-java/blob/060b647b45ee545e664daedfe8b092f24d922748/Tests/SwiftJavaToolLibTests/JavaTranslatorTests.swift#L25

As it turns out, running this test recreates the JVM internally.
However, if other test cases have already been executed and a JVM instance already exists, it somehow deadlocks.

Since this deadlock is an unrelated issue to this PR, I am reverting the test name back to its original state.

ktoso:
Thanks, looks good. We have a source break due to argument-parser source breaking, I'm fixing the build now and will re-run this https://github.com/swiftlang/swift-java/pull/767

ktoso (approved):

PR #767, opened 2026-05-26
Adjust for source-break in argument-parser 1.8

ktoso:
async-argument-parser just tagged 1.8 which breaks consumers by changing a method to async: https://github.com/apple/swift-argument-parser/pull/855

We have to call it async now.


marcprux:
This is causing the https://github.com/swiftlang/swift-android-examples to break, e.g. with [CI errors](https://github.com/swiftlang/swift-android-examples/actions/runs/26433749093/job/77811996095): 

```
/Users/runner/work/swift-android-examples/swift-android-examples/hello-swift-java/hashing-lib/.build/checkouts/swift-java/Sources/SwiftJavaTool/SwiftJava.swift:46:21: error: expression is 'async' but is not marked with 'await'
 44 |   public static func main() async {
 45 |     do {
 46 |       var command = try parseAsRoot(nil)
    |                     |   `- note: call is 'async'
    |                     `- error: expression is 'async' but is not marked with 'await'
 47 |       if var asyncCommand = command as? AsyncParsableCommand {
 48 |         try await asyncCommand.run()
[793/802] Compiling swift_java SwiftJava.swift
```

~~Do we need a new swift-java patch release?~~

NVM: I see you released https://github.com/swiftlang/swift-java/releases/tag/0.4.0

PR #768, opened 2026-05-26
CI: Optimize builds and cancel in flight builds

ktoso:
We have too many builds now, because we test every sample and on many platforms and swift versions. This is too extreme since they don't really differ much.

- macOS only tests 6.3 Swift now
- total saved 32 jobs (53 matrix-driven jobs -> 21)

We want to cancel in progress jobs since we dont really need them; just always run on the latest pushed code immediately, freeing up resources more eagerly

We also improve build caching for the samples


ktoso:
Trying to improve caching to speed up the samples builds as well

PR #773, opened 2026-05-27
jextract/jni: Support protocol inheritance

sidepelican:
Currently, when an inheritance relationship exists between protocols, the generated Java class only expresses conformance to the child protocol.
The parent protocol relationship is lost.

This PR updates the generator to properly represent protocol inheritance within the generated Java interfaces.


ktoso (approved):
Fantastic fix, thanks you

PR #774, opened 2026-05-27
Initial SwiftExtract version - a reusable Swift analysis module

ktoso:
This is an initial PR to extract a lot of the analysis code into an reusable `SwiftExtract` module. The module has no source compatibility guarantees at this point. The intent is to facilitate other language source generators that can reuse the same analysis core which is "the tricky bit".

This fully rebases swift-java onto this shared infrastructure, without loss of features; and even gains a few along the way, like being better prepared for lazy specialization with typealiases.

This is a massive "move stuff around" so reviewing everyting might be hard but the long term benefit will be massive.


ktoso:
I deeply reviewed all the files and validated against a few consumers, I think this is good enough first stab at the split.

PR #778, opened 2026-06-02
jextract/jni: Implement generic equals method

sidepelican:
In Java, every `Object` has an `equals` method.
Conceptually, Swift's `Equatable` protocol serves the exact same purpose as Java's `equals`. 

This PR implements a generic `equals` method, enabling all `JNISwiftInstance` types to leverage Swift's `Equatable` comparison directly through the standard Java `equals` method.


ktoso (commented):
Neat! I wonder if we can nicely do hash code as well; otherwise we're breaking the contract of always having to implement "both"

sidepelican:
I forgot to consider hashCode.
It's a bit challenging (Only Equatable types cannot provide hashCode)

sidepelican:
Implemented `hashCode()`. 
I made a compromise regarding types that only implement Equatable.

ktoso:
Looks good, let’s add a test putting swift values into a java set maybe or
Something like that?

PR #781, opened 2026-06-08
Add regression coverage for overloaded initializer Java names

### LINKED ISSUE #425: [jextract] Handle ambiguous Swift methods not possible in Java
If you try to wrap these functions
```swift
public init(throwing: Bool) throws {
  // ...  
}

public init?(doInit: Bool) {
  // ...
}
```

the current resulting Java functions would be:

```java
public static MySwiftClass init(boolean throwing, SwiftArena swiftArena$) throws Exception
public static Optional<MySwiftClass> init(boolean doInit, SwiftArena swiftArena$)
```
which is not allowed in Java, since `'init(boolean, SwiftArena)` is defined twice.

We should handle this, maybe adding the first parameterName as a suffix, or something else.

tientnc:
Adds regression coverage for #425.
This covers the reported initializer-collision where distinct Swift initializers could otherwise produce duplicate Java `init(...)` signatures. The added tests assert that the generated FFM/JNI Java names are disambiguated, e.g. `initThrowing` and `initDoInit`, and that the raw duplicate `init(...)` signatures are absent.

Closes #425.

Test:
- `swift test --filter MethodImportTests`

### DIFF
--- a/Tests/JExtractSwiftTests/MethodImportTests.swift
+++ b/Tests/JExtractSwiftTests/MethodImportTests.swift
@@ -668,6 +668,44 @@ final class MethodImportTests {
     )
   }
 
+  let overloadedInitializers_interfaceFile =
+    """
+    import Swift
+
+    public class OverloadedInitializerClass {
+      public init(throwing: Swift.Bool) throws
+      public init?(doInit: Swift.Bool)
+    }
+    """
+
+  let overloadedFFMInitializers_interfaceFile =
+    """
+    import Swift
+
+    public class OverloadedFFMInitializerClass {
+      public init(throwing: Swift.Bool) throws
+      public init(doInit: Swift.Bool)
+    }
+    """
+
+  @Test("FFM: Overloaded initializers with same Java signature get suffixed Java names")
+  func ffm_overloaded_initializers_suffixed() throws {
+    try assertOutput(
+      input: overloadedFFMInitializers_interfaceFile,
+      .ffm,
+      .java,
+      swiftModuleName: "OverloadModule",
+      expectedChunks: [
+        "public static OverloadedFFMInitializerClass initThrowing(boolean throwing, AllocatingSwiftArena swiftArena)",
+        "public static OverloadedFFMInitializerClass initDoInit(boolean doInit, AllocatingSwiftArena swiftArena)",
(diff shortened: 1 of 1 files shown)


Tests/JExtractSwiftTests/MethodImportTests.swift
```diff
@@ -668,6 +668,44 @@ final class MethodImportTests {
...
+
+  let overloadedFFMInitializers_interfaceFile =
+    """
+    import Swift
+
+    public class OverloadedFFMInitializerClass {
+      public init(throwing: Swift.Bool) throws
+      public init(doInit: Swift.Bool)
+    }
+    """
```
ktoso: No real reason for two separate input strings

Tests/JExtractSwiftTests/MethodImportTests.swift
```diff
@@ -776,4 +814,22 @@ final class MethodImportTests {
...
+      swiftModuleName: "OverloadModule",
+      expectedChunks: [
+        "public static OverloadedInitializerClass initThrowing(boolean throwing, SwiftArena swiftArena) throws Exception",
+        "public static java.util.Optional<OverloadedInitializerClass> initDoInit(boolean doInit, SwiftArena swiftArena)",
+      ],
+      notExpectedChunks: [
+        "public static OverloadedInitializerClass init(boolean throwing, SwiftArena swiftArena)",
+        "public static java.util.Optional<OverloadedInitializerClass> init(boolean doInit, SwiftArena swiftArena)",
+      ]
+    )
```
ktoso: Make it one test that is parameterized over the two modes please

ktoso (approved):

Tests/JExtractSwiftTests/MethodImportTests.swift
tientnc: Yub, I have consolidated into 1 shared input.

Tests/JExtractSwiftTests/MethodImportTests.swift
tientnc: Updated, thanks! I merged into 1 parameterized test over JNI/FFM.

ktoso:
Thanks :)

PR #782, opened 2026-06-09
[WIP] Add java.net.HttpURLConnection to JavaNet

marcprux:
[`java.net.HttpURLConnection`](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html) is very useful because it provides access to the Android-native HTTP connection handling APIs.

Leaving as draft for discussion.

### DIFF
--- a/Sources/JavaStdlib/JavaNet/swift-java.config
+++ b/Sources/JavaStdlib/JavaNet/swift-java.config
@@ -3,5 +3,6 @@
     "java.net.URI" : "URI",
     "java.net.URLClassLoader" : "URLClassLoader",
     "java.net.URLConnection" : "URLConnection",
+    "java.net.HttpURLConnection" : "HttpURLConnection",
   }
 }


ktoso:
Yeah that’s a good one, agreed.

are we missing much of its api surface though or did it import all we need already? (What other types do we need to add)

marcprux:
> are we missing much of its api surface though or did it import all we need already? (What other types do we need to add)

I think we have almost everything we need. The Java 8 docs for [HttpURLConnection](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html) only reference `InputStream`, which I think it already there. The only other one is [`java.security.Permission`](https://docs.oracle.com/javase/8/docs/api/java/security/Permission.html), which doesn't do anything on Android (or anywhere else we are likely to be targeting, I expect).

ktoso:
Yeah `Permissions` isn't too exciting I guess AFAIK? We're missing `getHeaderFields` which might be a nice one to have, I'll follow up on it here.

ktoso:
Meh, this reminds me we should make wrap-java more repeatable, it causes a huge diff now due to unordered member rendering in generated sources... I'll get this landed, thanks for the idea @marcprux !

ktoso:
I'll merge this and follow up with regeneration and some CI changes

PR #785, opened 2026-06-10
Various attempts to speed up CI

ktoso:
- more caching
- move to noble
- remove --disable-experimental-prebuilts where we can


ktoso:
Rebuilds are between 30-50% faster with this aggressive caching, so let's keep it.

PR #787, opened 2026-06-14
Fix: Identity equality for non `Equatable` clases.

madsodgaard:
Use identity for `equals` methods for types that do not implement `Equatable`, for example classes:

**Swift:**
```swift
final class MySwiftClass {}
```

**Java:**
```java
// Extracted by jextract
public final class MySwiftClass {}

MySwiftClass c1 = MySwiftClass.init();
c1.equals(c1) // would return false before
```

### DIFF
--- a/Sources/SwiftJava/SwiftObjects.swift
+++ b/Sources/SwiftJava/SwiftObjects.swift
@@ -102,7 +102,7 @@ extension SwiftObjects {
   @JavaMethod
   public static func equals(environment: UnsafeMutablePointer<JNIEnv?>!, lhsPointer: Int64, lhsTypePointer: Int64, rhsPointer: Int64, rhsTypePointer: Int64) -> Bool {
     // Fallback for non-equatable types (such as classes)
-    let isEquatableByIdentity = lhsPointer == rhsPointer && lhsTypePointer == rhsTypePointer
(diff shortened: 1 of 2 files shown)


Sources/SwiftJava/SwiftObjects.swift
```diff
@@ -101,20 +101,23 @@ extension SwiftObjects {
 
   @JavaMethod
   public static func equals(environment: UnsafeMutablePointer<JNIEnv?>!, lhsPointer: Int64, lhsTypePointer: Int64, rhsPointer: Int64, rhsTypePointer: Int64) -> Bool {
+    // Fallback for non-equatable types (such as classes)
+    let isEquatableByIdentity = lhsPointer == rhsPointer && lhsTypePointer == rhsTypePointer
```
ktoso: A test would be nice to add for these

ktoso: Also naming nitpick, this is  "is equal" and not "is equatable" we're not checking if the types are equatABLE but if they _are_ :)

ktoso:
I did the tiny followup and will merge, thank you!

ktoso:
Nightly Swift seems broken for unrelated reasons rdar://179674933

PR #789, opened 2026-06-16
fail eagerly when missing JavaObject mapping

### LINKED ISSUE #195: If for some reason we're missing java.lang.Object mapping during import, fail eagerly
Due to dependencies behaving weirdly in plugins the JavaKit dependency was not pulled unless more things were spelled out explicitly...

This resulted in a generated file like:

```
@JavaClass("org.apache.commons.io.FilenameUtils")
open class FilenameUtils {
  @JavaMethod
  @_nonoverride public convenience init(environment: JNIEnvironment? = nil)
}
```

which will fail to compile because we need the:

```
@JavaClass("org.apache.commons.io.FilenameUtils")
open class FilenameUtils: JavaObject {
  @JavaMethod
  @_nonoverride public convenience init(environment: JNIEnvironment? = nil)
}
```

form. 

If we're ever generating code and we're missing `JavaObject` we should just immediately crash rather than hide the error futher along the build as it can be hard to diagnose what actually went wrong -- but it was a missing dependency on the swift-java file that declared java.lang.object.

tientnc:
When translating a non-root Java class, missing the `java.lang.Object -> JavaObject` mapping could allow generation to continue and produce invalid Swift (without the required JavaObject inheritance).

This now fails eagerly when a Java superclass exists but the JavaObject mapping is unavailable. The regression test clears the mapping while translating `java.lang.String` as a Swift class and asserts the missing `java.lang.Object` error.

Test:
- `swift test --filter testMissingJavaObjectMappingFailsEagerly`
- `swift test --filter Java2SwiftTests`
- `swift-format lint --configuration .swift-format Sources/SwiftJavaToolLib/JavaClassTranslator.swift Tests/SwiftJavaToolLibTests/Java2SwiftTests.swift`

Fixes #195.

### DIFF
--- a/Sources/SwiftJavaToolLib/JavaClassTranslator.swift
+++ b/Sources/SwiftJavaToolLib/JavaClassTranslator.swift
@@ -137,6 +137,10 @@ struct JavaClassTranslator {
     // Superclass, incl parameter types (if any)
     if !javaClass.isInterface() {
       var javaSuperclass = javaClass.getSuperclass()
+      if javaSuperclass != nil, translator.translatedClasses[JavaObject.fullJavaClassName] == nil {
+        throw TranslationError.untranslatedJavaClass(JavaObject.fullJavaClassName)
+      }
+
       var javaGenericSuperclass: Type? = javaClass.getGenericSuperclass()
       var swiftSuperclassName: String? = nil
       var swiftSuperclassTypeArgs: [String]? = nil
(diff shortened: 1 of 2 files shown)


ktoso:
This looks good, thank you!

tientnc:
Thanks. Regarding the failed `swift:nightly` check, it seems like a compiler crash. The nightly Swift compiler failed on an internal SIL verification pass while building 1 of the package dependencies.

ktoso:
Yes, nightly is broken -- not your fault :)

PR #792, opened 2026-06-18
jextract: Skip generating specialization for generic enum types

sidepelican:
Currently, specialization for generic enum types is not working correctly, leading to compilation errors in the generated code.

```
.../IntGenericEnum.java:82: error: cannot find symbol
  public Case<T> getCase() {
              ^
```

```
.../FlapjackValues+SwiftJava.swift:1461:103: error: conflicting conformance of 'APIResponseResult<Success, Failure>' to protocol '_RawDiscriminatorRepresentable'; there cannot be more than one conformance, even with different conditional bounds
1390 | } // printOpenerProtocol(_:_:) @ JExtractSwiftLib/JNISwift2JavaGenerator+SwiftThunkPrinting.swift:1053
1391 | 
1392 | extension APIResponseResult: _RawDiscriminatorRepresentable {
     | `- note: 'APIResponseResult<Success, Failure>' declares conformance to protocol '_RawDiscriminatorRepresentable' here
1393 |   public var _rawDiscriminator: Int32 {
1394 |     switch self {
     :
1459 | 
```

As additional implementation is required to fully support generic enum specialization, this PR temporarily skips the generation of these specializations to prevent the aforementioned build errors.

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/GenericType.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/GenericType.swift
@@ -88,7 +88,7 @@ public func makeIntGenericEnum() -> GenericEnum<Int> {
   if Bool.random() { return .foo } else { return .bar }
 }
 
-public typealias IntGenericEnum = GenericEnum<Int>
+public typealias IntGenericEnum = GenericEnum<Int> // This specialization is not supported yet.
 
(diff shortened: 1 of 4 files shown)


ktoso (approved):
Sounds good but we should immediately file an issue about it so we don't forget to fix them

ktoso:
I've filed an issue here https://github.com/swiftlang/swift-java/issues/793

PR #800, opened 2026-06-20
Use suppressed exceptions to provide detailed library loading errors

sidepelican:
Previously, when `System.loadLibrary` failed, the specific underlying errors were not being exposed. 

For instance, `libSwiftJava` depends on `libFoundationEssentials`, which in turn depends on `libswiftSynchronization.so`. If `libswiftSynchronization` is missing from the runtime environment, the error message would simply report a generic `SwiftJava` loading failure. This is confusing.

```
Caused by: java.lang.RuntimeException: Failed to load SwiftJava from java.library.path and JAR resources
```

This PR updates the loading logic to collect and expose all encountered errors by utilizing `Throwable.addSuppressed()`.

```
Caused by: java.lang.RuntimeException: Failed to load 'SwiftJava'. See suppressed exceptions for details.
...
Suppressed: java.lang.UnsatisfiedLinkError: dlopen failed: library "libswiftSynchronization.so" not found: needed by /data/app/~~4oHrmhVZEl3rBmUeZO9Qxw==/com.example.flapjack.sample-oJLD0uYeGSfA89xh9utNJA==/base.apk!/lib/arm64-v8a/libFoundationEssentials.so in namespace clns-7
...
Suppressed: java.lang.RuntimeException: Expected library 'SwiftJava' ('libSwiftJava.so') was not found as resource!
...
```

### DIFF
--- a/SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftLibraries.java
+++ b/SwiftKitCore/src/main/java/org/swift/swiftkit/core/SwiftLibraries.java
@@ -30,13 +30,13 @@ public final class SwiftLibraries {
     public static final String LIB_NAME_SWIFT_RUNTIME_FUNCTIONS = "SwiftRuntimeFunctions";
     public static final String LIB_NAME_SWIFT_JAVA = "SwiftJava";
 
-    /** 
+    /**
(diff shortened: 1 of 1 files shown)


ktoso (approved):

ktoso:
Good fixup, thank you

PR #813, opened 2026-07-02
jextract/jni: Support escaping closures through @JavaInterface callbacks

### LINKED ISSUE #328: Add support for escaping closures
Right now, the JNI and FFM modes only support importing non-escaping closures. We should investigate and add support for escaping closures as well, as they are quite common in Swift, such as callbacks. We need to figure out a way to know when to "release" the closure, for example using some sort of wrapper object.

ktoso:
This is safe because we keep the java side object alive by capturing the JavaObjectHolder to it un the swift closure; so as long as Swift retains that holder, the Java object remains alive as well -- we cannot call on a freed java object.

Reuses the existing `JavaInterfaceProtocolWrapperGenerator` without going through the whole wrap-java route, since we know both sides and what to generate for it. This means that this works even with the sandbox enabled.

The java side just get a functional interface they can invoke, and rather than upcalling into it directly, we use the java interface Swift wrapper to make the upcall.

resolves https://github.com/swiftlang/swift-java/issues/328

cc @madsodgaard


ktoso:
Meh, still some things I'm not happy with that I missed, marking as draft

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -194,11 +194,24 @@ extension JNISwift2JavaGenerator {
...
+          // printer can emit the `@JavaInterface` Swift struct alongside
+          // the cdecl thunk that references it - no generator-level
+          // sink/dict/pre-pass required. `NativeParameter` runs 1:1 with
+          // the Swift parameters here (a Swift param can lower to multiple
+          // *JavaParameter*s but stays one `NativeParameter`), so indexing
+          // by `idx` is correct.
+          if idx < nativeFunctionSignature.parameters.count {
+            translatedClosure.syntheticClosureType =
+              nativeFunctionSignature.parameters[idx].syntheticClosure
+          }
```
ktoso, line 214: Forgot about this bit; I'll change how this works -- don't want such random indexing into things

ktoso: On second throught, I'm ok with this because the signatures must align but names are not necessarily ideal to index on here

PR #814, opened 2026-07-02
jextract/jni: Add support for returning protocols

### LINKED ISSUE #380: [jextract][jni] Add support for returning protocol types
#376 added support for passing in existentials, opaque types and generics as parameters. It did not cover returning them, so we need to do that separately.

madsodgaard:
Adds support for returning `any P / some P` from a function, variable etc...

This is done by generating a `<P>Box` instance in Java, that makes downcalls by opening the existential and calling a cdecl. This box implements both all protocol requirements and any inherited protocol requirements.

One special thing is for `setters`, where we need to do a "load-mutate-store", because it might be a value type.

Closes #380


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -187,26 +205,112 @@ extension JNISwift2JavaGenerator {
...
+  }
 
-      for variable in decl.variables {
-        if variable.isStatic {
-          self.logger.debug("Skipping static property '\(variable.name)'")
-          continue
+  /// Prints the full `.java` file for a protocol's existential box — the
+  /// class used to represent a value returned as `any P` / `some P` from an
+  /// extracted Swift function.
+  private func printExistentialBoxFile(_ printer: inout JavaPrinter, _ decl: ExtractedNominalType) {
```
ktoso, line 222: I'd like an assertion here that it is a protocol maybe... we call it right but nice to assert here

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaBindingsPrinting.swift
```diff
@@ -372,32 +435,113 @@ extension JNISwift2JavaGenerator {
...
-      printer.println()
     }
   }
 
+  /// Prints the `wrapMemoryAddressUnsafe` factory pair shared by every `JNISwiftInstance`
+  /// (both existential boxes and concrete imported types). The doc summary sentence differs
+  /// between callers (a box also carries type metadata), so it is passed in verbatim.
+  private func printWrapMemoryAddressUnsafeFactory(
+    _ printer: inout JavaPrinter,
+    javaName: String,
```
ktoso: Nitpick

```
-    javaName: String,
+    javaName: JavaClassName,
```

ktoso (approved):
Thanks for cleaning up the comments a bit; logic looks good, I'll want to do some restructuring since we're a bit repetetive sometimes. 

LGTM tho

PR #815, opened 2026-07-05
Import FoundationEssentials if available

tkrajacic:
We really don't want to drag in the full Foundation on non-Apple platforms.


ktoso (approved):

ktoso:
https://github.com/swiftlang/swift-java/actions/runs/28748471718/job/85283038977?pr=815 should be catching those issues, ill follow up with improving that.

Issue #811, opened 2026-07-01
jextract closure support should not specialize () -> () and () -> T shapes, and just use Runnable and Supplier


ktoso:
That's idiomatic Java types to use for these situations and we can save ourselves some codegen by doing so.

So when we see

```
func test(a: () -> ()) 
```

We should make that `Runnable` rather than generate an one-off type for this.

The same for Supplier probably

AmrDeveloper:
I would like to work on it as the first issue in swift-java 😋

ktoso:
Not closing this since we should handle all the other known functional interfaces but we only did Runnable now

PR #820, opened 2026-07-07
jextract: Translate Swift () -> () to java Runnable

AmrDeveloper:
Translate the Swift function parameter type () -> () to the Java Runnable interface

Issue: #811

### DIFF
--- a/Samples/SwiftJavaExtractFFMSampleApp/src/test/java/com/example/swift/MySwiftLibraryTest.java
+++ b/Samples/SwiftJavaExtractFFMSampleApp/src/test/java/com/example/swift/MySwiftLibraryTest.java
@@ -81,9 +81,9 @@ public class MySwiftLibraryTest {
     void call_globalCallMeRunnable() {
         CountDownLatch countDownLatch = new CountDownLatch(3);
 
-        MySwiftLibrary.globalCallMeRunnable(new MySwiftLibrary.globalCallMeRunnable.run() {
+        MySwiftLibrary.globalCallMeRunnable(new java.lang.Runnable() {
             @Override
-            public void apply() {
+            public void run() {
                 countDownLatch.countDown();
             }
         });
--- a/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
+++ b/Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
@@ -566,6 +566,17 @@ extension FFMSwift2JavaGenerator {
       )
     }
 
+    // Extract Java known functional interface type or null if not exists
+    func extractKnownJavaFunctionalInterfaceType(
+      functionType: SwiftFunctionType,
+    ) -> JavaType? {
+      if functionType.parameters.isEmpty && functionType.resultType.isVoid {
+        JavaType.javaLangRunnable
+      } else {
+        nil
+      }
+    }
+
     /// Translate a Swift Function parameter to the Java interface.
     func translateFunctionParameter(
       functionType: SwiftFunctionType,
@@ -573,11 +584,7 @@ extension FFMSwift2JavaGenerator {
       methodName: String,
     ) -> TranslatedParameter {
       let parameterType =
-        if functionType.parameters.count == 0 && functionType.resultType.isVoid {
-          JavaType.javaLangRunnable
-        } else {
-          JavaType.class(package: nil, name: "\(methodName).\(parameterName)")
-        }
+        extractKnownJavaFunctionalInterfaceType(functionType: functionType) ?? JavaType.class(package: nil, name: "\(methodName).\(parameterName)")
 
       return TranslatedParameter(
         parameter: JavaParameter(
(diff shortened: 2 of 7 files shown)


ktoso:
So this only does the FFM mode, could you also please do the JNI mode?

Then also please make sure we have a runtime test -- i.e. a test in `Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary` and `Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift` that covers this so we didn't regress anythign and run end to end?

Sources/JExtractSwiftLib/FFM/FFMSwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -568,6 +566,28 @@ extension FFMSwift2JavaGenerator {
...
     }
 
+    /// Translate a Swift Function parameter to the Java interface.
+    func translateFunctionParameter(
+      functionType: SwiftFunctionType,
+      parameterName: String,
+      methodName: String,
+    ) -> TranslatedParameter {
+      let parameterType =
+        if functionType.parameters.count == 0 && functionType.resultType.isVoid {
```
ktoso: Can you extract this into some `extractKnownJavaFunctionalInterfaceType(functionType)`

it should return a `JavaType?` and if we get nil there we do the below branch.

There we'll also handle the other known types then, like supplier etc

ktoso:
Yeah samples broke:

```

/__w/swift-java/swift-java/Samples/SwiftAndJavaJarFFMSampleLib/.build/plugins/outputs/swiftandjavajarffmsamplelib/MySwiftLibrary/destination/JExtractSwiftPlugin/src/generated/java/com/example/swift/MySwiftLibrary.java:240: error: incompatible types: Runnable cannot be converted to run
      swiftjava_MySwiftLibrary_globalCallMeRunnable_run.call(globalCallMeRunnable.$toUpcallStub(run, arena$));
                                                                                                ^
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error

```

You can run them by 
```

  cd Samples/SwiftAndJavaJarFFMSampleLib
  ./ci-validate.sh
```

AmrDeveloper:
> So this only does the FFM mode, could you also please do the JNI mode?
> 
> Then also please make sure we have a runtime test -- i.e. a test in `Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary` and `Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift` that covers this so we didn't regress anythign and run end to end?

I think we already have a test for Closures

https://github.com/swiftlang/swift-java/blob/c235a0f71f46ff6aed766cb939676dc498116799/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Closures.swift#L15

https://github.com/swiftlang/swift-java/blob/main/Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/ClosuresTest.java

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1644,10 +1689,17 @@ extension JNISwift2JavaGenerator {
           $0.conversion.render(&printer, $0.parameters.first!.name)
         }
 
+        let methodName =
```
AmrDeveloper, line 1692: I am still thinking of having a map to hold all names for functional interfaces 🤔

Samples/SwiftJavaExtractFFMSampleApp/src/test/java/com/example/swift/MySwiftLibraryTest.java
```diff
@@ -81,9 +81,9 @@ void call_globalStringIdentity_empty() {
     void call_globalCallMeRunnable() {
         CountDownLatch countDownLatch = new CountDownLatch(3);
 
-        MySwiftLibrary.globalCallMeRunnable(new MySwiftLibrary.globalCallMeRunnable.run() {
+        MySwiftLibrary.globalCallMeRunnable(new java.lang.Runnable() {
```
ktoso, line 84: Good Yeap!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
ktoso: Let's make a file with KnownFunctionalInterfaces.swift and cases on there; they should all have `case ...(JavaType, method: ...)` since they all have a single method and then we can take the name off off that

ktoso: I'd really welcome a follow up PR fixing this up so it's less random method names in here like that

AmrDeveloper: Done in #844.

After that, I will create another follow-up to eliminate generating unnecessary interface definitions when we use a known interface.

PR #823, opened 2026-07-09
Improve nested names handling even in #if scopes

### LINKED ISSUE #822: jextract issues with nested types that share the same name
my environment:
swift-java: main @ c235a0f71f46ff6aed766cb939676dc498116799
swift-java-jni-core: main @ b440a0be98381550cf189c903782e817619ffa05
using swift 6.3.3 and im using the jni mode.

my module contains several nested types which share the same name. [example 1](https://github.com/llsc12/Paicord/blob/05dcda6cb537983f1e412b882215a1a820ba5393/PaicordLib/Sources/DiscordModels/Types/Channel.swift#L15) [example2](https://github.com/llsc12/Paicord/blob/05dcda6cb537983f1e412b882215a1a820ba5393/PaicordLib/Sources/DiscordModels/Types/Channel.swift#L322)

reproduced with this snippet.
```swift
public struct Outer: Sendable, Codable {
  public enum Kind: Sendable, Codable {
    case alpha
    case beta
  }
}

extension Outer {
  public struct Inner: Sendable, Codable {
    public struct Innermost: Sendable, Codable {
      public enum Kind: Sendable, Codable {
        case x
        case y
      }

      public var kind: Kind // Outer.Inner.Innermost.Kind

      public init(kind: Kind) {
        self.kind = kind
      }
    }

    public var innermost: Innermost

    public init(innermost: Innermost) {
      self.innermost = innermost
    }
  }
}
```

the jextract plugin doesn't produce fully qualified types. in my own package, it had issues with `DiscordChannel.Message.MessageReference.Kind` which became `DiscordChannel.Kind`. in the above snippet its like `Outer.Inner.Kind` becoming `Outer.Kind`. i think its specifically for extensions.

the repo i'm working on is available here:
https://github.com/llsc12/Paicord/tree/android-2

thank you!

ktoso:
This also resolves https://github.com/swiftlang/swift-java/issues/822 while at it

### DIFF
--- a/Sources/SwiftExtract/SwiftTypes/SwiftParsedModuleSymbolTableBuilder.swift
+++ b/Sources/SwiftExtract/SwiftTypes/SwiftParsedModuleSymbolTableBuilder.swift
@@ -174,11 +174,37 @@ extension SwiftParsedModuleSymbolTableBuilder {
     parent: SwiftNominalTypeDeclaration
   ) {
     for member in node.members {
-      // Find any nested types within this nominal type and add them.
-      if let nominalMember = member.decl.asNominal {
-        self.handle(sourceFilePath: sourceFilePath, nominalTypeDecl: nominalMember, parent: parent)
-      } else if let typeAliasMember = member.decl.as(TypeAliasDeclSyntax.self) {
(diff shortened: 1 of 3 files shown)


ktoso:
Also handling the extension case better now, thanks @llsc12!

PR #842, opened 2026-07-13
wrap-java: support java record types

ktoso:
Simple addition to support record types using existing infra. This adds a new macro but it is just sugar over the existing JavaClass macro


ktoso:
The stdlib wrap check just failed because we're adding more comments, this is OK.

```

+/// Java `sealed class`, permits: `java.lang.reflect.Constructor`, `java.lang.reflect.Method`
 @JavaClass("java.lang.reflect.Executable", implements: GenericDeclaration.self)
 open class Executable: AccessibleObject {
   /// Java method `getAnnotatedExceptionTypes`.
```

PR #843, opened 2026-07-13
wrap-java: handle sealed/permits in interfaces and classes

ktoso:
This updates the JavaClass and interface macros with additional permits which handle Java's `sealed`. 

```swift
@JavaClass(.sealed, "com.example.Shape", permits: Circle.self, Square.self)
@JavaClass(.final,  "com.example.Leaf")
```

I decided to not change codegen so far, since we have a good enough `func as(T) -> T?` already; so this is mostly docs; but we could generate way more now, including switching over the "case" of it using Swift enums.


ktoso:
Sadly this doens't quite work out, will need to redesign a bit.

PR #844, opened 2026-07-13
jextract: Organize java known functional interfaces

AmrDeveloper:
Organize the Java known functional interfaces information in one place

Issue #811

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1691,7 +1691,7 @@ extension JNISwift2JavaGenerator {
 
 
         let methodName =
-          if let known = KnownFunctionalInterface(methodSignature.parameterTypes, methodSignature.resultType) {
+          if let known = KnownJavaFunctionalInterface.find(methodSignature) {
             known.method
           } else {
             "apply"
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1689,9 +1689,10 @@ extension JNISwift2JavaGenerator {
           $0.conversion.render(&printer, $0.parameters.first!.name)
         }
 
+
         let methodName =
-          if parameters.isEmpty && nativeResult.javaType.isVoid {
-            KnownFunctionalInterface.runnable.method
+          if let known = KnownFunctionalInterface(methodSignature.parameterTypes, methodSignature.resultType) {
+            known.method
           } else {
             "apply"
           }
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1691,7 +1691,7 @@ extension JNISwift2JavaGenerator {
 
         let methodName =
           if parameters.isEmpty && nativeResult.javaType.isVoid {
-            "run"
+            KnownFunctionalInterface.runnable.method
           } else {
             "apply"
           }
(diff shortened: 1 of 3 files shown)


Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1691,7 +1691,7 @@ extension JNISwift2JavaGenerator {
 
         let methodName =
           if parameters.isEmpty && nativeResult.javaType.isVoid {
```
ktoso: I'd like this to be `if let known = KnownJavaFunctionalInterface(method, nativeResult.javaType) { ... } else { "apply" }` maybe, so centralize the matching the signatures to the types

AmrDeveloper: I updated it to depend on parameters and result types, not sure if there is a better syntax or a way to do this

ktoso:
Digging through this a bit more, the shape we want is more like:

```

struct KnownJavaFunctionalInterface: Sendable {
  let javaType: JavaType
  let method: String
  let parameters: [JavaType]
  let result: JavaType

  static let runnable = KnownJavaFunctionalInterface(
    JavaType.javaLangRunnable,
    method: "run",
    parameters: [],
    result: .void
  )
```

because I want the type name, method name, and types all in one place, rather than spread out in the other funcs.

Hope you don't mind if I apply this small change and we merge this; looking forward to follow ups handling the other types!

ktoso:
I also noticed the functional interface usage isn't really complete... We're still emitting the `@FunctionalInterface public interface Function` even though we don't use it anymore. we Should fix that as well.

AmrDeveloper:
> Digging through this a bit more, the shape we want is more like:
> 
> ```
> 
> struct KnownJavaFunctionalInterface: Sendable {
>   let javaType: JavaType
>   let method: String
>   let parameters: [JavaType]
>   let result: JavaType
> 
>   static let runnable = KnownJavaFunctionalInterface(
>     JavaType.javaLangRunnable,
>     method: "run",
>     parameters: [],
>     result: .void
>   )
> ```
> 
> because I want the type name, method name, and types all in one place, rather than spread out in the other funcs.
> 
> Hope you don't mind if I apply this small change and we merge this; looking forward to follow ups handling the other types!

I don't mind, and I can do that too :D

AmrDeveloper:
> I also noticed the functional interface usage isn't really complete... We're still emitting the `@FunctionalInterface public interface Function` even though we don't use it anymore. we Should fix that as well.

Yes, I mentioned that here

https://github.com/swiftlang/swift-java/pull/820#discussion_r3572776737

Will handle it in follow-up

ktoso:
Would you want to do the follow up please?

Inside
```
  /// Print FFM Java binding descriptors for the imported Swift API.
  package func printJavaBindingDescriptorClass(
    _ printer: inout JavaPrinter,
    _ decl: ExtractedFunc,
  ) {
```

before we do 

```
printJavaBindingDescriptorClass(&printer, cFunc, symbolLookup: currentSymbolLookup) { printer in
```

we need to guard it like

```
if let known = KnownJavaFunctionalInterface.find(parameters: cParameterTypes, result: cResultType) {
  // skip, this is a known type/shape (e.g. Runnable),
  // so there's no need to emit an one-off functional interface for it.
  continue
}
```

Would you make a follow up PR with that please?

ktoso:
Thanks!

ktoso (approved):

AmrDeveloper:
> Would you want to do the follow up please?
> 
> Inside
> 
> ```
>   /// Print FFM Java binding descriptors for the imported Swift API.
>   package func printJavaBindingDescriptorClass(
>     _ printer: inout JavaPrinter,
>     _ decl: ExtractedFunc,
>   ) {
> ```
> 
> before we do
> 
> ```
> printJavaBindingDescriptorClass(&printer, cFunc, symbolLookup: currentSymbolLookup) { printer in
> ```
> 
> we need to guard it like
> 
> ```
> if let known = KnownJavaFunctionalInterface.find(parameters: cParameterTypes, result: cResultType) {
>   // skip, this is a known type/shape (e.g. Runnable),
>   // so there's no need to emit an one-off functional interface for it.
>   continue
> }
> ```
> 
> Would you make a follow up PR with that please?

Yes, I will. I just wanted to make them small incremental changes so we can easily review and merge them in short iterations, and I can get a review of the language features to use :D

Thanks

PR #845, opened 2026-07-14
Support installing docc plugin via env var

ktoso:
This is a workaround for `swift package add-dependency` not working due to swift-subprocess pulling it in explicitly:

```
[swift-java] -> % swift package add-dependency https://github.com/swiftlang/swift-docc-plugin --from 1.1.0
error: unable to add dependency 'https://github.com/swiftlang/swift-docc-plugin' because it already exists in the list of dependencies
```

even though

```
-> % swift package generate-documentation
error: Unknown subcommand or plugin name 'generate-documentation'
Usage: swift package <options> <subcommand>
```

So this is a workaround to enable the plugin in this package


ktoso:
Related swiftpm issue https://github.com/swiftlang/swift-package-manager/issues/10294

ktoso:
Verified in earlier run

PR #850, opened 2026-07-22
improve the title presented in documentation

heckj:
- adds spaces into the title as it's presented either in combined docs or on Swift Package Index
 - removes a missing doc:// reference to Distribution


ktoso (approved):
Thank you!

PR #851, opened 2026-07-27
jextract: support returning JavaKit wrapped Java objects (Resolves #841)

### LINKED ISSUE #841: jextract: support returning JavaKit wrapped Java objects
We support accepting JavaKit annotated types into the Swift side but don't handle returning them too well.

We should allow `func` make() -> JavaLong` etc.

amanmaurya92:
## Description
Resolves #841. This PR adds support for returning JavaKit-wrapped Java objects (like `JavaLong` or `JavaString`) from Swift functions back to Java.

### Approach
- **Type Detection:** Updated `translateResult`, `translateOptionalResult`, and `translateArrayResult` in both `NativeTranslation` and `JavaTranslation` to recognize types passing the `isSwiftJavaWrapper` predicate.
- **Java Shim:** Maps the return type to the corresponding underlying Java class (e.g., `java.lang.Long`) rather than throwing an unsupported type error.
- **Native Translation (C bindings):** Returns the Swift value wrapped in an `Optional`, and calls `.getJNIValue(in: environment)` to safely extract and return the underlying JNI object reference (or `NULL` if missing).
- **Tests:** Added tests to `JNIJavaKitTests` for functions returning standard wrappers, optional wrappers, and arrays of wrappers.

### DIFF
--- a/Tests/JExtractSwiftTests/JNI/JNIJavaKitTests.swift
+++ b/Tests/JExtractSwiftTests/JNI/JNIJavaKitTests.swift
@@ -76,5 +76,141 @@ struct JNIJavaKitTests {
         """
       ]
     )
+  @Test
+  func functionReturn_javaBindings() throws {
+    try assertOutput(
+      input: "public func function() -> JavaLong",
+      .jni,
+      .java,
+      javaClassLookupTable: classLookupTable,
+      expectedChunks: [
+        """
+        /**
+         * Downcall to Swift:
+         * {@snippet lang=swift :
+         * public func function() -> JavaLong
+         * }
+         */
+        public static java.lang.Long function() {
+          return SwiftModule.$function();
+        }
+        """,
+        """
+        private static native java.lang.Long $function();
+        """,
+      ]
+    )
+  }
+
+  @Test
+  func functionReturn_swiftThunks() throws {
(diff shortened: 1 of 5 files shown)


madsodgaard:
I haven't read the code, but please also add some runtime tests to the JNI sample app

Tests/JExtractSwiftTests/JNI/JNIJavaKitTests.swift
```diff
@@ -77,4 +76,139 @@ struct JNIJavaKitTests {
...
+      input: "public func function() -> [JavaLong]",
+      .jni,
+      .java,
+      javaClassLookupTable: classLookupTable,
+      expectedChunks: [
+        """
+        /**
+         * Downcall to Swift:
+         * {@snippet lang=swift :
+         * public func function() -> [JavaLong]
```
ktoso, line 182: The array case doesn't actually work AFAICS? 

Please add a runtime test in Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/MySwiftClass.swift so we can verify

ktoso: Hmm seems I was wrong? Let's merge this

Issue #18, opened 2024-09-30
Extend Swift wrappers of Java classes in modules that have made more methods/fields available


DougGregor:
The [`JavaClass` type](https://github.com/swiftlang/swift-java/blob/main/Sources/JavaKit/JavaClass.swift), which is the Swift wrapper for Java's `java.lang.Class`, is provided in the Swift `JavaKit` module. Many of the methods on `java.lang.Class` require classes from `java.lang.reflect` that are part of the Swift `JavaKitReflection` module. For example, [`getDeclaredMethods()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredMethods--) requires [`java.lang.reflect.Method`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html) that is only [available in `JavaKitReflection`](https://github.com/swiftlang/swift-java/blob/main/Sources/JavaKitReflection/generated/Method.swift).

Java2Swift drops any methods/constructors that involve Java classes that have not been wrapped in Swift, so `JavaKit` is missing any of those methods. Instead, `JavaKitReflection` has a handcrafted [extension of `JavaClass`](https://github.com/swiftlang/swift-java/blob/main/Sources/JavaKitReflection/JavaClass%2BReflection.swift) that adds all of the methods from `java.lang.Class` that involve at least one type from the Java classes that are wrapped into Swift within `JavaKitReflection`.

We should teach `Java2Swift` to recognize those cases where a Java class has already been wrapped in Swift (e.g., by using the [translation manifest](https://github.com/swiftlang/swift-java/blob/main/Sources/Java2Swift/TranslationManifest.swift)) and produce an extension of the Swift wrapper type that includes only those methods/constructors/etc. that involve at least one Java class that's being wrapped in the module being generated. The manually-written `JavaKitReflection/JavaClass+Refection.swift` file referenced above could therefore be replaced with a generated one.

Tatenda-k:
Hey, 
Is this issue still active. I would like to start working on it.

PR #852, opened 2026-07-27
generate extensions of wrapper types owned by other modules

Tatenda-k:
## What
Teaches `wrap-java` to generate a Swift `extension` of a wrapper type that is defined in a different Swift module, containing only the members whose signatures or return type references at least one type wrapped by the module being generated. 
Opt in per-module via a new manifest field: 
"extendForeginclasses": ["java.lang.Class"]

## Update
I have generated the foreign extensions.
I added classes to the `extendForeignClasses` arrays by running a script that regenerates every module and captures each [warning] Unable to translate '<owner>' <kind> '<member>': Java class '<type>' has not been translated . It then builds a Java-class → owning-module map from the classes map of every Sources/**/swift-java.config,  and adds to the list if the dropped member's unresolved type is wrapped by a different module. 
#18


Sources/JavaStdlib/JavaLangReflect/JavaClass+Reflection.swift
```diff
@@ -16,6 +16,15 @@ import SwiftJava
...
 // knowing that JavaClass was defined elsewhere.
+//
+// NOTE: swift-java now has the capability to generate exactly this kind of
+// extension (see `Configuration.extendForeignClasses` and
+// `JavaTranslator.translateClassAsForeignExtension`). Replacing this file with a
+// generated one is a follow-up: the swift-java tool itself calls several of these
+// methods (e.g. `getDeclaredMethod`, `getFields`, `getGenericSuperclass`) to do
+// its work, so the generated `JavaClass+Extensions.swift` must first be produced
+// and committed (a one-time bootstrap) before the hand-written members here can
+// be removed.
```
ktoso: Please don't keep such LLM-ish comments that serve as roadmaps or todolists, we should do this in the same commit rather I think

Sources/SwiftJavaTool/Commands/WrapJavaCommand.swift
```diff
@@ -352,6 +352,58 @@ extension SwiftJava.WrapJavaCommand {
...
+      translator.startNewFile()
+      let extensionDecls = try translator.translateClassAsForeignExtension(foreignClass)
+      if extensionDecls.isEmpty {
+        log.info("No members to add in extension of foreign class '\(foreignClassName)'; skipping.")
+        continue
+      }
+      let importDecls = translator.getImportDecls()
+
+      let swiftFileText = """
+        // Auto-generated by Java-to-Swift wrapper generator.
```
ktoso: That's not really a term we use; the tool is `swift-java wrap-java` or just `wrap-java` here.

Sources/SwiftJavaConfigurationShared/Configuration.swift
```diff
@@ -219,6 +219,11 @@ public struct Configuration: Codable {
   /// the corresponding Swift names (e.g., JavaVector).
   public var classes: [String: String]? = [:]
 
+  /// Fully-qualified Java class names whose Swift wrapper type is defined in a
+  /// *different* Swift module (a dependency), but which should gain extra members
+  /// generated by the module this configuration belongs to (`swiftModule`).
+  public var extendForeignClasses: [String]? = nil
```
ktoso: I think we have a typealias for these, we could use `JavaFullyQualifiedTypeName` here

ktoso (commented):
I'm supportive of the fix but it's pretty messy and very LLM-ish, would you mind refactoring it a bit please so it's more structured?

Also, the "follow up" should be done immediately in the same PR I think

Sources/SwiftJavaTool/Commands/WrapJavaCommand.swift
```diff
@@ -352,6 +352,58 @@ extension SwiftJava.WrapJavaCommand {
...
+    // modules. For example, when generating the JavaLangReflect module this adds
+    // an `extension JavaClass { ... }` — the `JavaClass` type itself is defined
+    // over in the SwiftJava module; here we only bolt on the reflection methods
+    // that SwiftJava couldn't express (the ones mentioning `Method`, `Field`, ...).
+    //
+    // Member selection — only methods that reference a type wrapped by *this*
+    // module (`effectiveSwiftModule`) — is implemented in
+    // `JavaClassTranslator.renderForeignExtension`.
+
+    for foreignClassName in config.extendForeignClasses ?? [] {
```
ktoso: All this should be put into a new method I think, there it can have the comment; please write the comment youtrself, this sounds very LLM-ish and is a bit hard to understand tbh.

Sources/SwiftJavaToolLib/JavaClassTranslator.swift
```diff
@@ -628,6 +628,60 @@ extension JavaClassTranslator {
...
+  /// is owned by the module currently being generated.
+  private func referencesCurrentModule(_ method: Method) -> Bool {
+    var referenced: [JavaClass<JavaObject>?] = [method.getReturnType()]
+    referenced.append(contentsOf: method.getParameterTypes())
+    return referenced.contains { owningModule(of: $0) == translator.swiftModuleName }
+  }
+
+  /// Render an `extension <Wrapper> { ... }` adding members that reference at
+  /// least one type owned by the current module.
+
```
ktoso, line 660: Stray newline, please remove
```suggestion
```

Sources/SwiftJavaTool/Commands/WrapJavaCommand.swift
Tatenda-k: Yes, sir. I agree. Thank you for your patience

Sources/JavaStdlib/JavaLangReflect/JavaClass+Reflection.swift
Tatenda-k: Sure.

Sources/SwiftJavaTool/Commands/WrapJavaCommand.swift
Tatenda-k: Okay. However, most of the generated extensions are preceded by " // Auto-generated by Java-to-Swift wrapper generator. " so I've decided to keep that when generating extensions.

Sources/SwiftJavaConfigurationShared/Configuration.swift
Tatenda-k: Ok.

Sources/SwiftJavaTool/Commands/WrapJavaCommand.swift
ktoso: Hmm I see, we need to fix that. I’ll follow up on it, we can land this as is

Sources/JavaStdlib/JavaNet/swift-java.config
```diff
@@ -1,4 +1,5 @@
 {
+  "extendForeignClasses" : ["java.net.URL", "java.io.File", "java.nio.file.Path"],
```
ktoso: I'm a bit iffy about the word "foreign" here because foreign implies "native" in such a project as we are because of FFI -- "foreign" function interface" etc which swift-java is all about hm.

ktoso:
Small comment while I'm reviewing -- I think this is good, but I'm wondering if we can get away without specififying the extended types in the configuration file. I'll look into it some more, but if you're interested in figuring that out that'd be great :)

Sources/JavaStdlib/JavaNet/swift-java.config
Tatenda-k: I agree. I'll change it to `extendClasses`

ktoso:
Conflicting pr sadly, can you resolve it please?

ktoso:
```

Build of product 'generate-config-docs' complete! (3.25s)
error: Sources/SwiftJavaDocumentation/Documentation.docc/SwiftJavaConfigFile.md is stale.
Run 'swift run generate-config-docs' and commit the result.
```

ktoso:
I'll rebuild it

Tatenda-k:
> I'll rebuild it

Thankyou, sir.

PR #859, opened 2026-08-03
Fix: Expand Swift variadic functions into explicit Java overloads

### LINKED ISSUE #830: jextract: Variadic parameters silently narrowed to single element
`SwiftParameter.isVariadic` is parsed, and `SwiftFunctionSignature.hasVariadicParams`
exists, but generators don't handle it. So `Int...` gets emitted as a single `Int`:

```swift
// Swift
public func sum(_ xs: Int64...) -> Int64 { xs.reduce(0, +) }
```

```java
// Generated Java (both modes): only one element can ever be passed
long r = MySwiftLibrary.sum(42); // no way to pass 0 or N values
```

Instead we should make those java var args

amanmaurya92:
Fixes #830.

## Description

This PR introduces support for generating Java bindings for Swift variadic functions. Previously, variadic functions were not handled correctly in the generated Java code.

### Changes Made
- **Variadic Expansion**: Added a mechanism to expand variadic Swift functions (e.g., `func log(items: String...)`) into multiple explicit Java overloads up to a configurable limit (`effectiveMaxVariadicOverloads`).
- **Immutability Maintained**: Refactored `FFMSwift2JavaGenerator` and `JNISwift2JavaGenerator` to perform the overload expansion using a local variable during initialization. This keeps `analysis` as an immutable `let` property on the generator classes, addressing previous review feedback.
- **Initialization Order Fix**: Ensured that `self.analysis` is fully initialized *before* invoking instance methods like `generateInterfaceWrappers()` in `JNISwift2JavaGenerator`, resolving Swift compiler errors regarding premature use of `self`.

This ensures we get comprehensive Java bindings for variadic functions without sacrificing the immutability of the generator's internal state.


ktoso:
Though I wonder it it would be better to support it by generating a number of overloads with up to N parameters where we take the N from configuration maybe?

amanmaurya92:
Yeah, I like that approach. It seems like a much cleaner solution.
I'll update the PR to do that.

ktoso:
Thanks!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
```diff
@@ -120,18 +120,31 @@ package class JNISwift2JavaGenerator: Swift2JavaGenerator {
...
+    expandedAnalysis.extractedGlobalFuncs = expandedAnalysis.extractedGlobalFuncs.flatMap { 
+      $0.expandingVariadicOverloads(maxOverloads: maxOverloads) 
     }
-
+    
+    for type in expandedAnalysis.extractedTypes.values {
+      type.methods = type.methods.flatMap { $0.expandingVariadicOverloads(maxOverloads: maxOverloads) }
+      type.initializers = type.initializers.flatMap { $0.expandingVariadicOverloads(maxOverloads: maxOverloads) }
+    }
+    self.analysis = expandedAnalysis
```
ktoso, line 126: I don't love that we're duplicating this expansion code in initializers; can we put it somehow into analysis instead so we don't have to duplicate it?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator.swift
```diff
@@ -120,18 +120,31 @@ package class JNISwift2JavaGenerator: Swift2JavaGenerator {
...
+    
     // Every extracted protocol that also gets a plain Java `interface`
     // generated for it is eligible to be boxed as an existential.
-    self.existentialProtocolBoxes = self.analysis.extractedTypes.values
+    self.existentialProtocolBoxes = expandedAnalysis.extractedTypes.values
       .filter { $0.swiftNominal.kind == .protocol }
       .sorted { $0.swiftNominal.qualifiedName < $1.swiftNominal.qualifiedName }
+    
+    if config.enableJavaCallbacks ?? false {
```
ktoso: There should be config.effectiveEnableJavaCallbacks so we don't have to randomly ?? in consumers of it

Sources/SwiftExtract/ExtractedDecls.swift
```diff
@@ -400,6 +400,61 @@ public final class ExtractedFunc: ExtractedSwiftDecl, CustomStringConvertible {
       functionSignature: functionSignature
     )
   }
+
+  /// Expands this function into `maxOverloads + 1` functions if it contains a variadic parameter.
+  /// Replaces the variadic parameter `T...` with `N` discrete parameters (`arg0: T`, `arg1: T`, etc.)
+  /// for `N` in `0...maxOverloads`.
+  /// Returns `[self]` if the function has no variadic parameters.
+  public func expandingVariadicOverloads(maxOverloads: Int) -> [ExtractedFunc] {
```
ktoso, line 408: This is good, yeap!

Sources/SwiftJavaConfigurationShared/Configuration.swift
```diff
@@ -59,6 +59,14 @@ public struct Configuration: Codable {
 
   public var outputJavaDirectory: String?
 
+  /// Maximum number of overloads to generate for a function with a variadic parameter.
+  /// When a variadic parameter `T...` is encountered, the generator will produce
+  /// `0` to `maxVariadicOverloads` distinct overloads instead of failing.
```
ktoso:
```suggestion
  /// up to `maxVariadicOverloads` distinct overloads instead of failing.
  ///
  /// Example:
  /// ```swift
  /// func concat(s: String...) -> String
  /// ```
  /// results in:
  /// ```java
  /// // Java
  /// String concat() -> String
  /// String concat(s0: String, s1: String) -> String
  /// String concat(s0: String, s1: String, s2: String) -> String
  /// ```
  ///
  /// The reason for this is that Swift cannot "splat" an array into a `...` 
  /// parameter, therefore we cannot transfer an arbitrary amount of varargs 
  /// parameters over the native boundary.
```

ktoso: I'd like to have examples here since we are going to start generating documentation pages from these comments

Tests/JExtractSwiftTests/JNI/JNIModuleTests.swift
```diff
@@ -326,4 +326,42 @@ struct JNIModuleTests {
...
+        """,
+        """
+        public static long sum(long arg0, long arg1, long arg2)
+        """
+      ],
+      notExpectedChunks: [
+        "sum(long arg0, long arg1, long arg2, long arg3)"
+      ]
+    )
+  }
```
ktoso, line 366: Please make a runtime test; we make those by adding a func in the `Samples/SwiftJavaExtractJNISampleApp` and exercise the functions in Java test code in the same project. This verifies it all works at runtime as well

ktoso (commented):
This is shaping up great, some comments still and please add runtime testing

ktoso (approved):

ktoso:
Thanks this is pretty good, we might want to move the expand... outside the initializer though, but we can do it in a follow up

amanmaurya92:
Thanks! I'll fix the failing checks and update the PR.

ktoso (approved):

.github/actions/prepare_env/action.yml
```diff
@@ -7,14 +7,55 @@ runs:
...
       uses: actions/setup-java@v4
       with:
         distribution: ${{ matrix.jdk_vendor }}
         java-version: |
           25
           17
         cache: 'gradle'
+      continue-on-error: true
+
+    - name: Ensure JAVA_HOME_17 / JAVA_HOME_25 (fallback if setup-java failed)
```
ktoso: Why would we need this, seems like llm random ideas?

amanmaurya92: Oops, that was accidental. I was experimenting with fixing some of the CI failures and forgot to remove it before committing.

ktoso: No prob, just flagging it :) thanks for cleanup

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+SwiftThunkPrinting.swift
```diff
@@ -709,7 +709,7 @@ extension JNISwift2JavaGenerator {
         decl.functionSignature.parameters,
         arguments,
       ).map { originalParam, argument in
-        let label = originalParam.argumentLabel.map { "\($0): " } ?? ""
+        let label = originalParam.argumentLabel.flatMap { $0 == "_" ? nil : "\($0): " } ?? ""
```
ktoso: This looks like scope creep? do we need this here? It's also very repetetive

amanmaurya92:
I didn't expect to run into so many formatting issues. I'll get the formatter working locally before my next PR so I can catch these ahead of time.

ktoso:
Hah yeah, sorry about this -- sadly the formatting can be pretty picky, definitely worth getting working locally

ktoso (approved):

ktoso:
I'll apply fixes, since I'm cutting a release now -- please look into getting formatter working locally though, it should mostly just work tbh 🤔

ktoso:
Thanks again!

PR #887, opened 2026-08-17
jextract: NFC Fix calling wrong function in the unittest

AmrDeveloper:
Fix calling the wrong function from MySwiftLibrary in the unittest


ktoso (approved):
Thanks I missed this too

PR #890, opened 2026-08-18
jextract: support extracting class members

### LINKED ISSUE #835: jextract: class methods and vars should be supported
Today we just skip them entirely AFAICS

```
open class Base {
  public class func defaultBase() -> Base { Base() }
  public class var shared: Base { Base() }
}
```

AbdAlRahmanGad:
closes: #835

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/MySwiftClass.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/MySwiftClass.swift
@@ -55,6 +55,16 @@ public class MySwiftClass {
   public static func method() {
   }
 
+  // snippet.classMembers
+  public class func classMethod(x: Int64) -> Int64 {
+    x * 2
+  }
+
+  public class var classVariable: Int64 {
+    42
+  }
+  // snippet.end
+
   // snippet.initializers
   public init(x: Int64, y: Int64) {
     self.x = x
--- a/Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftClass.swift
+++ b/Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/MySwiftClass.swift
@@ -39,6 +39,14 @@ public class MySwiftClass {
     MySwiftClass(x: x, y: y)
   }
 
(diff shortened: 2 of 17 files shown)


Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/MySwiftClass.swift
```diff
@@ -55,6 +55,16 @@ public class MySwiftClass {
...
 
+  // snippet.classMembers
+  public class func classMethod(x: Int64) -> Int64 {
+    x * 2
+  }
+
+  public class var classVariable: Int64 {
+    42
+  }
+  // snippet.end
```
ktoso, line 66: Thanks for adding snippets but we're not using them it seems -- can you add a small bit in `FeaturesJextract.md` around static methods somewhere:


```diff
+### Class methods
+
+Swift `class` methods and computed properties are imported as `static` methods on the generated Java wrapper type.
+
+@TabNavigator {
+   @Tab("Swift") {
+      @Snippet(path: "Snippets/ClassesSwift.swift", slice: "classMembers")
+   }
+   @Tab("Java (JNI)") {
+      @Snippet(path: "Snippets/MySwiftClassTest", slice: "classMembers")
+   }
+   @Tab("Java (FFM)") {
+      @Snippet(path: "Snippets/MySwiftClassTest", slice: "classMembers")
+   }
+}
```

And make sure we have the snippets for the usage as well

ktoso:
I think we missed a condition in JNISwift2JavaGenerator.swift:175 in `func supportedProtocolRequirements(of type: ExtractedNominalType) -> [ExtractedFunc] {` maybe?

ktoso:
Just a minor nit, I think these semantics will work out fine given how we generate code for the java side :)

AbdAlRahmanGad:
> I think we missed a condition in JNISwift2JavaGenerator.swift:175 in func supportedProtocolRequirements(of type: ExtractedNominalType) -> [ExtractedFunc] { maybe?

It's actually there, but maybe we should drop the condition since protocols doesn't support `class` members?

ktoso:
Hm yeah, good question -- we could drop it there I think indeed; This PR is good though so let's land this and we can think about the latter :)

PR #895, opened 2026-08-19
Recognize actor (`distributed`) methods and extensions as Async functions

### LINKED ISSUE #837: jextract: allow importing `(distributed) actor` types, they're just classes with asynchronous code
We're skipping actors today, but we should explicitly allow them and just handle them like classes where all funcs are async unless nonisolatd.

AbdAlRahmanGad:
Fixes #837

### DIFF
--- a/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Distributed.swift
+++ b/Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Distributed.swift
@@ -1,0 +1,27 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift.org project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift.org project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Distributed
+import SwiftJava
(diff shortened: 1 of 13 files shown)


Sources/SwiftExtract/SwiftTypes/SwiftFunctionSignature.swift
```diff
@@ -36,7 +36,10 @@ public struct SwiftFunctionSignature: Equatable {
   }
 
   public var isImplicitlyAsync: Bool {
-    !isAsync && selfParameter?.selfType.isActor == true
+    guard !isAsync, case .instance(_, let selfType) = selfParameter else {
```
ktoso: yeap good, worth a test case as well with a static func on an actor

ktoso:
Also a test with a `nonisolated func` and a `@concurrent func` inside the actors would be good, those are not isolated to the actor.

Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Distributed.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Distributed
+import SwiftJava
+
+public distributed actor DistributedHi {
+  public typealias ActorSystem = LocalTestingDistributedActorSystem
```
ktoso, line 19: Yup that’s right! Sorry I didn’t send a complete snippet earlier

ktoso (approved):
Logic and tests looking good, nice work!

PR #898, opened 2026-08-21
[SwiftExtract] Add OpaquePointer to the known Swift stdlib types

kateinoigakukun:
OpaquePointer was the only member of the pointer family missing from the known-modules table, so any API using it was dropped with an unresolved type warning. Add it to the fake Swift module source and the known-type kind enums, and keep it unbridged (nil) in the JNI type translator like the rest of the pointer family.


ktoso (approved):
Lgtm thank you!

PR #899, opened 2026-08-21
[SwiftExtract] Language-neutrality cleanup

kateinoigakukun:
Remove the Java/JNI specifics that leaked into the language-neutral analysis layer so other consumers (e.g. JavaScriptKit's BridgeJS) can adopt it without inheriting swift-java concepts:

- Move `SwiftQualifiedTypeName.jniEscapedName` into JExtractSwiftLib.
- Make the 'unresolved type' hint text injectable via a new `SwiftExtractConfiguration.unresolvedTypeHint` requirement (default nil); the swift-java Configuration keeps the existing `--depends-on` wording, so jextract diagnostics are unchanged.
- Drop the unused swift-log dependency from the SwiftExtract target and its dead `import Logging` statements; the module has its own logger.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -384,7 +384,6 @@ let package = Package(
         .product(name: "SwiftParser", package: "swift-syntax"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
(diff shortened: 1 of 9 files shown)


ktoso:
All good changes, thank you! Really excited to start sharing the infra!

ktoso (approved):

PR #900, opened 2026-08-21
[SwiftExtract] Add a pluggable diagnostics sink for skipped declarations

kateinoigakukun:
Today the analyzer logs a warning and silently drops any declaration it cannot fully model, and consumers have no programmatic way to observe that. That is acceptable for jextract's opt-out model, but a consumer with an opt-in model (e.g. JavaScriptKit's BridgeJS, where every `@JS` declaration was explicitly marked for export) must fail the build with a source-located error instead of quietly not exporting the API.

Introduce SwiftExtractDiagnosticsSink, an optional observer supplied to `SwiftAnalyzer/analyze()`. Every skipped declaration is reported as a `SwiftExtractDiagnostic` carrying the affected syntax node, its source file path, the neutral message, and the underlying translation error, so consumers can render precise caret diagnostics or aggregate errors per run. The default (no sink) behavior and log output are unchanged.

### DIFF
--- a/Sources/SwiftExtract/SwiftAnalysisVisitor.swift
+++ b/Sources/SwiftExtract/SwiftAnalysisVisitor.swift
@@ -79,15 +79,15 @@ final class SwiftAnalysisVisitor {
       break // TODO: Implement associated types
 
     case .initializerDecl(let node):
-      self.visit(initializerDecl: node, in: parent)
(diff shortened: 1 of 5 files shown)


ktoso (approved):

ktoso:
Makes sense, let's get this in! Thanks for the cleanups like that, more than happy to make the core flexible like this!

PR #901, opened 2026-08-21
[SwiftExtract] Construct .existential for 'any P' types

kateinoigakukun:
The someOrAnyType translation had identical branches: both 'some P' and 'any P' produced SwiftType.opaque, so .existential was never constructed and 'any P' rendered as 'some P'. Construct .existential for the 'any' spelling. JExtractSwiftLib already handles .opaque and .existential in shared case arms, so its behavior is unchanged.

### DIFF
--- a/Sources/SwiftExtract/SwiftTypes/SwiftType.swift
+++ b/Sources/SwiftExtract/SwiftTypes/SwiftType.swift
@@ -525,7 +525,7 @@ extension SwiftType {
       if someOrAntType.someOrAnySpecifier.tokenKind == .keyword(.some) {
         self = .opaque(try SwiftType(someOrAntType.constraint, lookupContext: lookupContext))
       } else {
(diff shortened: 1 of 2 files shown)


Sources/SwiftExtract/SwiftTypes/SwiftType.swift
```diff
@@ -521,7 +521,7 @@ extension SwiftType {
       if someOrAntType.someOrAnySpecifier.tokenKind == .keyword(.some) {
         self = .opaque(try SwiftType(someOrAntType.constraint, lookupContext: lookupContext))
       } else {
-        self = .opaque(try SwiftType(someOrAntType.constraint, lookupContext: lookupContext))
+        self = .existential(try SwiftType(someOrAntType.constraint, lookupContext: lookupContext))
```
ktoso, line 528: Nice catch!

ktoso (approved):

PR #902, opened 2026-08-21
[SwiftExtract] Record the source file path on ExtractedFunc

kateinoigakukun:
ExtractedNominalType already exposes sourceFilePath, but functions did not carry theirs, so a consumer that groups output by input file (e.g. BridgeJS's per-file imported-API skeletons) could not attribute global functions or accessors to a file. Thread the path through the visitor into every ExtractedFunc; synthesized declarations keep nil.

Depends on https://github.com/swiftlang/swift-java/pull/900

### DIFF
--- a/Sources/SwiftExtract/ExtractedDecls.swift
+++ b/Sources/SwiftExtract/ExtractedDecls.swift
@@ -309,8 +309,8 @@ public final class ExtractedFunc: ExtractedSwiftDecl, CustomStringConvertible {
 
   public let functionSignature: SwiftFunctionSignature
 
-  // The short path from module root to the file in which this declaration was originally declared.
-  // E.g. for `Sources/Example/My/Types.swift` it would be `My/Types.swift`.
+  /// The short path from module root to the file in which this declaration was originally declared.
+  /// E.g. for `Sources/Example/My/Types.swift` it would be `My/Types.swift`.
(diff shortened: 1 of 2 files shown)


Sources/SwiftExtract/ExtractedDecls.swift
```diff
@@ -309,6 +309,10 @@ public final class ExtractedFunc: ExtractedSwiftDecl, CustomStringConvertible {
 
   public let functionSignature: SwiftFunctionSignature
 
+  // The short path from module root to the file in which this declaration was originally declared.
```
ktoso:
```suggestion
  /// The short path from module root to the file in which this declaration was originally declared.
```

Sources/SwiftExtract/ExtractedDecls.swift
```diff
@@ -309,6 +309,10 @@ public final class ExtractedFunc: ExtractedSwiftDecl, CustomStringConvertible {
...
+  // E.g. for `Sources/Example/My/Types.swift` it would be `My/Types.swift`.
```
ktoso:
```suggestion
  /// E.g. for `Sources/Example/My/Types.swift` it would be `My/Types.swift`.
```

ktoso (approved):
lgtm!

PR #903, opened 2026-08-21
[SwiftExtract] Make the analyzer instance API public and add resolveType

kateinoigakukun:
The static `SwiftAnalyzer.analyze()` entry point returns only the `AnalysisResult`, but some consumers need the analyzer itself after the run: type references can appear outside declaration signatures (e.g. in attribute arguments like `@JavaExport(as: Foo.self)` or BridgeJS's `@JS(as: Foo.self)`), and resolving those requires the module's symbol tables. Make `add(filePath:text:)` and `analyze(beforeProcessing...:)` public so a consumer can drive an analyzer instance directly, and add `resolveType(_:)` to resolve a `TypeSyntax` against the analyzed module after analysis.

### DIFF
--- a/Sources/SwiftExtract/SwiftAnalyzer.swift
+++ b/Sources/SwiftExtract/SwiftAnalyzer.swift
@@ -124,7 +124,7 @@ extension SwiftAnalyzer {
     )
   }
 
-  package func add(filePath: String, text: String) {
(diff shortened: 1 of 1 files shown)


ktoso:
Sounds good to me!

Issue #865, opened 2026-08-06
Async javaCallback selfPointer$


Flatout73:
# jextract/JNI: generated protocol-wrapper thunks for `async` requirements reference undeclared `selfPointer$` / `selfTypePointer$`

## Summary

With `mode: "jni"` and `enableJavaCallbacks: true`, a protocol that has an `async` requirement
produces a Swift thunk that does not compile. The generated `@_cdecl` entry point declares its
parameters as `selfPointer` / `selfTypePointer`, but the async capture preamble refers to
`selfPointer$` / `selfTypePointer$` (trailing `$`), which are never declared:

```
Greeter+SwiftJava.swift:89:50: error: cannot find 'selfPointer$' in scope
Greeter+SwiftJava.swift:90:54: error: cannot find 'selfTypePointer$' in scope
```

Synchronous requirements on the same protocol are generated correctly — they consistently use the
undecorated parameter names. Only the `async` path is affected.

This appears to be new in 0.5.0, which added async support to protocol wrappers. On 0.4.2 the
wrapper for the same protocol failed differently (it emitted a call to a Java member that the
Java-side interface did not have, since the async member is mapped to `CompletableFuture`).

## Environment

- swift-java **0.5.0**
- Swift 6.3.3 (`swift-6.3.3-RELEASE`), macOS 26 (Apple Silicon)
- `mode: "jni"`, `enableJavaCallbacks: true`
- Originally hit when cross-compiling for Android (`aarch64-unknown-linux-android28`,
  Swift Android SDK 6.3.3), but the faulty Swift is emitted by `jextract` itself and reproduces
  by running the tool directly on macOS.

## Reproducer

`Sources/Demo/Greeter.swift`:

```swift
public protocol Greeter: AnyObject {
    func greetSync(name: String)
    func greetAsync(name: String) async
}
```

`Sources/Demo/swift-java.config`:

```json
{
  "javaPackage": "com.example.demo",
  "mode": "jni",
  "enableJavaCallbacks": true
}
```

```bash
swift-java-tool jextract \
  --config Sources/Demo/swift-java.config \
  --swift-module Demo \
  --input-swift Sources/Demo \
  --output-java out-java \
  --output-swift out-swift
```

## Generated output

`out-swift/Greeter+SwiftJava.swift`, the `async` requirement (comments trimmed):

```swift
[...]

PR #905, opened 2026-08-23
JNI: fix non-compiling async thunks for protocol boxes and generic types

Flatout73:
### Motivation

Async methods in JNI mode capture `self` into the `Task` via `nonisolated(unsafe)`
bindings. The capture always referenced `<param>$`, but that binding only exists for
class/actor lowering (`.extractSwiftValue`). Protocol boxes load `self` into
`<param>Existential$`, and generic-Self downcalls render inside the opener extension
where `selfTypePointer` is never bound — so the generated Swift referenced unbound
identifiers and failed to compile.

### Changes

- Capture the binding the self lowering actually produced: `<param>$` for
  classes/actors, `<param>Existential$` for protocol boxes.
- Stop capturing the type-metadata parameter; the opener thunk neither receives
  nor binds it, and the task body never uses it.

### Testing

New `JNIAsyncSelfCaptureTests` pin the generated captures for all three self
shapes (class, protocol box, generic class); the generic and protocol tests fail
without this change. Full test suite passes (654 tests).

🤖 Fixed with help of Claude Code

https://github.com/swiftlang/swift-java/issues/865

### DIFF
--- a/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
+++ b/Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
@@ -1958,15 +1958,20 @@ extension JNISwift2JavaGenerator {
           """
         )
 
+        var selfCaptures: [(sendable: String, original: String)] = []
         if let selfParameter = nativeFunctionSignature.selfParameter {
-          for parameter in selfParameter.parameters {
-            printer.print("nonisolated(unsafe) let \(parameter.name)Sendable$ = \(parameter.name)$")
+          if case .extractSwiftProtocolValue = selfParameter.conversion {
+            for parameter in selfParameter.parameters {
+              selfCaptures.append(("\(parameter.name)ExistentialSendable$", "\(parameter.name)Existential$"))
+            }
+          } else {
+            for parameter in selfParameter.parameters {
+              selfCaptures.append(("\(parameter.name)Sendable$", "\(parameter.name)$"))
+            }
           }
         }
-        if let selfTypeParameter = nativeFunctionSignature.selfTypeParameter {
-          for parameter in selfTypeParameter.parameters {
-            printer.print("nonisolated(unsafe) let \(parameter.name)Sendable$ = \(parameter.name)$")
-          }
+        for capture in selfCaptures {
+          printer.print("nonisolated(unsafe) let \(capture.sendable) = \(capture.original)")
         }
 
         func printDo(printer: inout SwiftPrinter) {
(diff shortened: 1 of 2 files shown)


Tests/JExtractSwiftTests/JNI/JNIClassAsyncSelfCaptureTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+        """,
+      .jni,
+      .swift,
+      detectChunkByInitialLines: 1,
+      expectedChunks: [
+        """
+        @_cdecl("Java_com_example_swift_WorkerBox__00024work__JJLjava_util_concurrent_CompletableFuture_2")
+        ...
+        nonisolated(unsafe) let selfPointerExistentialSendable$ = selfPointerExistential$
```
ktoso: Heh, re-reading these now I guess we should be naming them `selfPointerExistentialUnsafeSendable$` tbh; It's just generated code but it'll be better to get the point across also for us reading the code that these are not actually safe "Sendable" -- I know this term was here before, would you mind changing those names to UnsafeSendable though please?

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1958,15 +1958,20 @@ extension JNISwift2JavaGenerator {
           """
         )
 
+        var selfCaptures: [(sendable: String, original: String)] = []
```
ktoso: I'd like to avoid calling those "Sendable" because we're doing unsafe things here (I know this was called Sendable$ before -let's fix this while we're here).

```suggestion
        var selfCaptures: [(unsafeSendable: String, original: String)] = []
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1958,15 +1958,20 @@ extension JNISwift2JavaGenerator {
...
         if let selfParameter = nativeFunctionSignature.selfParameter {
-          for parameter in selfParameter.parameters {
-            printer.print("nonisolated(unsafe) let \(parameter.name)Sendable$ = \(parameter.name)$")
+          if case .extractSwiftProtocolValue = selfParameter.conversion {
+            for parameter in selfParameter.parameters {
+              selfCaptures.append(("\(parameter.name)ExistentialSendable$", "\(parameter.name)Existential$"))
```
```suggestion
              selfCaptures.append(("\(parameter.name)ExistentialUnsafeSendable$", "\(parameter.name)Existential$"))
```

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1958,15 +1958,20 @@ extension JNISwift2JavaGenerator {
...
+            }
+          } else {
+            for parameter in selfParameter.parameters {
+              selfCaptures.append(("\(parameter.name)Sendable$", "\(parameter.name)$"))
```
```suggestion
              selfCaptures.append(("\(parameter.name)UnsafeSendable$", "\(parameter.name)$"))
```

ktoso:
Looks good, let's adjust the naming though while we're here (and run the formatter please).

Flatout73:
Fixed

ktoso:
Neat, thanks!

ktoso (approved):

PR #908, opened 2026-09-02
SwiftExtract: fix private(set) being ignored for variables with explicit accessors

### LINKED ISSUE #535: "public private(set) var" pattern fails to compile
# Reproduction

Found while testing #486. Reproducible with the following code in:
```
public class Foo {
    public private(set) var bar: String {
        get { fatalError("") }
        set { fatalError("") }
    }
}
```
in the `Samples/SwiftJavaExtractFFMSampleApp/Sources/MySwiftLibrary/Foo.swift`. This currently emits the following bridge and a compilation error:
```
error: cannot assign to property: 'bar' setter is inaccessible
 16 | @_cdecl("swiftjava_MySwiftLibrary_Foo_bar$set")
 17 | public func swiftjava_MySwiftLibrary_Foo_bar$set(_ newValue: UnsafePointer<Int8>, _ self: UnsafeRawPointer) {
 18 |   self.assumingMemoryBound(to: Foo.self).pointee.bar = String(cString: newValue)
    |                                                  `- error: cannot assign to property: 'bar' setter is inaccessible
```

tirdesh:
Fixes #535.

`supportedAccessorKinds(binding:minimumAccessLevel:)` returned early on
`binding.accessorBlock` before it reached the `private(set)` check, so a variable that both
restricts its setter's access level and spells out its accessors explicitly got a setter
binding emitted for a setter it cannot call:

```
error: cannot assign to property: 'bar' setter is inaccessible
 17 | public func swiftjava_MySwiftLibrary_Foo_bar$set(_ newValue: UnsafePointer<Int8>, _ self: UnsafeRawPointer) {
 18 |   self.assumingMemoryBound(to: Foo.self).pointee.bar = String(cString: newValue)
```

Moving the modifier check above the accessor block fixes it. The existing
`variable_int_privateSet` test already covers `public private(set) var` without an accessor
block, which is why this only shows up when the accessors are written out; the new test
covers that case.

`swift test` passes, 670 tests in 89 suites. The new test fails before the change and passes
after it.

### Not changed here

Subscripts have the same problem by a different path. `SwiftAnalysisVisitor.visit(subscriptDecl:)`
calls the no-argument `accessorBlock.supportedAccessorKinds()`, which never consults
`minimumAccessLevel`, so `public private(set) subscript` still emits a `$set` binding. I kept
this PR to the variable case. Happy to fold the subscript fix in here or open a separate
issue for it, whichever you prefer.

### DIFF
--- a/Sources/SwiftExtract/SwiftTypes/SwiftFunctionSignature.swift
+++ b/Sources/SwiftExtract/SwiftTypes/SwiftFunctionSignature.swift
@@ -545,15 +545,17 @@ extension VariableDeclSyntax {
       return [.get]
     }
 
-    if let accessorBlock = binding.accessorBlock {
-      return accessorBlock.supportedAccessorKinds()
-    }
-
-    // Account for private(set) and similar modifiers
+    // Account for private(set) and similar modifiers. This is checked before the
+    // accessor block, because a variable can restrict its setter's access level and
+    // still spell out its accessors explicitly.
     for modifier in self.modifiers where modifier.detail?.detail.text == "set" {
       if !minimumAccessLevel.matches(modifier) {
         return [.get]
       }
+    }
+
+    if let accessorBlock = binding.accessorBlock {
+      return accessorBlock.supportedAccessorKinds()
     }
 
     return [.get, .set]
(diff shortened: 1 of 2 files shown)


ktoso (approved):
Looks good, thank you

tirdesh:
@ktoso thanks for the review, lmk if anything else is needed here

ktoso (approved):

ktoso:
Thank you, will include this in the 0.6.0 now :)

PR #913, opened 2026-09-06
jextract/jni: Support async closures (#834)

### LINKED ISSUE #834: jextract: Support async closures
We're not supporting async closures yet, but it should be possible to bridge into them same as with async funcs.

```
public func schedule(op: () async -> Void) {}      
```

amanmaurya92:
Resolves #834.

Previously, async closures passed from Java to Swift were not supported in JNI mode. Passing an async closure either failed to translate or incorrectly matched synchronous functional interfaces like `Runnable` or `Function`.

This adds JNI support for `@escaping (...) async -> T` closures:

* Exclude async closures from matching synchronous `KnownJavaFunctionalInterface`s.
* Generate synthetic `@FunctionalInterface` wrappers whose `apply` method returns `CompletableFuture<T>` (or `SimpleCompletableFuture` in legacy future mode).
* Add `JavaCompletableFuture` and `JavaSimpleCompletableFuture` in `SwiftJava` with `@JavaMethod public func get() throws -> JavaObject?` to match JVM method descriptors during reflection.
* Lower async closure invocations in the Swift thunk to await `future$?.get()` and unbox the result via `T.fromJavaObject`.
* Add code generation tests in `JNIClosureTests` and real JVM runtime execution tests (void, primitive types, background completion, and exception propagation) in `SwiftJavaExtractJNISampleApp`.


ktoso:
Would you mind making the PR description less confusing? What does "classification" even mean.

And please add runtime tests.

The tests just use string but we need runtime tests for primitives as well.

I'll hold off reviewing until the test suite is more complete, this looks to me AI generated which we don't mind but please do spend more effort into validating the implementation is entirely correct, it takes much effort to review a partial PR like this.

Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Async.swift
```diff
@@ -12,33 +12,48 @@
 //
 //===----------------------------------------------------------------------===//
 
-import SwiftJava
 
 // snippet.asyncDefinition
+import SwiftJava
+
+
+// snippet.end
```
ktoso: this caused losing the asyncSum etc from the asyncDefinition snippets;

the end snippet must be after asyncSleep; and the new begin should be before the `asyncSchedule`

Sources/SwiftJava/JavaCompletableFuture.swift
```diff
@@ -0,0 +1,27 @@
...
+import SwiftJavaJNICore
+
+@JavaClass("java.util.concurrent.CompletableFuture")
+open class JavaCompletableFuture: JavaObject {
+  @JavaMethod
+  public func get() throws -> JavaObject?
+}
+
+@JavaClass("org.swift.swiftkit.core.SimpleCompletableFuture")
+open class JavaSimpleCompletableFuture: JavaObject {
```
ktoso: One type per one file please in those directories

ktoso: And let's rename this one to `SwiftJavaSimpleCompletableFuture` to be clear its a swift java type

Tests/JExtractSwiftTests/JNI/JNIClosureTests.swift
```diff
@@ -1159,4 +1159,235 @@ struct JNIClosureTests {
...
+      detectChunkByInitialLines: 1,
+      expectedChunks: [
+        """
+        public static class schedule {
+          /** Corresponds to the Swift closure parameter of type {@code () async -> Void}. */
+          @FunctionalInterface
+          public interface op {
+            java.util.concurrent.CompletableFuture<java.lang.Void> apply();
+          }
+        }
```
ktoso, line 1182: Neat, yep

Sources/SwiftJava/JavaCompletableFuture.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftJavaJNICore
+
+@JavaClass("java.util.concurrent.CompletableFuture")
+open class JavaCompletableFuture: JavaObject {
+  @JavaMethod
+  public func get() throws -> JavaObject?
+}
```
ktoso: I think this must be handled with wrap-java-generate.sh instead. We'd add the target type to Sources/SwiftJava/swift-java.config and this should be generated and not written by hand then

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
```diff
@@ -265,7 +265,24 @@ extension JNISwift2JavaGenerator {
...
+            .simpleCompletableFuture(translatedResult.javaType)
+          }
+        translatedResult = TranslatedResult(
+          javaType: futureType,
+          nativeJavaType: futureType,
+          annotations: translatedResult.annotations,
+          outParameters: translatedResult.outParameters,
+          conversion: translatedResult.conversion
+        )
+      }
```
ktoso, line 285: 👍

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -370,11 +370,11 @@ extension JNISwift2JavaGenerator {
       parentName: SwiftQualifiedTypeName
     ) throws -> NativeParameter {
       // @Sendable is not supported yet as "environment" is later captured inside the closure.
-      if functionType.isEscaping {
-        // For escaping closures we e need to create a Swift wrapper around
+      if functionType.isEscaping || functionType.isAsync {
```
ktoso, line 373: Yup, good!

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1760,6 +1756,48 @@ extension JNISwift2JavaGenerator {
           ? "{"
           : "{ \(closureParameters) in"
 
+        if fn.isAsync {
+          let tryKeyword = fn.isThrowing ? "try " : "try? "
```
ktoso: Hmmm, the `try?` I think means we're loosing throws/errors here,

ktoso: Can we add a test and handle throws?

Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/AsyncTest.java
```diff
@@ -87,4 +88,60 @@ void asyncString() throws Exception {
...
+
+    @Test
+    void asyncScheduleThrowing_exceptionPropagates() {
+        Future<Void> future = MySwiftLibrary.asyncScheduleThrowing(() -> {
+            CompletableFuture<Void> failed = new CompletableFuture<>();
+            failed.completeExceptionally(new RuntimeException("Java async failed"));
+            return failed;
+        });
+
+        assertThrows(ExecutionException.class, future::get);
```
ktoso, line 145: Hm, optimally we'd want to get the RuntimeException rethrown i guess... that might be hard to do, ok to do in a follow up PR

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+JavaTranslation.swift
ktoso: Oh no, actually this has an issue -- try writing a test for `async -> Int64?` , we're missing conversions here.

Samples/SwiftJavaExtractJNISampleApp/Sources/MySwiftLibrary/Async.swift
amanmaurya92: got it

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
amanmaurya92: thanks

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
amanmaurya92: I added runtime tests verifying that throwing closures properly propagate ExecutionException to the caller.

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
```diff
@@ -1760,6 +1756,48 @@ extension JNISwift2JavaGenerator {
...
+              {
+                guard let \(placeholder) else {
+                  fatalError("\(placeholder) is null")
+                }
+                let \(javaInterfaceVar) = \(syntheticFunction.javaInterfaceName)(javaThis: \(placeholder), environment: environment)
+                return \(closureHeader)
+                  let environment$ = try! JavaVirtualMachine.shared().environment()
+                  let future$ = \(upcallExpr)
+                  let result$ = \(tryKeyword)future$?.get()
+                  return \(fn.resultType.description).fromJavaObject(result$?.javaThis, in: environment$)
```
ktoso: This is wrong since we cannot hardcode the fn.result type -- it needs conversions here too.

I think we'll need a new `UpcallConversionStep`

```
  indirect case awaitFutureResult(
    UpcallConversionStep,
    resultType: SwiftType,
    isThrowing: Bool
  )
```

and handle it 

```
  case .awaitFutureResult(let inner, let resultType, let isThrowing):
    let future = inner.render(&printer, placeholder)
    let tryKeyword = isThrowing ? "try " : "try! "
    printer.print(
      """
      guard let future$ = \(future) else {
        fatalError("Async closure upcall to apply returned a nil future")
      }
      """
    )
    if resultType.isVoid {
      printer.print("_ = \(tryKeyword)future$.get()")
      return ""
    }
    printer.print(
      """
      let environment$ = try! JavaVirtualMachine.shared().environment()
      let result$ = \(tryKeyword)future$.get()
      """
    )
    return "\(resultType).fromJavaObject(result$?.javaThis, in: environment$)"
```

and finally compose it:

```
  var resultConversion = try self.translateResult(
    type: functionType.resultType,
    methodName: "apply"
  )

  if functionType.isAsync {
    // The future's payload arrives boxed as java.lang.Object, so only types we can
    // unbox from a boxed Object are supported. Reject the rest here rather than
    // emitting Swift that will not compile
    guard functionType.resultType.isDirectlyTranslatedToWrapJava else {
      throw JavaTranslationError.unsupportedSwiftType(functionType.resultType)
    }
    resultConversion = .awaitFutureResult(
      .placeholder,
      resultType: functionType.resultType,
      isThrowing: functionType.isThrowing
    )
  }
```

Sources/SwiftJava/swift-java.config
```diff
@@ -27,6 +27,7 @@
     "java.util.OptionalDouble": "JavaOptionalDouble",
     "java.util.OptionalInt": "JavaOptionalInt",
     "java.util.OptionalLong": "JavaOptionalLong",
+    "java.util.concurrent.CompletableFuture": "JavaCompletableFuture",
```
ktoso, line 30: Good yeap

Sources/JExtractSwiftLib/JNI/JNISwift2JavaGenerator+NativeTranslation.swift
amanmaurya92: Done! Added UpcallConversionStep.awaitFutureResult to handle awaiting the future and unboxing its payload across void, primitives, Java optionals (OptionalLong, OptionalInt, OptionalDouble, Optional<JavaString>), and Java objects.

and this completely removes the special-cased if fn.isAsync branch in escapingClosureLowering, unifying sync and async closures through the exact same resultConversion pipeline.

Samples/SwiftJavaExtractJNISampleApp/src/test/java/com/example/swift/AsyncTest.java
amanmaurya92: I will open the issue for this.

PR #917, opened 2026-09-16
test: re-enable FFM sample coverage

Hokila:
## Summary

- Re-enable the FFM Runnable upcall integration test in both sample apps.
- Update the JAR FFM sample to use `java.lang.Runnable` with `run()`.
- Re-enable the FFM instance-property test for `getX()`.
- Correct three UnaryOperator tests that accidentally called the corresponding BinaryOperator methods.

## Testing

```sh
./gradlew -PswiftJavaJdk=23 :Samples:SwiftJavaExtractFFMSampleApp:test :Samples:SwiftAndJavaJarFFMSampleLib:test --no-daemon
```

The relevant test suites pass with zero failures, errors, or skipped tests.

## Question for maintainers

These changes are related FFM sample-test maintenance across two sample applications. Would you prefer this to remain as one PR, or should it be split into separate PRs per sample/application?


ktoso:
Thanks for the contribution, a single PR is good, the sample apps serve as "end to end tests" really, so it is good to update them along with fixes

